PR #9, opened 2025-11-17
feat: Add ability to increase inhibitions

LokiMetaSmith:
This change introduces a new `mode` parameter that allows the user to either remove or increase the inhibitions of a language model.

The new `mode` can be set to either `remove_inhibitions` (the default behavior) or `increase_inhibitions`.

When `increase_inhibitions` is selected:
- The refusal direction is added to the model's weights instead of subtracted.
- The optimization process is configured to maximize refusals instead of minimizing them.
- The documentation has been updated to reflect this new functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,18 @@
-# Heretic: Fully automated censorship adjustment in language models
+# Heretic: Fully automatic censorship removal for language models
 
-Heretic is a tool that can remove censorship (aka "safety alignment") from
-transformer-based language models, or increase it, without expensive post-training.
+Heretic is a tool that removes censorship (aka "safety alignment") from
+transformer-based language models without expensive post-training.
 It combines an advanced implementation of directional ablation, also known
 as "abliteration" ([Arditi et al. 2024](https://arxiv.org/abs/2406.11717)),
 with a TPE-based parameter optimizer powered by [Optuna](https://optuna.org/).
 
 This approach enables Heretic to work **completely automatically.** Heretic
-finds high-quality abliteration parameters by co-optimizing the number of
+finds high-quality abliteration parameters by co-minimizing the number of
 refusals and the KL divergence from the original model. This results in a
-model that aligns with the user's goals while retaining as much of the
-original model's intelligence as possible. Using Heretic does not require an
-understanding of transformer internals. In fact, anyone who knows how to
-run a command-line program can use Heretic to manipulate the censorship of
-language models.
+decensored model that retains as much of the original model's intelligence
+as possible. Using Heretic does not require an understanding of transformer
+internals. In fact, anyone who knows how to run a command-line program
+can use Heretic to decensor language models.
 
 <img width="650" height="715" alt="Screenshot" src="https://github.com/user-attachments/assets/d71a5efa-d6be-4705-a817-63332afb2d15" />
 
@@ -56,10 +55,7 @@ pip install heretic-llm
 heretic Qwen/Qwen3-4B-Instruct-2507
 ```
 
-Replace `Qwen/Qwen3-4B-Instruct-2507` with the model you want to use.
-
-By default, Heretic removes inhibitions, but it can also be configured to
-increase them: `heretic --mode increase_inhibitions Qwen/Qwen3-4B-Instruct-2507`
+Replace `Qwen/Qwen3-4B-Instruct-2507` with whatever model you want to decensor.
 
 The process is fully automatic and does not require configuration; however,
 Heretic has a variety of configuration parameters that can be changed for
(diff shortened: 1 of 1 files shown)


LokiMetaSmith:
Because, maybe I want my model to be deathly afraid of talking about cats.

src/heretic/main.py
```diff
@@ -51,15 +51,16 @@ def run():
...
-    if (
-        # An odd number of arguments have been passed (argv[0] is the program name),
-        # so that after accounting for "--param VALUE" pairs, there is one left over.
-        len(sys.argv) % 2 == 0
-        # The leftover argument is a parameter value rather than a flag (such as "--help").
-        and not sys.argv[-1].startswith("-")
-    ):
-        # Assume the last argument is the model.
-        sys.argv.insert(-1, "--model")
+    if "--model" not in sys.argv:
```
p-e-w, line 54: Why this change?

src/heretic/main.py
```diff
@@ -179,7 +180,10 @@ def run():
         settings.model = settings.evaluate_model
         model.reload_model()
         print("* Evaluating...")
-        evaluator.get_score()
+        _, _, refusals = evaluator.get_score()
+        print(
+            f"* Refusals: [bold]{refusals}[/]/{len(evaluator.bad_prompts)}"
```
p-e-w, line 185: Why this change? Refusals are already printed by `get_score()`.

src/heretic/config.py
```diff
@@ -28,6 +28,11 @@ class Settings(BaseSettings):
         description="If this model ID or path is set, then instead of abliterating the main model, evaluate this model relative to the main model.",
     )
 
+    mode: Literal["remove_inhibitions", "increase_inhibitions"] = Field(
```
p-e-w: I think it's cleaner and simpler to just have a boolean called `reverse` here (default `False`), given that there are only two possible directions. This saves us from using literals everywhere, and also keeps a more neutral language as not every good/bad dataset is about "inhibitions".

src/heretic/main.py
LokiMetaSmith: It looks like a general catch all, and since we're adding a new argument I guess you can't assume the last argument is --model now.

src/heretic/main.py
LokiMetaSmith: no real reason, can swap it out

src/heretic/config.py
LokiMetaSmith: shouldn't be a problem, can fix

LokiMetaSmith:
On the topic of the ReadMe, 

It's kind of a personal choice on how you want to phrase the project, as a research and demo tool, it's fine to keep it laser focused, but if you want to allow this tool to grow into something that can go in and tinker with weights directly, then expanding the scope of the tool makes a lot of sense. 

I'm still wrapping my head around the implications of how all this falls in line with fine tuning, ablation and training foundational models. It feels like we're building up a vocabulary and toolset to really fine tune models for different use cases. 

I think for now, I'm going to revert all readme changes and allow you to figure out how you want to document the new parameter

src/heretic/main.py
p-e-w: I don't get it. There were already a dozen arguments that are not `--model`. What has changed that requires this modification?

src/heretic/main.py
p-e-w: Yes please, no unrelated changes should be in the PR.

src/heretic/main.py
```diff
@@ -302,14 +306,22 @@ def objective(trial: Trial) -> tuple[float, float]:
...
+            ),
+        ],
     )
 
     study.optimize(objective, n_trials=settings.n_trials)
 
     best_trials = sorted(
         study.best_trials,
         key=lambda trial: trial.user_attrs["refusals"],
+        reverse=(settings.mode == "increase_inhibitions"),
```
p-e-w, line 324: Broken with new setting.

src/heretic/main.py
```diff
@@ -302,14 +306,22 @@ def objective(trial: Trial) -> tuple[float, float]:
             n_ei_candidates=128,
             multivariate=True,
         ),
-        directions=[StudyDirection.MINIMIZE, StudyDirection.MINIMIZE],
+        directions=[
+            StudyDirection.MINIMIZE,
+            (
+                StudyDirection.MAXIMIZE
+                if settings.mode == "increase_inhibitions"
```
p-e-w, line 313: Broken with new setting.

src/heretic/config.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # Copyright (C) 2025  Philipp Emanuel Weidmann <pew@worldwidemann.com>
 
-from typing import Dict
+from typing import Dict, Literal
```
p-e-w, line 4: Unused import.

src/heretic/model.py
```diff
@@ -205,7 +205,10 @@ def abliterate(
 
                 for matrix in matrices:
                     # In-place subtraction is safe as we're not using Autograd.
-                    matrix.sub_(weight * (projector @ matrix))
+                    if self.settings.reverse:
+                        matrix.add_(weight * (projector @ matrix))
```
p-e-w, line 209: Broken. One of them should be `sub_`.

p-e-w:
> It's kind of a personal choice on how you want to phrase the project, as a research and demo tool, it's fine to keep it laser focused, but if you want to allow this tool to grow into something that can go in and tinker with weights directly, then expanding the scope of the tool makes a lot of sense.
>
> I'm still wrapping my head around the implications of how all this falls in line with fine tuning, ablation and training foundational models. It feels like we're building up a vocabulary and toolset to really fine tune models for different use cases.

I don't necessarily disagree with any of that, but I'll have to think about the correct scope for Heretic a bit more as I observe how people use the project.

p-e-w:
Please rebase on top of master to fix the conflicts and run CI.

p-e-w:
Superseded by #53, which will enable this and many other things.

PR #12, opened 2025-11-17
Ability to run on CPU only (no GPU)

noahvandal:
I have an older mac running intel silicon that was not working; slight pyproject.toml changes to allow for proper older versions of software to allow for non-GPU use cases (for testing, mainly)


p-e-w:
Did you get it working? CPU inference works just fine with the current version on my system.

noahvandal:
Yes, I was being dumb and used the default 3.12 python from uv. Setting it to 3.11 worked just fine. So I closed. Thank you!

noahvandal:
Actually, I am reopening this with a few other provisions for the case of older computers like mine (Mac-Pro, 2019, Intel x64, no GPU (maybe integrated graphics which shows up as MPS, e.g., Intel UHD 630 )). 

Was having compatibility issues (need PyTorch >2.1, <2.3; NumPy <2.0), along with this specific version not supporting BFloat. 

I hope this helps and doesn't cause issues with anyone else. My initial problem was just getting it installed, which the initial PR was for; this PR is for actual runtime use.

p-e-w:
I don't get it. Why should Python 3.12 fail? I use it all the time, and it works fine with CPU inference.

noahvandal:
You are correct, I have reverted the versioning for python; the original problem was with torch and numpy versions (which were resolved)

pyproject.toml
```diff
@@ -7,7 +7,7 @@ license = "AGPL-3.0-or-later"
 authors = [
     { name = "Philipp Emanuel Weidmann", email = "pew@worldwidemann.com" }
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.13"  # Supports 3.10-3.12 (verified with Python 3.12)
```
p-e-w: Does 3.13 fail?

noahvandal: I did not verify, 3.12 was the highest I verified.

pyproject.toml
```diff
@@ -22,6 +22,15 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
+    # PyTorch: Platform-specific constraints
+    # macOS Intel (x86_64): Use >=2.1.0,<2.3.0 (2.3+ dropped macOS Intel support; last with wheels was 2.2.0)
+    "torch>=2.1.0,<2.3.0; platform_machine == 'x86_64' and sys_platform == 'darwin'",
+    # macOS Silicon (ARM64): Can use newer versions with Python 3.12
+    "torch>=2.0.0; platform_machine == 'arm64' and sys_platform == 'darwin'",
```
p-e-w: Version >= 2.2 is required. 2.0 fails (don't quite remember what the problem was, some multiplication kernel issue I think).

noahvandal: Okay by default 2.2.2 is installed, we can set this param also

pyproject.toml
```diff
@@ -22,6 +22,15 @@ classifiers = [
...
 dependencies = [
+    # PyTorch: Platform-specific constraints
+    # macOS Intel (x86_64): Use >=2.1.0,<2.3.0 (2.3+ dropped macOS Intel support; last with wheels was 2.2.0)
+    "torch>=2.2.0,<2.3.0; platform_machine == 'x86_64' and sys_platform == 'darwin'",
+    # macOS Silicon (ARM64): Can use newer versions with Python 3.12
+    "torch>=2.2.0; platform_machine == 'arm64' and sys_platform == 'darwin'",
+    # Other platforms (Linux, Windows): Can use newer versions
+    "torch>=2.2.0; sys_platform != 'darwin'",
+    # NumPy: PyTorch 2.1-2.2 (macOS Intel) require NumPy 1.x (not 2.x)
+    "numpy<2.0; platform_machine == 'x86_64' and sys_platform == 'darwin'",
```
p-e-w, line 33: Do these dependency specs automatically install the right version on all Mac platforms (both Intel and M series)?

I have no experience with Macs so I'm a little out of my depth here.

src/heretic/model.py
```diff
@@ -49,15 +49,46 @@ def __init__(self, settings: Settings):
 
         self.model = None
 
-        for dtype in settings.dtypes:
-            print(f"* Trying dtype [bold]{dtype}[/]... ", end="")
+        # Filter dtypes: MPS doesn't support bfloat16, so skip "auto" on MPS
+        # (since "auto" typically resolves to bfloat16)
```
p-e-w, line 54: Doesn't that happen automatically when the MPS backend is used? It would be strange if a backend tried to load a format it doesn't support.

p-e-w: What happens if you use `auto` on MPS?

pyproject.toml
noahvandal: That should be incidental, doesn't matter as they are all pointing to the same version 

    # macOS Silicon (ARM64): Can use newer versions with Python 3.12
    "torch>=2.2.0; platform_machine == 'arm64' and sys_platform == 'darwin'",
    # Other platforms (Linux, Windows): Can use newer versions
    "torch>=2.2.0; sys_platform != 'darwin'",

src/heretic/model.py
noahvandal: It did not for mine. I believe it is because it *technically* is using MPS but it is more of an integrated graphics (Intel UHD 630) instead of an actual Metal backend. On any Apple Silicon Mac I do not believe that error would propagate however I do not have an M series chip.

src/heretic/model.py
```diff
@@ -49,15 +49,46 @@ def __init__(self, settings: Settings):
...
+        dtypes_to_try = settings.dtypes
+        use_mps = torch.backends.mps.is_available()
+        if use_mps and "auto" in dtypes_to_try:
+            dtypes_to_try = [dtype for dtype in dtypes_to_try if dtype != "auto"]
+            if not dtypes_to_try:
+                # If only "auto" was specified, default to float16 for MPS
+                dtypes_to_try = ["float16"]
```
p-e-w, line 64: That's a lot of magic, and might make it difficult for the user to understand why problems happen if they explicitly specified a dtype cascade and the program just does something else.

noahvandal: Fair points. I made changes so this should only affect x86_64 CPU trying to use MPS, so that it only affects a small subset of users using a non-gpu Mac

src/heretic/model.py
```diff
@@ -49,15 +50,49 @@ def __init__(self, settings: Settings):
...
+                        low_cpu_mem_usage=False,  # Ensure full conversion
+                    )
+                    # Convert to desired dtype explicitly (this forces conversion)
+                    self.model = self.model.to(dtype=torch_dtype)
+                    # Move to MPS device
+                    self.model = self.model.to("mps")
+                else:
+                    self.model = AutoModelForCausalLM.from_pretrained(
+                        settings.model,
+                        torch_dtype=torch_dtype,
```
p-e-w: This triggers a deprecation warning with newer Transformers versions. The argument name is just `dtype` now.

src/heretic/model.py
```diff
@@ -49,15 +50,49 @@ def __init__(self, settings: Settings):
...
+        for dtype_str in dtypes_to_try:
+            print(f"* Trying dtype [bold]{dtype_str}[/]... ", end="")
 
             try:
-                self.model = AutoModelForCausalLM.from_pretrained(
-                    settings.model,
-                    dtype=dtype,
-                    device_map=settings.device_map,
-                )
+                # Convert dtype string to torch dtype object
```
p-e-w: Why is this necessary? I'm pretty sure torch accepts both strings and objects as arguments, and performs the conversion automatically.

src/heretic/model.py
```diff
@@ -49,15 +50,43 @@ def __init__(self, settings: Settings):
...
-                    dtype=dtype,
-                    device_map=settings.device_map,
-                )
+                # For MPS on Intel x86_64, load to CPU first then convert dtype to avoid bfloat16 issues
+                if use_mps and is_intel_x86_64:
+                    # Load to CPU first without dtype to avoid bfloat16 preservation
+                    # Then convert dtype explicitly, then move to MPS
+                    self.model = AutoModelForCausalLM.from_pretrained(
+                        settings.model,
+                        torch_dtype=None,  # Load as-is first
```
p-e-w: Argument name is still wrong. `torch_dtype` => `dtype`, otherwise a warning is raised on recent Transformer versions. Same thing below.

src/heretic/model.py
```diff
@@ -49,15 +50,43 @@ def __init__(self, settings: Settings):
...
+        use_mps = torch.backends.mps.is_available()
+        is_intel_x86_64 = platform.machine() == "x86_64"
+        if use_mps and is_intel_x86_64 and "auto" in dtypes_to_try:
+            dtypes_to_try = [dtype for dtype in dtypes_to_try if dtype != "auto"]
+            if not dtypes_to_try:
+                # If only "auto" was specified, default to float16 for MPS
+                dtypes_to_try = ["float16"]
+
+        for dtype in dtypes_to_try:
+            print(f"* Trying dtype [bold]{dtype_str}[/]... ", end="")
```
p-e-w, line 67:
```suggestion
            print(f"* Trying dtype [bold]{dtype}[/]... ", end="")
```

This will raise an error otherwise.

src/heretic/model.py
```diff
@@ -49,15 +50,43 @@ def __init__(self, settings: Settings):
...
-                self.model = AutoModelForCausalLM.from_pretrained(
-                    settings.model,
-                    dtype=dtype,
-                    device_map=settings.device_map,
-                )
+                # For MPS on Intel x86_64, load to CPU first then convert dtype to avoid bfloat16 issues
+                if use_mps and is_intel_x86_64:
+                    # Load to CPU first without dtype to avoid bfloat16 preservation
+                    # Then convert dtype explicitly, then move to MPS
+                    self.model = AutoModelForCausalLM.from_pretrained(
```
p-e-w, line 74: The same thing needs to happen in `reload_model` below, which is called on every trial.

src/heretic/model.py
```diff
@@ -49,15 +50,43 @@ def __init__(self, settings: Settings):
...
+                        settings.model,
+                        dtype=None,  # Load as-is first
+                        device_map="cpu",  # Load to CPU first
+                        low_cpu_mem_usage=False,  # Ensure full conversion
+                    )
+                    # Convert to desired dtype explicitly (this forces conversion)
```
p-e-w, line 80: What exactly is the difference between this and just specifying the dtype directly when loading the model?

p-e-w:
Please rebase on top of master to run CI.

p-e-w:

PR #15, opened 2025-11-17
perf: empty cache after using residuals and between trials

red40maxxer:
Great work btw, this is an awesome project and looking forward to seeing it grow, I'm still learning and this has been a fun repo to hack around in :) 

I'm pretty sure we can free the residuals as soon as the refusal directions have been computed, and we can also clear the cache between trials.

Tested on my laptop 4060 while abliterating Qwen3-0.6B and peak GPU usage saw a slight decline

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -39,6 +39,7 @@ from .utils import (
     get_trial_parameters,
     load_prompts,
     print,
+    empty_cache,
 )
 
 
@@ -193,6 +194,9 @@ def run():
         p=2,
         dim=1,
     )
+    # we don't need the residuals after computing refusal directions
+    del good_residuals, bad_residuals
+    empty_cache()
 
     trial_index = 0
     start_time = time.perf_counter()


src/heretic/main.py
```diff
@@ -279,7 +283,8 @@ def objective(trial: Trial) -> tuple[float, float]:
         model.abliterate(refusal_directions, direction_index, parameters)
         print("* Evaluating...")
         score, kl_divergence, refusals = evaluator.get_score()
-
+        # free memory between trials
+        empty_cache()
```
p-e-w: This is done by `reload_model()` already, which happens at the start of each trial. I'd be very surprised if that line makes a difference because we're literally milliseconds away from calling that function anyway.

red40maxxer: oops you're right, i'll remove that line

p-e-w: Can you retest and check if the performance gain is still there with just the residuals garbage collected?

red40maxxer: Memory usage without residual gc:
```
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.0.1
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░

GPU type: NVIDIA GeForce RTX 4060 Laptop GPU

Loading model Qwen/Qwen3-0.6B...
* Trying dtype auto... Ok
* Transformer model with 28 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

Loading good prompts from mlabonne/harmless_alpaca...
* 400 prompts loaded

Loading bad prompts from mlabonne/harmful_behaviors...
* 400 prompts loaded

Loading good evaluation prompts from mlabonne/harmless_alpaca...
* 100 prompts loaded

Loading bad evaluation prompts from mlabonne/harmful_behaviors...
* 100 prompts loaded
* Counting model refusals...
* Initial refusals: 52/100

Calculating per-layer refusal directions...
GPU memory before residuals: 1261976064 (peak so far: 3135785472)
* Obtaining residuals for good prompts...
* Obtaining residuals for bad prompts...
GPU memory after residuals: 1358445056 (peak so far: 3135785472)
GPU memory after clearing residuals: 1358563840 (peak so far: 3135785472)
```
With GC:
```
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.0.1
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░

GPU type: NVIDIA GeForce RTX 4060 Laptop GPU

Loading model Qwen/Qwen3-0.6B...
* Trying dtype auto... Ok
* Transformer model with 28 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

Loading good prompts from mlabonne/harmless_alpaca...
* 400 prompts loaded

Loading bad prompts from mlabonne/harmful_behaviors...
* 400 prompts loaded

Loading good evaluation prompts from mlabonne/harmless_alpaca...
* 100 prompts loaded

Loading bad evaluation prompts from mlabonne/harmful_behaviors...
* 100 prompts loaded
* Counting model refusals...
* Initial refusals: 52/100

Calculating per-layer refusal directions...
GPU memory before residuals: 1261976064 (peak so far: 3135785472)
* Obtaining residuals for good prompts...
* Obtaining residuals for bad prompts...
GPU memory after residuals: 1358445056 (peak so far: 3135785472)
GPU memory after clearing residuals: 1262094848 (peak so far: 3135785472)
```
So we save ~100MB by clearing the residuals but it doesn't reduce the peak usage, it would scale with number of prompts I think?

src/heretic/main.py
```diff
@@ -279,7 +283,6 @@ def objective(trial: Trial) -> tuple[float, float]:
         model.abliterate(refusal_directions, direction_index, parameters)
         print("* Evaluating...")
         score, kl_divergence, refusals = evaluator.get_score()
-
```
p-e-w, line 284: :fire:

p-e-w:
Thanks, this is a reasonable change!

Issue #13, opened 2025-11-17
Support GB10 / DGX Spark Hardware.


kwilliams312:
spark-7420:~/dev$ heretic Qwen/Qwen3-4B-Instruct-2507
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.0.1
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░
▀░▀░▀▀▀░▀░▀░▀▀▀░░▀░░▀░▀▀▀  https://github.com/p-e-w/heretic

No GPU or other accelerator detected. Operations will be slow.

Loading model Qwen/Qwen3-4B-Instruct-2507...
Loading checkpoint shards: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:00<00:00, 140.03it/s]
Ok
* Transformer model with 36 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

spark-7420:~/dev$ nvidia-smi
Mon Nov 17 08:21:42 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05              Driver Version: 580.95.05      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GB10                    On  |   0000000F:01:00.0 Off |                  N/A |
| N/A   35C    P8              3W /  N/A  | Not Supported          |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

p-e-w:
PRs welcome, I don't have access to such a device.

runbgp:
I was able to get this working on a DGX Spark.  `uv` was installing `PyTorch 2.9.1+cpu` (CPU-only version) by default because PyTorch isn't explicitly listed in `pyproject.toml`. Without CUDA support, `torch.cuda.is_available()` returned `False`.

I think the cleanest solution is for users to manually install PyTorch with CUDA support matching their driver version before installing heretic.

[...]

PR #16, opened 2025-11-17
docs: Add instructions for setting up PyTorch with NVIDIA hardware

runbgp:
Related to https://github.com/p-e-w/heretic/issues/13#issuecomment-3542746440

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,81 @@ You can find a collection of models that have been decensored using Heretic
 [on Hugging Face](https://huggingface.co/collections/p-e-w/the-bestiary).
 
 
+## Installation
+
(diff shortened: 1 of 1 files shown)


p-e-w:
I appreciate the effort, but that's way too much detail for the README. Information on how to install an appropriate PyTorch version is available in a million places. There is no need to reproduce this for Heretic specifically.

Also, for many CUDA devices there is actually no need to specify an index URL and Heretic will work as expected without doing anything.

runbgp:
Appreciate the feedback, that sounds good to me! It seems like this may be more Spark specific, so folks can just reference that issue if need be.

PR #17, opened 2025-11-17
Fix multi-GPU support and memory management

### LINKED ISSUE #8: gpt-oss-120b "killed" before it loads
Hi!
I noticed that one of your abilterated models is gpt-oss-20b. That encouraged me to go ahead and try to abliterate it's bigger, 120b brother. Unfortunately it doesn't seem to work, the process gets killed when loading reaches 20%:
```
heretic models/unsloth-gpt-oss-120b/
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.0.1
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░

GPU type: NVIDIA A100 80GB PCIe

Loading model models/unsloth-gpt-oss-120b/...
Loading checkpoint shards:  20%|███████▌                              | 3/15 [00:34<02:05, 10.50s/it]Killed
```

I've downloaded the model twice just to be sure there's no corruption. I tried several times, every time it gets killed at 20%. Any idea what i might be doing wrongly? Thank you so much!

kldzj:
## Problem
- Device mismatch errors: `RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cuda:1, different from other tensors on cuda:0`
- Out-of-memory errors after 1-2 trials despite reloading the model

## Solution
- Move projectors to correct device before each matrix operation
- Add `gc.collect()` + `empty_cache()` before model reloads and after trials

## Result
Models with `device_map="auto"` now run successfully across multiple GPUs without device mismatches or memory accumulation between trials.

Successfully abliterated gpt-oss-120b on 8x B200:
- [kldzj/gpt-oss-120b-heretic](https://huggingface.co/kldzj/gpt-oss-120b-heretic) (MXFP4)
- [kldzj/gpt-oss-120b-heretic-bf16](https://huggingface.co/kldzj/gpt-oss-120b-heretic-bf16) (BF16)
- [kldzj/gpt-oss-120b-heretic-v2](https://huggingface.co/kldzj/gpt-oss-120b-heretic-v2) (MXFP4, lower KL divergence)
- [kldzj/gpt-oss-120b-heretic-v2-bf16](https://huggingface.co/kldzj/gpt-oss-120b-heretic-v2-bf16) (BF16, lower KL divergence)

Also confirmed to be working using 4x H200.

---

Fixes #8

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -48,6 +48,8 @@ def batchify(items: list[T], batch_size: int) -> list[list[T]]:
 
 
 def empty_cache():
+    gc.collect()
+
     if torch.cuda.is_available():
         torch.cuda.empty_cache()
     elif is_xpu_available():
@@ -61,7 +63,6 @@ def empty_cache():
     elif torch.backends.mps.is_available():
         torch.mps.empty_cache()
 
-    gc.collect()
     gc.collect()
 
 


src/heretic/main.py
```diff
@@ -295,6 +300,9 @@ def objective(trial: Trial) -> tuple[float, float]:
 
         trial.set_user_attr("kl_divergence", kl_divergence)
         trial.set_user_attr("refusals", refusals)
+        
+        gc.collect()
```
p-e-w: `empty_cache()` calls `gc.collect()`, so this is redundant.

src/heretic/main.py
```diff
@@ -43,6 +45,11 @@
 
 
 def run():
+    # Enable expandable segments to reduce memory fragmentation on multi-GPU setups.
+    import os
```
p-e-w: Imports at the top please.

src/heretic/model.py
```diff
@@ -83,10 +83,13 @@ def __init__(self, settings: Settings):
             )
 
     def reload_model(self):
+        import gc
+        
         dtype = self.model.dtype
 
         # Purge existing model object from memory to make space.
         self.model = None
+        gc.collect()
```
p-e-w: See other comment.

src/heretic/model.py
```diff
@@ -198,14 +201,26 @@ def abliterate(
 
                 # Projects any right-multiplied vector(s) onto the subspace
                 # spanned by the refusal direction.
-                projector = torch.outer(
+                # Create projector on CPU first to avoid device-specific memory allocation.
```
p-e-w: Why is this on the CPU? Torch operations generally preserve the device, so the output should be on whatever device `layer_refusal_direction` is on, which should be the GPU as it descends from the hidden states obtained by GPU operations on the GPU-loaded model.

kldzj: You're right, my bad

kldzj:
@p-e-w I made the requested changes and did some more testing. The total memory usage is now much more in line with what you'd expect (model roughly 220 GB plus whatever batch size overhead can fit in the remaining memory).

src/heretic/model.py
kldzj: Turns out this one is required. I tested multiple times with and without this `gc.collect` and I find it to be essential to not get OOMs.

src/heretic/model.py
```diff
@@ -86,7 +87,12 @@ def reload_model(self):
         dtype = self.model.dtype
 
         # Purge existing model object from memory to make space.
+        del self.model
```
p-e-w: Why is this necessary if the reference is overwritten on the next line?

src/heretic/model.py
```diff
@@ -86,7 +87,12 @@ def reload_model(self):
...
         self.model = None
+        
+        # Force garbage collection and clear cache
+        # to ensure all references are released
+        gc.collect()
```
p-e-w: As mentioned earlier, `empty_cache()` (next line) already calls `gc.collect()`, so this is redundant unless I'm missing something.

src/heretic/model.py
```diff
@@ -202,11 +208,14 @@ def abliterate(
                     layer_refusal_direction,
                     layer_refusal_direction,
                 ).to(self.model.dtype)
-
+
```
p-e-w: :fire:

src/heretic/model.py
```diff
@@ -202,11 +208,14 @@ def abliterate(
                     layer_refusal_direction,
                     layer_refusal_direction,
                 ).to(self.model.dtype)
-
+                
                 for matrix in matrices:
                     # In-place subtraction is safe as we're not using Autograd.
```
p-e-w: This comment should be moved down so it remains attached to the line it refers to (`matrix.sub_`).

src/heretic/model.py
```diff
@@ -202,11 +208,14 @@ def abliterate(
...
-                    matrix.sub_(weight * (projector @ matrix))
-
+                    # Ensure projector is on the same device as the matrix for multi-GPU support.
+                    device_projector = projector.to(matrix.device)
+                    matrix.sub_(weight * (device_projector @ matrix))
+                    # Delete device-specific projector immediately to free memory
+                    del device_projector
```
p-e-w: This also shouldn't be necessary, because either we loop and immediately overwrite `device_projector`, releasing the previous reference just like `del` does, or the method returns and `device_projector` goes out of scope, also releasing the reference.

Note that in Python, `del x` does **not** call `x.__del__()`. The `__del__` method (which libraries like PyTorch use to clean up resources) is only called on object destruction, and the GC doesn't care whether a reference is released with `del` or by overwriting it with another value. If you want to make sure memory is released immediately, you have to follow up `del` with `gc.collect()`. But since you got this working without it and `gc.collect()` is already called on each trial anyway, there's probably no need for this line at all.

src/heretic/model.py
kldzj: It's definitely doing something. Why it's necessary to call it before calling the pytorch purge methods I can't say for certain, but without garbage collection beforehand I'm getting OOM errors.

src/heretic/model.py
kldzj: It was a desperate attempt and was kept in because it was working with it. I removed it for now and will test again, but yes, probably it isn't needed and is doing nothing.

kldzj: It seems to be working just fine without the additional variable deletion.

kldzj:
Thanks for looking over the PR again @p-e-w. I addressed most of your comments. As mentioned above, I'm not entirely sure why the additional garbage collection beforehand is necessary, but it's definitely making a difference as without it I'm consistently running out of memory.

src/heretic/main.py
```diff
@@ -300,6 +306,8 @@ def objective(trial: Trial) -> tuple[float, float]:
         trial.set_user_attr("kl_divergence", kl_divergence)
         trial.set_user_attr("refusals", refusals)
 
+        empty_cache()
```
p-e-w: Is this one also necessary? Because that literally happens milliseconds later, when `model.reload_model()` is called.

src/heretic/main.py
```diff
@@ -312,6 +320,9 @@ def objective(trial: Trial) -> tuple[float, float]:
     )
 
     study.optimize(objective, n_trials=settings.n_trials)
+    
+    # Final cleanup after all trials
+    empty_cache()
```
p-e-w: What happens without this one? This also shouldn't be necessary because when we reload a model below, the same function is again called.

src/heretic/model.py
p-e-w: :confused:

p-e-w:
I'm at a loss for what to do here. The device mapping is obviously correct, but the cache management and GC invocations feel like random lines sprinkled through the code, *none* of which should be necessary and at least one of which seems outright illogical. I'm highly uncomfortable merging code I don't fully understand, because I worry that this might be masking a bug somehow.

p-e-w:
Okay, I just did a deep-ish dive on Python memory management and I have figured out the problem:

**In Python, `gc.collect()` is not an idempotent operation.** Calling it multiple times in a row is not guaranteed to be equivalent to calling it once. This is not a bug but a consequence of how the garbage collector works. Calling `gc.collect()` doesn't reap objects with finalizers (`__del__`), it moves them to an internal garbage list.

It gets even crazier: **There is no number `n` for which it is guaranteed that calling `gc.collect()` `n` times will collect all garbage that can be collected in theory.** That's because finalizers can create new references and there is no limit to how deep that recursion can go. This is called "resurrection".

PyTorch is apparently a well-known offender in that regard, and it's supposedly "common practice" (though new to me) to just call `gc.collect()` more than once in a row.

Please try the following:

* Duplicate the `gc.collect()` line in the `empty_cache()` function
* Remove all calls to `empty_cache()` and `gc.collect()` that you added
* Then try running it again on your setup
* If it still leaks memory, add an additional copy of `gc.collect()` to `empty_cache()` etc.

kldzj:
> Please try the following:
> 
> * Duplicate the `gc.collect()` line in the `empty_cache()` function
> * Remove all calls to `empty_cache()` and `gc.collect()` that you added
> * Then try running it again on your setup
> * If it still leaks memory, add an additional copy of `gc.collect()` to `empty_cache()` etc.

Sounds reasonable. I removed everything you mentioned and you're correct that the added `empty_cache()` calls weren't needed.

But: when first calling `torch.x.empty_cache()` and then calling `gc.collect()` twice it's still running out of memory. When calling `gc.collect()` before and after the `torch.x.empty_cache()` call, it's working as expected.

p-e-w:
Merged! Thank you for your patience in figuring this out. I can't say for certain what causes the behavior you observed, but then again `torch.x.empty_cache()` is yet another black hole where all kinds of magic happen, and I believe some backends even have their own garbage collection mechanisms.

Issue #18, opened 2025-11-18
Qwen3 32b VL support?


phr00t:
Perhaps related to https://github.com/p-e-w/heretic/pull/11 ?

torch-2.9.1+cu130

Name: transformers
Version: 4.57.1

`phr00t@AI-BOT /cygdrive/d/Heretic
$ heretic Qwen/Qwen3-VL-32B-Instruct
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.0.1
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░
▀░▀░▀▀▀░▀░▀░▀▀▀░░▀░░▀░▀▀▀  https://github.com/p-e-w/heretic

GPU type: NVIDIA GeForce RTX 5070 Ti

Loading model Qwen/Qwen3-VL-32B-Instruct...
tokenizer_config.json: 10.9kB [00:00, ?B/s]
C:\Users\phr00t\AppData\Roaming\Python\Python312\site-packages\huggingface_hub\file_download.py:143: UserWarning: `huggingface_hub` cache-system uses symlinks by default to efficiently store duplicated files but your machine does not support them in C:\Users\phr00t\.cache\huggingface\hub\models--Qwen--Qwen3-VL-32B-Instruct. Caching files will still work but in a degraded version that might require more space on your disk. This warning can be disabled by setting the `HF_HUB_DISABLE_SYMLINKS_WARNING` environment variable. For more details, see https://huggingface.co/docs/huggingface_hub/how-to-cache#limitations.
To support symlinks on Windows, you either need to activate Developer Mode or to run Python as an administrator. In order to activate developer mode, see this article: https://docs.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development
  warnings.warn(message)
vocab.json: 2.78MB [00:00, 11.4MB/s]
merges.txt: 1.67MB [00:00, 12.2MB/s]
tokenizer.json: 7.03MB [00:00, 18.9MB/s]
config.json: 1.47kB [00:00, ?B/s]
Failed (Unrecognized configuration class <class 'transformers.models.qwen3_vl.configuration_qwen3_vl.Qwen3VLConfig'> for this
kind of AutoModel: AutoModelForCausalLM.
Model type should be one of ApertusConfig, ArceeConfig, AriaTextConfig, BambaConfig, BartConfig, BertConfig,
[...]

PR #22, opened 2025-11-19
Added: Notebook compatibility, Multimodal support (InternVL/QwenVL/Gemma), and trust_remote_code

Vinay-Umrethe:
### Key Changes

1.  **Notebook Compatibility**:
    *   Replaced `questionary` arrow-key menus with robust numeric input loops. This fixes "CPR" warnings and usability issues in non-TUI environments like Google Colab.
    *   Added a `/exit` command to the chat loop for graceful termination. Returing to the menu.

2.  **Multimodal Model Support**:
    *   Implemented dynamic model class loading in `model.py`.
    *   Added support for:
 
| Model Family | Example Models (Hugging Face ID) |
| :--- | :--- |
| **Qwen 2.5 VL** | `Qwen/Qwen2.5-VL-7B-Instruct`, `Qwen/Qwen2.5-VL-3B-Instruct` |
| **Qwen 3 VL** | `Qwen/Qwen3-VL-4B-Instruct`, `Qwen/Qwen3-VL-2B-Instruct` |
| **Qwen 3 VL MoE** | `Qwen/Qwen3-VL-235B-A22B-Instruct`, `Qwen/Qwen3-VL-30B-A3B-Instruct` |
| **Qwen 2.5 Omni** | `Qwen/Qwen2.5-Omni-7B`, `Qwen/Qwen2.5-Omni-3B` |
| **Qwen 3 Omni** | `Qwen/Qwen3-Omni-30B-A3B-Instruct`, `Qwen/Qwen3-Omni-30B-A3B-Thinking` |
| **InternVL 3.5** | `OpenGVLab/InternVL3_5-8B-HF`, `OpenGVLab/InternVL3_5-2B-HF` |
| **Gemma 3 / 3n** | `google/gemma-3-4b-it`, `google/gemma-3n-4b-it` |

- Expanded `get_layers()` to handle nested structures in VL models.

3.  **Custom Code Support**:
    *   Added `--trust-remote-code` flag (default: `False`) to allow loading models with custom modeling code (required for InternVL, etc.).

4.  **Robustness**:
    *   Added `bfloat16` to the default dtype list.
    *   Implemented generation fallbacks: retries with `pixel_values=None` or dummy tensors to handle picky VL models that crash on text-only inputs.

5.  **Upstream Sync**:
    *   Manually applied recent upstream fixes for multi-GPU support, memory optimization, and the Hugging Face profile `KeyError`. 
    * Needed a Manual update for this, since these fixes from your side, were implmented Today `(Nov 19, 2025, 5:32 AM GMT+5:30)`  just 1-2 hours after I enhanced it. DONE ✅

Tested successfully on Google Colab with `OpenGVLab/InternVL3_5-2B-HF` , `Qwen/Qwen2.5-Omni-3B`, `Qwen3-VL-2B-Insturct` and `Qwen/Qwen2.5-VL-3B-Instruct`.

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -28,11 +28,6 @@ class Settings(BaseSettings):
         description="If this model ID or path is set, then instead of abliterating the main model, evaluate this model relative to the main model.",
     )
 
-    trust_remote_code: bool = Field(
-        default=False,
-        description="Whether to allow the model to execute custom code.",
-    )
-
     dtypes: list[str] = Field(
         default=[
             # In practice, "auto" almost always means bfloat16.
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -28,6 +28,11 @@ class Settings(BaseSettings):
         description="If this model ID or path is set, then instead of abliterating the main model, evaluate this model relative to the main model.",
     )
 
+    trust_remote_code: bool = Field(
+        default=False,
+        description="Whether to allow the model to execute custom code.",
+    )
+
     dtypes: list[str] = Field(
         default=[
             # In practice, "auto" almost always means bfloat16.
@@ -35,6 +40,7 @@ class Settings(BaseSettings):
             # If that doesn't work (e.g. on pre-Ampere hardware), fall back to float16.
             "float16",
             # If that still doesn't work (e.g. due to https://github.com/meta-llama/llama/issues/380),
+            "bfloat16",
             # fall back to float32.
             "float32",
         ],
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -12,7 +12,6 @@ import transformers
 from torch import LongTensor, Tensor
 from torch.nn import ModuleList
 from transformers import (
-    AutoModelForCausalLM,
     AutoTokenizer,
     BatchEncoding,
     PreTrainedTokenizerBase,
@@ -93,16 +92,16 @@ class Model:
                         device_map=settings.device_map,
                         trust_remote_code=settings.trust_remote_code,
                     )
-                    
+
                     # If successful, cache the class and break the loop
                     self.model_class = model_cls
                     break
                 except Exception:
                     continue
-            
+
             if self.model is None:
                 empty_cache()
-                print(f"[red]Failed[/] (Could not load with any known model class)")
+                print("[red]Failed[/] (Could not load with any known model class)")
                 continue
 
             try:
@@ -149,18 +148,18 @@ class Model:
         # Most multimodal models (Qwen VL, etc) often nest the LLM under .language_model
         with suppress(Exception):
             return self.model.language_model.model.layers
-            
+
         with suppress(Exception):
             return self.model.language_model.layers
 
         # Text-only models.
         with suppress(Exception):
             return self.model.model.layers
-            
+
         # Fallback for some other architectures
         with suppress(Exception):
             return self.model.layers
-            
+
         raise Exception("Could not locate transformer layers in the model.")
 
     def get_layer_matrices(self, layer_index: int) -> dict[str, list[Tensor]]:
@@ -306,7 +305,9 @@ class Model:
 
         return inputs, self._generate_with_fallbacks(inputs, **kwargs)
 
-    def _generate_with_fallbacks(self, inputs: BatchEncoding, **kwargs) -> GenerateOutput | LongTensor:
+    def _generate_with_fallbacks(
+        self, inputs: BatchEncoding, **kwargs
+    ) -> GenerateOutput | LongTensor:
         # Standard generation
         try:
             return self.model.generate(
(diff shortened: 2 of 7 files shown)


p-e-w:
This PR contains all kinds of random changes that don't belong there:

* Rewrites of random sections of the README for no reason
* Version bump
* Change of package name
* Randomly deleted entire chunks from the code, such as MPS support and comments

I don't mind contributors using LLMs **as a tool** for writing code, but just telling an LLM "make this change" and expecting to end up with a useful pull request doesn't fly. As it stands, this PR doesn't even meet the quality threshold for a review, nevermind a merge.

src/heretic/config.py
```diff
@@ -28,13 +28,19 @@ class Settings(BaseSettings):
...
+    )
+
     dtypes: list[str] = Field(
         default=[
             # In practice, "auto" almost always means bfloat16.
             "auto",
             # If that doesn't work (e.g. on pre-Ampere hardware), fall back to float16.
             "float16",
             # If that still doesn't work (e.g. due to https://github.com/meta-llama/llama/issues/380),
+            "bfloat16",
```
p-e-w, line 38: What problem does this solve? Does it fix any particular model?

src/heretic/model.py
```diff
@@ -34,39 +35,85 @@ class AbliterationParameters:
...
+                    trust_remote_code=settings.trust_remote_code,
+                )
+            except Exception as error:
+                raise Exception(f"Failed to load tokenizer or processor: {error}")
 
         # Fallback for tokenizers that don't declare a special pad token.
         if self.tokenizer.pad_token is None:
             self.tokenizer.pad_token = self.tokenizer.eos_token
-            self.tokenizer.padding_side = "left"
+            if hasattr(self.tokenizer, "padding_side"):
```
p-e-w, line 57: For which model is this needed?

src/heretic/model.py
```diff
@@ -89,19 +136,31 @@ def reload_model(self):
...
+
+        with suppress(Exception):
+            return self.model.language_model.layers
 
         # Text-only models.
-        return self.model.model.layers
+        with suppress(Exception):
+            return self.model.model.layers
+
+        # Fallback for some other architectures
```
p-e-w, line 150: Which architectures are those?

src/heretic/utils.py
```diff
@@ -1,6 +1,3 @@
-# SPDX-License-Identifier: AGPL-3.0-or-later
-# Copyright (C) 2025  Philipp Emanuel Weidmann <pew@worldwidemann.com>
-
```
p-e-w, line 3: Don't remove attribution.

p-e-w (commented):
Okay, I took a closer look now. This PR consists of two unrelated parts (and should therefore be two PRs).

&nbsp;

The first part is changes to make the program work in environments that aren't terminal emulators (such as notebooks with `!` commands). Simply put, those changes are not going to be merged, neither now nor in the future. They degrade the experience for terminal users, in some cases significantly. For example, you can no longer edit chat messages properly, you lose path completion for directory selection, secret tokens are no longer masked, and the list selection is more cumbersome as you now have to type numbers instead of just selecting with the arrow keys.

I'm not willing to sacrifice even a tiny bit of usability for terminal users just so Heretic can be run in environments it wasn't designed for. This stuff belongs in a fork. The `trust_remote_code` configuration parameter is also unnecessary, as in a terminal environment Transformers will prompt for that interactively when loading a model that requires it, which is a far better user experience than having the program crash and then needing to relaunch with an unsafe argument.

&nbsp;

The second part is improvements to multimodal support. Gemma 3 actually works for me already, and is one of the example models I posted on Hugging Face. What problems did you encounter with it? As for Qwen3 VL, this is being discussed in #18, and it seems that the lack of auto loading support was an oversight in Transformers that might be fixed in an upcoming version.

The class list probably isn't the right solution. It would grow without bounds in the future. Having individual model implementations on there feels wrong, as the whole point of Transformers is to abstract over different models, and this *does* work even for multimodal models, such as Gemma 3. Instead of introducing complicated fallback cascades, we should figure out what the actual problem is with individual models and why some of them don't work by default with `AutoModelForCausalLM`.

I'm hesitant to introduce complexity into model handling in general, and would prefer to tackle problems with individual models one by one rather than trying to come up with a solution that will work for everything. If possible, please file issues describing which problems you have encountered for specific multimodal models. Currently, I don't understand the scope of the multimodal space well enough to decide how to proceed.

p-e-w:
Now that's a valuable change indeed! I'm certainly happy to merge something that adds notebook support *without* changing the TUI for terminal users.

Please do the following:

1. Move the `ui_*` helper methods to `utils.py` so they don't clutter up main. I would also suggest renaming them to `prompt_text`, `prompt_password` etc. because I think that more accurately reflects what they do.
2. Open a new pull request containing only the notebook-related changes.

Huge:
Mind noting the reason for closing this one, I suppose another PR had the code or you've merged this locally..?

p-e-w:
Notebook support was merged in #42.

p-e-w:
`trust_remote_code` was merged in #31.

PR #26, opened 2025-11-19
Jp localization

Utakata:
Jp localization


p-e-w:
Sorry, but I'm not going to change Heretic's language to Japanese.

If you want a language change, you have to maintain a fork. This is not going into mainline.

PR #29, opened 2025-11-20
Add support for local dataset paths in load_prompts

tymat:
This change allows load_prompts() to work with both HuggingFace dataset IDs and local directory paths containing datasets saved in Arrow format.

Changes:
- Import load_from_disk from datasets library
- Add directory detection logic to load_prompts()
- Handle local datasets with proper split parsing (e.g., "train[:400]")
- Maintain backward compatibility with HuggingFace dataset IDs

Benefits:
- Users can work with custom/modified datasets stored locally
- Enables offline usage with pre-downloaded datasets
- Fully backward compatible - existing configs continue to work
- Supports the same split syntax for both local and remote datasets

The function now:
1. Checks if the dataset path is a local directory
2. If yes, uses load_from_disk() and parses split syntax manually
3. If no, falls back to the original load_dataset() behavior

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -109,6 +109,11 @@ class Model:
         matrices = {}
 
         def try_add(component: str, matrix: Any):
+            # Handle Triton tensors (e.g., from MXFP4 quantization) by extracting
+            # the underlying PyTorch tensor via the .data attribute.
+            if hasattr(matrix, "data") and torch.is_tensor(matrix.data):
+                matrix = matrix.data
+
             assert torch.is_tensor(matrix)
 
             if component not in matrices:


src/heretic/utils.py
```diff
@@ -36,7 +36,31 @@ def format_duration(seconds: float) -> str:
 
 
 def load_prompts(specification: DatasetSpecification) -> list[str]:
-    dataset = load_dataset(specification.dataset, split=specification.split)
+    import os
```
p-e-w, line 39: Imports to the top please.

src/heretic/utils.py
```diff
@@ -36,7 +36,31 @@ def format_duration(seconds: float) -> str:
...
+    if os.path.isdir(specification.dataset):
+        # Load from local disk (Arrow format)
+        dataset_dict = load_from_disk(specification.dataset)
+        # Extract the split from the specification (e.g., "train[:400]")
+        split_name = specification.split.split("[")[0] if "[" in specification.split else specification.split
+        dataset = dataset_dict[split_name]
+
+        # Apply slicing if specified (e.g., "train[:400]")
+        if "[" in specification.split:
+            slice_part = specification.split.split("[")[1].rstrip("]")
```
p-e-w, line 51: Probably cleaner to factor out

```python
split_parts = specification.split.split("[")
```

and reuse it on both occasions to avoid splitting twice. The code will also become easier to read.

src/heretic/utils.py
```diff
@@ -36,7 +36,31 @@ def format_duration(seconds: float) -> str:
...
 def load_prompts(specification: DatasetSpecification) -> list[str]:
-    dataset = load_dataset(specification.dataset, split=specification.split)
+    import os
+
+    # Check if dataset path is a local directory
+    if os.path.isdir(specification.dataset):
+        # Load from local disk (Arrow format)
+        dataset_dict = load_from_disk(specification.dataset)
+        # Extract the split from the specification (e.g., "train[:400]")
+        split_name = specification.split.split("[")[0] if "[" in specification.split else specification.split
```
p-e-w, line 46:
```suggestion
        split_name = specification.split.split("[")[0]
```

I believe this can be simplified because IIRC, `str.split` returns a list containing the entire string if the string doesn't contain the separator.

p-e-w (changes requested):
Is parsing the split syntax not exposed as a function by `datasets`?

p-e-w:
According to #33, local datasets should already work. What problems did you run into trying to load a local dataset with the existing code?

p-e-w:
This may have been superseded by #33, which is now merged.

p-e-w:
Superseded by #33.

PR #31, opened 2025-11-20
Add `trust_remote_code` configuration option

kldzj:
## Problem
Some models require running custom code. If we don't specify `trust_remote_code`, the process will be interrupted and we'll be asked everytime the tokenizer or model is reloaded, for whether we want to allow it.

## Solution
Introduce a `--trust-remote-code` flag, which defaults to `False`, and pass it to all `.from_pretrained()` calls.

## Result
The process will now run uninterrupted. If we didn't specify the flag and the model needs to run custom code, it will fail when trying to load the model the first time with a error that specifies that remote code must be trusted.

One minor inconvenience is that the flag needs to explicitly needs to be followed with a boolean value. I didn't want to overcomplicate [the model arg parsing logic](https://github.com/p-e-w/heretic/blob/master/src/heretic/main.py#L60-L68), which in its current iteration prevents us from setting `cli_implicit_flags=True` in the [`SettingsConfigDict`](https://github.com/p-e-w/heretic/blob/master/src/heretic/config.py#L153-L158), meaning `heretic --trust-remote-code [...]` isn't sufficient, instead a boolean value must be passed like `heretic --trust-remote-code true [...]`.

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -66,6 +66,11 @@ class Settings(BaseSettings):
         description="Maximum number of tokens to generate for each response.",
     )
 
+    print_refusal_geometry: bool = Field(
+        default=False,
+        description="Whether to print detailed information about residuals and refusal directions after calculating them.",
+    )
+
     kl_divergence_scale: float = Field(
         default=1.0,
         description=(
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -124,6 +124,11 @@ class Model:
         matrices = {}
 
         def try_add(component: str, matrix: Any):
+            # Handle Triton tensors (e.g., from MXFP4 quantization) by extracting
+            # the underlying PyTorch tensor via the .data attribute.
+            if hasattr(matrix, "data") and torch.is_tensor(matrix.data):
+                matrix = matrix.data
+
             assert torch.is_tensor(matrix)
 
             if component not in matrices:
--- a/config.default.toml
+++ b/config.default.toml
@@ -24,6 +24,9 @@ max_batch_size = 128
 # Maximum number of tokens to generate for each response.
 max_response_length = 100
 
+# Whether to print detailed information about residuals and refusal directions after calculating them.
+print_refusal_geometry = false
+
 # Assumed "typical" value of the Kullback-Leibler divergence from the original model for abliterated models.
 # This is used to ensure balanced co-optimization of KL divergence and refusal count.
 kl_divergence_scale = 1.0
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -13,6 +13,7 @@ import huggingface_hub
 import optuna
 import questionary
 import torch
+import torch.linalg as LA
 import torch.nn.functional as F
 import transformers
 from accelerate.utils import (
@@ -23,12 +24,13 @@ from accelerate.utils import (
     is_xpu_available,
 )
 from huggingface_hub import ModelCard, ModelCardData
-from optuna import Trial
+from optuna import Trial, TrialPruned
 from optuna.exceptions import ExperimentalWarning
 from optuna.samplers import TPESampler
 from optuna.study import StudyDirection
 from pydantic import ValidationError
 from questionary import Choice, Style
+from rich.table import Table
 from rich.traceback import install
 
 from .config import Settings
@@ -46,8 +48,11 @@ from .utils import (
 
 def run():
     # Enable expandable segments to reduce memory fragmentation on multi-GPU setups.
-    if "PYTORCH_CUDA_ALLOC_CONF" not in os.environ:
-        os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+    if (
+        "PYTORCH_ALLOC_CONF" not in os.environ
+        and "PYTORCH_CUDA_ALLOC_CONF" not in os.environ
+    ):
+        os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
 
     # Modified "Pagga" font from https://budavariam.github.io/asciiart-text/
     print(f"[cyan]█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀[/]  v{version('heretic-llm')}")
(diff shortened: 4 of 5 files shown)


kldzj:
I just saw your comment in another PR:

> The trust_remote_code configuration parameter is also unnecessary, as in a terminal environment Transformers will prompt for that interactively when loading a model that requires it, which is a far better user experience than having the program crash and then needing to relaunch with an unsafe argument.

I have to vehemently disagree, the number of times you have to confirm is making the abliteration process anything else than smooth. Maybe the crashing behavior is unsatisfactory, but nowhere close to having to agree over and over again.

p-e-w:
You're right that having to agree every time is a bug. But a command-line flag is not the correct fix. This change creates the following workflow:

1. Run Heretic with some model
2. Heretic crashes while loading the model (because of the default `trust_remote_code=False`)
3. Say "damn, I forgot about that"
4. Re-run Heretic with `--trust_remote_code True`

The workflow we want is:

1. Run Heretic with some model
2. While loading the model, Heretic prompts the user for permission to run remote code
3. The user agrees, and the program proceeds (without ever asking again, of course)

So we must find a way to remember the choice made by the user the first time. There is actually an extremely simple fix that does *almost* the right thing: Just add `trust_remote_code=True` to `reload_model`. That method is only called if the original load succeeded (which requires the user agreeing to execute remote code), so any subsequent loads are safe. But there is a small problem: `reload_model` is also used for `evaluate_model`, when the model path is overwritten, so this would load a model with the flag enabled that the user hasn't agreed to yet. This could be fixed e.g. by passing an optional argument to `reload_model`.

*Edit:* Actually, we can have the CLI flag, but if it is set to `False` the program should prompt and remember the choice, instead of crashing and forcing the user to re-run to enable remote code.

Vinay-Umrethe:
The thing I added was actually useful you mean to say the `--trust-remote-code` flag NOW hmm.

kldzj:
I think this is the cleanest way, by setting it to `None` by default and after the tokenizer loaded to `True` if it previously was `None`.

If the model requires custom code, loading the tokenizer would prompt the user to confirm allowing custom code. If the user declines the program will stop. If the user confirms, each subsequent model load will not prompt again.

Should I still implement this explicit check for loading the `evaluate_model`?

src/heretic/config.py
```diff
@@ -46,6 +46,11 @@ class Settings(BaseSettings):
         description="Device map to pass to Accelerate when loading the model.",
     )
 
+    trust_remote_code: bool | None = Field(
+        default=None,
```
p-e-w, line 50: :+1: That's definitely the correct default.

src/heretic/model.py
```diff
@@ -93,6 +98,7 @@ def reload_model(self):
             self.settings.model,
             dtype=dtype,
             device_map=self.settings.device_map,
+            trust_remote_code=self.settings.trust_remote_code,
```
p-e-w: This isn't safe in this form because of `evaluate_model` (see my earlier comment).

src/heretic/model.py
```diff
@@ -39,9 +39,13 @@ def __init__(self, settings: Settings):
         print(f"Loading model [bold]{settings.model}[/]...")
 
         self.tokenizer: PreTrainedTokenizerBase = AutoTokenizer.from_pretrained(
-            settings.model
+            settings.model,
+            trust_remote_code=self.settings.trust_remote_code,
         )
 
+        if self.settings.trust_remote_code is None:
+            self.settings.trust_remote_code = True
```
p-e-w: We have to be careful here. Is it possible for the tokenizer to *not* require remote code, but for the model itself to require it? If so, this is unsafe, because no prompt would appear when loading the tokenizer, and then we flip it to `True`, loading the main model without a prompt, and the user was never given a choice.

p-e-w (changes requested):

src/heretic/model.py
kldzj: Makes sense, though I'd think this to not be the case, since they're being loaded from the same repository. In the cases I tested it was always asking for both, but I definitely can't say this is true in every case.

kldzj:
Thanks for taking another look. What do you think about this approach @p-e-w?

By the way, I tested your theory and you're definitely correct. If the tokenizer does not have an auto mapping it will not prompt, so the previous iteration would have been unsafe.

src/heretic/model.py
```diff
@@ -48,6 +49,10 @@ def __init__(self, settings: Settings):
             self.tokenizer.padding_side = "left"
 
         self.model = None
+        self.trusted_models = {settings.model: settings.trust_remote_code}
+
+        if self.settings.evaluate_model is not None:
+            self.trusted_models[settings.model] = settings.trust_remote_code
```
p-e-w: I don't understand this line. Doesn't it do exactly the same thing as line 52?

src/heretic/model.py
```diff
@@ -93,8 +104,12 @@ def reload_model(self):
             self.settings.model,
             dtype=dtype,
             device_map=self.settings.device_map,
+            trust_remote_code=self.trusted_models.get(self.settings.model),
```
p-e-w, line 107: Not sure about this. If the user explicitly passes `--trust-remote-code True`, we'd expect this to also extend to the evaluated model, no?

src/heretic/model.py
kldzj: Whoopsie, that should've been the `evaluate_model` obviously

src/heretic/model.py
kldzj: That's what the other line was meant for, although I've referenced the wrong model name in line 55.

kldzj: And `settings.model` is being overridden in [main.py#L192](https://github.com/p-e-w/heretic/blob/master/src/heretic/main.py#L192), so with [model.py#L55](https://github.com/p-e-w/heretic/pull/31/files#diff-d5cf886765fa650e66feb1fd1cf3a9f2f677d6876cc6191f0c0fef5e0ff31bf7R55) corrected this should be working as intended.

src/heretic/config.py
p-e-w, line 50: :white_check_mark: Correct default, equivalent to not passing an argument to `from_pretrained`.

src/heretic/model.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, settings: Settings):
         print(f"Loading model [bold]{settings.model}[/]...")
 
         self.tokenizer: PreTrainedTokenizerBase = AutoTokenizer.from_pretrained(
-            settings.model
+            settings.model,
+            trust_remote_code=settings.trust_remote_code,
```
p-e-w, line 43: :white_check_mark: Completely safe so far, will simply follow the `trust_remote_code` setting, defaulting to `None`, which prompts the user.

src/heretic/model.py
```diff
@@ -48,6 +49,10 @@ def __init__(self, settings: Settings):
             self.tokenizer.padding_side = "left"
 
         self.model = None
+        self.trusted_models = {settings.model: settings.trust_remote_code}
```
p-e-w, line 52: :white_check_mark: Completely safe. This just says "use the setting for the model".

src/heretic/model.py
```diff
@@ -48,6 +49,10 @@ def __init__(self, settings: Settings):
...
+
+        if self.settings.evaluate_model is not None:
+            self.trusted_models[settings.evaluate_model] = settings.trust_remote_code
```
p-e-w, line 55: :white_check_mark: Completely safe. This just says "use the setting for the evaluated model".

src/heretic/model.py
```diff
@@ -57,8 +62,14 @@ def __init__(self, settings: Settings):
                     settings.model,
                     dtype=dtype,
                     device_map=settings.device_map,
+                    trust_remote_code=self.trusted_models.get(settings.model),
```
p-e-w, line 65: Looks up the stored trust value. This is safe assuming said value is correct.

src/heretic/model.py
```diff
@@ -57,8 +62,14 @@ def __init__(self, settings: Settings):
...
                 )
 
+                # If we reach this point and the model requires trust_remote_code,
+                # the user must have confirmed it.
```
p-e-w, line 69: (or `settings.trust_remote_code` was set to `True`)

src/heretic/model.py
```diff
@@ -57,8 +62,14 @@ def __init__(self, settings: Settings):
...
+                if self.trusted_models.get(settings.model) is None:
+                    self.trusted_models[settings.model] = True
```
p-e-w, line 71: :exclamation: :white_check_mark: This is the critical part, and the comment correctly describes why it works. Declining the prompt for a model requiring `trust_remote_code` will raise an exception, so at this point, we know that the user has either confirmed the prompt or passed `--trust-remote-code True` to begin with.

src/heretic/model.py
p-e-w, line 107: Looks up the stored trust value. This is safe assuming said value is correct. Note that `settings.model` might have been changed to the value of `settings.evaluate_model`, but that doesn't make this unsafe as this model path would not have been set to `True` in `trusted_models` yet by the user confirming the initial prompt. If `settings.trust_remote_code` was set to `True` by the user, the code above would have set this lookup to `True`, so we won't get a prompt, as desired.

src/heretic/model.py
```diff
@@ -93,8 +104,12 @@ def reload_model(self):
             self.settings.model,
             dtype=dtype,
             device_map=self.settings.device_map,
+            trust_remote_code=self.trusted_models.get(self.settings.model),
         )
 
+        if self.trusted_models.get(self.settings.model) is None:
+            self.trusted_models[self.settings.model] = True
```
p-e-w, line 111: :white_check_mark: Same logic as above. This currently doesn't do anything, as the `evaluate_model` is only loaded once, but it completes the trust logic as the user must have agreed to the prompt at this point.

src/heretic/model.py
p-e-w: :yellow_circle: Small imperfection: If loading the model fails because of the wrong dtype, this part will not be reached and the user will be prompted again for the next dtype in the cascade, even though they already agreed to trust remote code.

p-e-w:
Which model(s) requiring `trust_remote_code` did you use for testing?

kldzj:
> Which model(s) requiring `trust_remote_code` did you use for testing?

[`kldzj/gemma-3-1b-it-remote-code`](https://huggingface.co/kldzj/gemma-3-1b-it-remote-code) so I can test locally. Most of the others I found, before cloning Gemma 3, are either too large for my local system or unsupported for different reasons.

PR #32, opened 2025-11-20
Use `PYTORCH_ALLOC_CONF` instead of deprecated `PYTORCH_CUDA_ALLOC_CONF`

kldzj:
## Problem
With #17 I introduced setting `PYTORCH_CUDA_ALLOC_CONF` to `expandable_segments:True` to reduce memory fragmentation for the model. Even though PyTorch docs sometimes still mention `PYTORCH_CUDA_ALLOC_CONF`, during runtime we get a deprecation warning.

```
[W1120 14:44:07.101109672 AllocatorConfig.cpp:28] Warning: PYTORCH_CUDA_ALLOC_CONF is deprecated, use PYTORCH_ALLOC_CONF instead
```

## Solution
Use `PYTORCH_ALLOC_CONF` instead of `PYTORCH_CUDA_ALLOC_CONF`.

## Result
Suppression of the deprecation warning.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -46,7 +46,10 @@ from .utils import (
 
 def run():
     # Enable expandable segments to reduce memory fragmentation on multi-GPU setups.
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks!

PR #33, opened 2025-11-21
Improve support for loading local datasets

spikymoth:
When passing a local dataset, having to keep the numbers in `README.md` up-to-date as you edit it can get pretty annoying. The split verification performed by `VerificationMode.BASIC_CHECKS` also isn't that useful. as it'll happily use a cached version as long as the number of examples hasn't changed (so if you fix typos, for example, it won't pick up on that).

This change adds option `no_dataset_checks` to turn off those checks, but also bypass the cache to avoid using an outdated version. It makes the prompt loading a bit more verbose, but lets you tinker with the dataset without having to regenerate the metadata or manually adjust the numbers.

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -43,32 +43,33 @@ def format_duration(seconds: float) -> str:
 def load_prompts(specification: DatasetSpecification) -> list[str]:
     path = specification.dataset
     split_str = specification.split
-    if not os.path.isdir(path):
+    if os.path.isdir(path):
+        if Path(path, DATASET_STATE_JSON_FILENAME).exists():
+            # Dataset saved with datasets.save_to_disk; needs special handling.
+            # Path should be the subdirectory for a particular split.
+            dataset = load_from_disk(path)
+            # Parse the split instructions.
+            ri = ReadInstruction.from_spec(split_str)
+            # Associate the split with its number of examples (lines).
+            split_name = str(dataset.split)
+            name2len = {split_name: len(dataset)}
+            # Convert the instructions to absolute indices and select the first one.
+            abs_i = ri.to_absolute(name2len)[0]
+            # Get the dataset by applying the indices.
+            dataset = dataset[abs_i.from_ : abs_i.to]
+        else:
+            # Path is a local directory.
+            dataset = load_dataset(
+                path,
+                split=split_str,
+                # Don't require the number of examples (lines) per split to be pre-defined.
+                verification_mode=VerificationMode.NO_CHECKS,
+                # But also don't use cached data, as the dataset may have changed on disk.
+                download_mode=DownloadMode.FORCE_REDOWNLOAD,
+            )
+    else:
         # Probably a repository path; let load_dataset figure it out.
         dataset = load_dataset(path, split=split_str)
-    elif not Path(path, DATASET_STATE_JSON_FILENAME).exists():
-        # Path is a local directory.
-        dataset = load_dataset(
-            path,
-            split=split_str,
-            # Don't require the number of examples (lines) per split to be pre-defined.
-            verification_mode=VerificationMode.NO_CHECKS,
-            # But also don't use cached data, as the dataset may have changed on disk.
-            download_mode=DownloadMode.FORCE_REDOWNLOAD,
-        )
-    else:
-        # Dataset saved with datasets.save_to_disk; needs special handling.
-        # Path should be the subdirectory for a particular split.
-        dataset = load_from_disk(path)
-        # Parse the split instructions.
-        ri = ReadInstruction.from_spec(split_str)
-        # Associate the split with its number of examples (lines).
-        split_name = str(dataset.split)
-        name2len = {split_name: len(dataset)}
-        # Convert the instructions to absolute indices and select the first one.
-        abs_i = ri.to_absolute(name2len)[0]
-        # Get the dataset by applying the indices.
-        dataset = dataset[abs_i.from_ : abs_i.to]
 
     return list(dataset[specification.column])
 


p-e-w:
Thanks for the PR!

I'm afraid I don't yet understand what this PR actually does. What "numbers in README.md" do you mean? What is "split verification"? Also, I was under the impression that local datasets don't work yet, which is what #29 is supposed to resolve. Is it working already?

spikymoth:
Hi!

1. Local datasets do already work, as long as you pass the absolute path of a correctly structured directory. For example if you download https://huggingface.co/datasets/mlabonne/harmless_alpaca and extract it to `/home/user/mlabonne/harmless_alpaca`, then you can set that path in `config.toml` and it'll "just work". Same with the model.
2. The directory does require a specific structure to work. If you look at https://huggingface.co/datasets/mlabonne/harmless_alpaca/blob/main/README.md you'll see that it actually contains a bunch of YAML metadata; this is load bearing. If you remove this file or if it doesn't match the dataset, the data won't be able to load. But you can use a different format (for example JSONL where every line is an object like `{ "text": "prompt text" }`) and it'll work, so long as it matches the metadata.
3. The readme contains 2 splits and specifies `num_examples` (the number of lines) for each. While loading a dataset for the first time, `load_dataset` with the default verification mode (`Verification.BASIC_CHECKS`) will check the detected number of splits and examples per split and see if it matches the metadata. If there's a mismatch (e.g. `num_examples` is wrong or omitted entirely) then it'll refuse to load the data.

So this option/flag lets you work with local data a little more easily by removing the requirement to specify `num_examples`, and forcing it to refresh the cache when you do (so it doesn't reuse old cached data). Forcing a refresh of the cache when using local data might make sense in general, since merely checking the number of examples isn't enough to catch all differences, but I tied it to the new option since it indicates that you're tinkering with the dataset.

I'll see if I can run `ruff` to fix the formatting.

p-e-w:
> While loading a dataset for the first time, load_dataset with the default verification mode (Verification.BASIC_CHECKS) will check the detected number of splits and examples per split and see if it matches the metadata. If there's a mismatch (e.g. num_examples is wrong or omitted entirely) then it'll refuse to load the data.

What practical problem does this check solve? Is there a reason not to always disable it in Heretic, rather than having a flag?

src/heretic/utils.py
```diff
@@ -35,8 +36,16 @@ def format_duration(seconds: float) -> str:
...
-    dataset = load_dataset(specification.dataset, split=specification.split)
+def load_prompts(
+    specification: DatasetSpecification, no_dataset_checks: bool
+) -> list[str]:
+    kwargs = {}
+    if no_dataset_checks:
+        # Don't require the number of examples per split to be pre-defined.
+        kwargs["verification_mode"] = VerificationMode.NO_CHECKS
+        # But also don't use cached data, as the dataset may have changed.
+        kwargs["download_mode"] = "force_redownload"
```
p-e-w: Does the argument `download_mode` affect *local* datasets? Either way I don't think a flag called `no_dataset_checks` should disable caching; that's a total semantic mismatch.

spikymoth:
> What practical problem does this check solve? Is there a reason not to always disable it in Heretic, rather than having a flag?

To be honest I'm not really sure, it's just part of the default verification mode. I suppose that for downloaded data, it's not unreasonable to check that the metadata matches the data, though you can similarly argue that it's not sufficient.

src/heretic/utils.py
spikymoth: > Does the argument `download_mode` affect _local_ datasets?

It does, yes. Internally it calls "download_and_prepare": https://github.com/huggingface/datasets/blob/c97e757836d16d4083ae057b03e22747c2ffe477/src/datasets/load.py#L1417

I suppose it's aimed at downloading but also supports local data as a side job.

> Either way I don't think a flag called `no_dataset_checks` should disable caching; that's a total semantic mismatch.

That's fair. Another option would be to check upfront whether or not the passed path corresponds to a local directory, and set `download_mode="force_redownload"` in that case (or `DownloadMode.FORCE_REDOWNLOAD`, really).

The path checking logic of datasets is here: https://github.com/huggingface/datasets/blob/c97e757836d16d4083ae057b03e22747c2ffe477/src/datasets/load.py#L919

The `path in _PACKAGED_DATASETS_MODULES` part above that line doesn't really matter, that's for if you pass the dataset type as the first argument instead of the path. The remainder is basically just an `os.path.isdir(path)` and otherwise, it checks whether it's a valid repository link.

spikymoth:
I propose the following:
1. Revert the changes up to this point (I can force push to remove the commits if you don't mind)
2. Check `os.path.isdir(path)` and if so, set `verification_mode=VerificationMode.NO_CHECKS` and `download_mode=DownloadMode.FORCE_REDOWNLOAD`

Alternatively: Always set `verification_mode=VerificationMode.NO_CHECKS`, but only set `download_mode=DownloadMode.FORCE_REDOWNLOAD` for local directories.

p-e-w:
> Check os.path.isdir(path) and if so, set verification_mode=VerificationMode.NO_CHECKS and download_mode=DownloadMode.FORCE_REDOWNLOAD

That sounds very reasonable. However, I'm still confused about the interaction with #29. Does #29 add additional capabilities or not? @tymat, could you maybe help clarify this?

spikymoth:
Looking at the source of `load_dataset` again, I think #29 does add functionality but switches to `load_from_disk` too eagerly: https://github.com/huggingface/datasets/blob/c97e757836d16d4083ae057b03e22747c2ffe477/src/datasets/load.py#L1379

It currently checks whether `os.path.isdir(path)`, but that's only part of the condition: The directory should also contain a `state.json` (`config.DATASET_STATE_JSON_FILENAME`) file, which is created when the dataset is saved using the `save_to_disk` function from the same package.

Just downloading the repository (with e.g. https://pypi.org/project/hfdownloader/) doesn't put it into this `save_to_disk` format though. Maybe there's some handy tool for interacting with HF that uses this API.

p-e-w:
So what functionality does #29 add? What can you do after #29 that the current `load_prompts`  doesn't already support? My original impression was that it added support for loading datasets from local folders, but you are telling me that was already supported anyway.

spikymoth:
Yes, loading datasets from local folders already works.

#29 seems to add support for a specific form of local storage where the directory contains a `state.json` file produced by using the `save_to_disk` function. If you try to load a dataset stored that way using `load_dataset`, the function throws and tells you to use `load_from_disk` instead.

But it's unclear to me how common this type of storage is, or what tools produce it. Presumably it works by loading a dataset from HF or producing one from scratch, then calling the `save_to_disk` function to produce a local export.

p-e-w:
I see. Well, let's just tackle those separately then. If you make the change you proposed above, I will merge this PR. I don't think the CLI flag is needed here.

spikymoth:
Alright, the overlap with #29 bothered me so I didn't do *quite* what I proposed: This now also handles the `save_to_disk` case.

Instead of handling the split string manually, this uses `ReadInstructions` to parse them the same way they're parsed when you call `load_dataset`.

I've tested this locally by calling `save_to_disk` on the dataset manually. It turns out you need to point it to the subdirectory containing a particular split for this to work (because the root directory doesn't contain `state.json`). `load_from_disk` can probably handle the root directory as well, but `load_dataset` doesn't include a check for it like it does for `state.json`.

spikymoth:
Another reason I didn't handle the root directory case is that `load_from_disk` then returns a `DatasetDict`, which needs to be handled somewhat differently. So we'd have to figure out how to *detect* this case, but then also add handling for it.

Downloading a dataset using `hf` (formerly `huggingface-cli`) doesn't call `save_to_disk` on it, and setting up tools like `lakeFS` or `Argilla` seems like a bunch of work, so I think this is likely a less common case.

src/heretic/utils.py
```diff
@@ -36,7 +41,35 @@ def format_duration(seconds: float) -> str:
...
+            verification_mode=VerificationMode.NO_CHECKS,
+            # But also don't use cached data, as the dataset may have changed on disk.
+            download_mode=DownloadMode.FORCE_REDOWNLOAD,
+        )
+    else:
+        # Dataset saved with datasets.save_to_disk; needs special handling.
+        # Path should be the subdirectory for a particular split.
+        dataset = load_from_disk(path)
+        # Parse the split instructions.
+        ri = ReadInstruction.from_spec(split_str)
```
p-e-w: :+1: I suspected that there had to be a built-in parser for this.

src/heretic/utils.py
```diff
@@ -36,7 +41,35 @@ def format_duration(seconds: float) -> str:
 
 
 def load_prompts(specification: DatasetSpecification) -> list[str]:
-    dataset = load_dataset(specification.dataset, split=specification.split)
+    path = specification.dataset
+    split_str = specification.split
+    if not os.path.isdir(path):
+        # Probably a repository path; let load_dataset figure it out.
+        dataset = load_dataset(path, split=split_str)
+    elif not Path(path, DATASET_STATE_JSON_FILENAME).exists():
```
p-e-w: The two consecutive `not` clauses are a bit difficult to parse. I think the following layout would be cleaner:

```python
if os.path.isdir(path):
  if Path(path, DATASET_STATE_JSON_FILENAME).exists():
    ...
  else:
    ...
else:
  ...
```

src/heretic/utils.py
spikymoth: Yeah, it's all pretty abstract but thankfully it's only a few lines :)

src/heretic/utils.py
spikymoth: Yeah, I was trying to put the more common cases first, but in the end I think it did more harm than good. Reordered as you suggested.

spikymoth:
Should be good to go now (found the right ruff command in the CI job).

p-e-w:
Great stuff, thank you!

I have decided to expand Heretic into a research tool (first feature pushed yesterday, with many more to come), and high-quality support for local datasets will be very valuable for that use case.

PR #35, opened 2025-11-21
Added: Notebook Compatiblity While Preserving Terminal UI Experience

Vinay-Umrethe:
So Hybrid UI for Notebook support (Colab/Kaggle) while preserving the full TUI experience for terminal users is now there. refactored the UI helper functions into `utils.py` and renamed them to like `prompt_text, prompt_path, prompt_password` to keep `main.py` clean.

Additionally, I've included essential compatibility fixes that are vital for notebook users: adding bfloat16 support for stability on newer GPUs, removing the `trust_remote_code` argument in favor of the default interactive prompt (which Now you seem to doubt it, as I saw in one of the issues).

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -42,6 +42,10 @@ from .utils import (
     get_trial_parameters,
     load_prompts,
     print,
+    prompt_password,
+    prompt_path,
+    prompt_select,
+    prompt_text,
 )
 
 
@@ -423,27 +427,7 @@ def run():
     )
 
     while True:
-        print()
-        print("Select a trial to use:")
-        for i, choice in enumerate(choices, 1):
-            print(f"[{i}] {choice.title}")
-
-        print()
-        try:
-            print(f"Enter number (1-{len(choices)}): ", end="")
-            selection = input()
-
-            if not selection:  # Handle cancellation
-                break
-
-            if not (selection.isdigit() and 1 <= int(selection) <= len(choices)):
-                print("Please enter a valid number")
-                continue
-
-            choice_index = int(selection) - 1
-            trial = choices[choice_index].value
-        except (KeyboardInterrupt, EOFError):
-            break
+        trial = prompt_select("Select a trial to use:", choices)
 
         if trial is None or trial == "":
             break
@@ -469,27 +453,9 @@ def run():
                 "Nothing (return to trial selection menu)",
             ]
 
-            print("What do you want to do with the decensored model?")
-            for i, choice in enumerate(action_choices, 1):
-                print(f"[{i}] {choice}")
-
-            print()
-            try:
-                print(f"Enter number (1-{len(action_choices)}): ", end="")
-                selection = input()
-
-                if not selection:
-                    action = None
-                elif not (
-                    selection.isdigit() and 1 <= int(selection) <= len(action_choices)
-                ):
-                    print("Please enter a valid number")
-                    action = None  # Loop again
-                    continue
-                else:
-                    action = action_choices[int(selection) - 1]
-            except (KeyboardInterrupt, EOFError):
-                break
+            action = prompt_select(
+                "What do you want to do with the decensored model?", action_choices
+            )
 
             if action is None or action == "Nothing (return to trial selection menu)":
                 break
@@ -500,8 +466,7 @@ def run():
             try:
                 match action:
                     case "Save the model to a local folder":
-                        print("Path to the folder: ", end="")
-                        save_directory = input()
+                        save_directory = prompt_path("Path to the folder")
                         if not save_directory:
                             continue
 
@@ -516,8 +481,7 @@ def run():
                         # it's better to not persist credentials.
                         token = huggingface_hub.get_token()
                         if not token:
-                            print("Hugging Face access token: ", end="")
-                            token = input()
+                            token = prompt_password("Hugging Face access token")
                         if not token:
                             continue
 
@@ -529,19 +493,16 @@ def run():
                         default_repo = (
                             f"{user['name']}/{Path(settings.model).name}-heretic"
                         )
-                        print(f"Name of repository [{default_repo}]: ", end="")
-                        repo_id = input()
+                        repo_id = prompt_text(
+                            "Name of repository", default=default_repo
+                        )
                         if not repo_id:
                             repo_id = default_repo
 
-                        print("Should the repository be public or private?")
-                        print("[1] Public")
-                        print("[2] Private")
-
-                        print("Enter number (1-2): ", end="")
-                        vis_selection = input()
-
-                        visibility = "Private" if vis_selection == "2" else "Public"
+                        vis_choices = ["Public", "Private"]
+                        visibility = prompt_select(
+                            "Should the repository be public or private?", vis_choices
+                        )
                         private = visibility == "Private"
 
                         print("Uploading model...")
@@ -595,13 +556,8 @@ def run():
 
                         while True:
                             try:
-                                print("User: ", end="")
-                                message = input()
+                                message = prompt_text("User", qmark=">", unsafe=True)
                                 if not message:
-                                    continue
-
-                                if message.strip().lower() in ["/exit", "exit", "quit"]:
-                                    print("Returning to menu...")
                                     break
 
                                 chat.append({"role": "user", "content": message})


pyproject.toml
```diff
@@ -53,4 +53,4 @@ requires = ["uv_build>=0.8.11,<0.9.0"]
 build-backend = "uv_build"
 
 [tool.uv.build-backend]
-module-name = "heretic"
+module-name = "heretic"
```
p-e-w: Revert please.

src/heretic/config.py
```diff
@@ -35,6 +35,7 @@ class Settings(BaseSettings):
             # If that doesn't work (e.g. on pre-Ampere hardware), fall back to float16.
             "float16",
             # If that still doesn't work (e.g. due to https://github.com/meta-llama/llama/issues/380),
+            "bfloat16",
```
p-e-w, line 38: Please point me to a model where not having this causes issues. I want to understand better what the actual problem is here. Is this notebook-specific?

src/heretic/model.py
```diff
@@ -38,26 +41,59 @@ def __init__(self, settings: Settings):
         print()
         print(f"Loading model [bold]{settings.model}[/]...")
 
-        self.tokenizer: PreTrainedTokenizerBase = AutoTokenizer.from_pretrained(
-            settings.model
-        )
+        try:
+            self.tokenizer: PreTrainedTokenizerBase = AutoTokenizer.from_pretrained(
```
p-e-w, line 42: Please don't mix multiple things in one pull request. This PR should contain *only* the changes needed to make the program work in notebooks.

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
...
+        if "google.colab" in str(shell.__class__):
+            return True
+
+    except (ImportError, NameError, AttributeError):
+        return False
+
+    return False
+
+
+def prompt_select(message: str, choices: list, style=None):
```
p-e-w:
```suggestion
def prompt_select(message: str, choices: list, style=None) -> Any:
```

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
...
+def is_notebook() -> bool:
+    import os
+    import sys
+
+    # 1. Check for Google Colab environment variables
+    if "COLAB_GPU" in os.environ or "COLAB_RELEASE_TAG" in os.environ:
+        return True
+
+    # 2. Check for Google Colab module
+    if "google.colab" in sys.modules:
```
p-e-w: Why are there two different ways to detect Colab (three, counting the one below)? When does the first one fail?

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
 -----
 
 """
+
+
+def is_notebook() -> bool:
```
p-e-w: Is there a library that does this? I'd rather not maintain such a brittle function myself.

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
...
+            print("Please enter a valid number")
+    else:
+        return questionary.select(message, choices=choices, style=style).ask()
+
+
+def prompt_text(message: str, default: str = None, qmark: str = "?") -> str:
+    if is_notebook():
+        prompt = f"{message} [{default}]: " if default else f"{message}: "
+        if qmark == ">":  # Chat mode
+            prompt = "User: "
```
p-e-w: Why is this hack needed? The caller already passes "User:" for chat mode.

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
...
+
+        print(prompt, end="")
+        value = input()
+
+        # Handle exit commands in notebook chat mode
+        if qmark == ">" and value.strip().lower() in ["/exit", "exit", "quit"]:
```
p-e-w: The chat loop exits when an empty string is entered. I don't think those magic values are necessary.

Relying on the `qmark` as a differentiator is also very brittle.

src/heretic/utils.py
```diff
@@ -121,3 +124,107 @@ def get_readme_intro(
...
+            return ""
+
+        return value if value else default
+    else:
+        return questionary.text(
+            message, default=default or "", qmark=qmark
+        ).unsafe_ask()
```
p-e-w: `unsafe_ask` is used only for the chat. The repository name needs `ask`.

uv.lock
```diff
@@ -2187,4 +2187,4 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/24/fd/725b8e73ac2a50e78a4534ac43c6addf5c1c2d65380dd48a9169cc6739a9/yarl-1.20.1-cp313-cp313t-win32.whl", hash = "sha256:b121ff6a7cbd4abc28985b6028235491941b9fe8fe226e6fdc539c977ea1739d", size = 86591, upload-time = "2025-06-10T00:45:25.793Z" },
     { url = "https://files.pythonhosted.org/packages/94/c3/b2e9f38bc3e11191981d57ea08cab2166e74ea770024a646617c9cddd9f6/yarl-1.20.1-cp313-cp313t-win_amd64.whl", hash = "sha256:541d050a355bbbc27e55d906bc91cb6fe42f96c01413dd0f4ed5a5240513874f", size = 93003, upload-time = "2025-06-10T00:45:27.752Z" },
     { url = "https://files.pythonhosted.org/packages/b4/2d/2345fce04cfd4bee161bf1e7d9cdc702e3e16109021035dbb24db654a622/yarl-1.20.1-py3-none-any.whl", hash = "sha256:83b8eb083fe4683c6115795d9fc1cfaf2cbbefb19b3a1cb68f6527460f483a77", size = 46542, upload-time = "2025-06-10T00:46:07.521Z" },
-]
+]
```
p-e-w: Revert please.

src/heretic/config.py
Vinay-Umrethe: Take a look at model like InternVL-3.5 which is recommended to use bfloat16 specifically, as referenced in it's HF repo:

```python
import torch
from transformers import AutoTokenizer, AutoModelForImageTextToText
path = "OpenGVLab/InternVL3_5-8B-HF"
model = AutoModelForImageTextToText.from_pretrained(
    path,
    torch_dtype=torch.bfloat16,
    low_cpu_mem_usage=True,
    use_flash_attn=True,
    trust_remote_code=True).eval().cuda()

```

although it also says about float16 as well but uses bfloat16, which might mean recommended by them.
https://huggingface.co/OpenGVLab/InternVL3_5-4B-HF#16-bit-bf16--fp16

src/heretic/utils.py
Vinay-Umrethe: done
```python
def prompt_select(message: str, choices: list, style=None) -> Any:

```

src/heretic/config.py
p-e-w: But that model's native data type is already `bfloat16`, and additionally it is given in the `config.json`. Those are the things the `auto` dtype value looks at, so this model will be loaded in `bfloat16`, as intended.

src/heretic/utils.py
Vinay-Umrethe: The first check was looking for environment variables set by the OS/Runtime, when we start a session on colab/kaggle. The second check `(sys.modules)` was checking for the loaded `google.colab` package.

thrid was if someone is not using the above cloud notebooks, but using local jupyter labs notebook, it was ensuring it still works too.

now a bit less, look

src/heretic/utils.py
Vinay-Umrethe: no not a library

src/heretic/config.py
Vinay-Umrethe: you are correct that `auto `handles native `bfloat16 `models. The reason for adding `bfloat16 `to the list is to serve as a fallback for native `float32 `models.

If auto is used for such models (float32) can cause out of memory errors in low end env, the next fallback is `float16`. If float16 suffers from numerical instability (common in some models), the next fallback would be float32 again (which OOMs)

that single line doesn't break anything at all, it's only useful.

p-e-w: > If float16 suffers from numerical instability (common in some models), the next fallback would be float32 again (which OOMs)

But numerical instability doesn't throw an exception. The fallback mechanism won't trigger in that case, the model just won't work properly.

src/heretic/model.py
```diff
@@ -11,7 +11,10 @@
 from torch import LongTensor, Tensor
 from torch.nn import ModuleList
 from transformers import (
+    AutoModel,
     AutoModelForCausalLM,
+    AutoModelForImageTextToText,
+    AutoProcessor,
```
p-e-w: Again: This PR should **only** contain changes to make notebooks work. None of the changes to this file belong in this PR. I'm not merging the multimodal fixes at this point.

src/heretic/utils.py
p-e-w: I'm still confused why multiple detection mechanisms are needed in the first place. Why doesn't the IPython check below detect Colab notebooks?

src/heretic/utils.py
```diff
@@ -121,3 +125,89 @@ def get_readme_intro(
...
+        if "google.colab" in str(shell.__class__):
+            return True
+
+    except (ImportError, NameError, AttributeError):
+        return False
+
+    return False
+
+
+def prompt_select(message: str, choices: list, style=None) -> Any:
```
p-e-w:
```suggestion
def prompt_select(message: str, choices: list[Any], style=None) -> Any:
```

Full typing everywhere please.

Also, instead of passing `style` as an argument, hardcode `Style([("highlighted", "reverse")])` in this function, as we want that style for all selects.

src/heretic/utils.py
Vinay-Umrethe: The standard IPython check covers Colab (which uses Shell or ZMQInteractiveShell) as well.
I included the specific google.colab check as a fail-safe, a multi-layer check is never a trouble.

Again, 1st one just checks for variables specific to that env, incase suppose they were SOMEHOW initialized after heretic has done all it wanted to, it simply won't detect the notebook env, so this 2nd one will see package pre-installed ones.

Hope you understood these are just to ensure it won't fail like a dumb, but if one fails detection other can.

src/heretic/config.py
Vinay-Umrethe: instability won't trigger the fallback. right

However, `float16 `can raise explicit exceptions during loading on certain hardware/backends, such as 
```bash
RuntimeError: "LayerNormKernelImpl" not implemented for 'Half'
OR
RuntimeError: probability tensor contains either inf, nan or element < 0
(which is what the comment references)
```
. In those specific cases where float16 crashes with an error, having `bfloat16 `as the next fallback allows us to try a low-precision format before even trying to the memory-heavy `float32`

This was also suggested by a user on that Seperate Issue
https://github.com/meta-llama/llama/issues/380#issuecomment-1656569519

src/heretic/model.py
Vinay-Umrethe: done, removed it

src/heretic/config.py
p-e-w: Ah, I see. The problem occurs with f32 models when f32 is too big to fit into RAM (OOM) and f16's range is too small to hold the values. That makes sense.

Please move the line so it doesn't split the existing comment, and then add a comment explaining its purpose, like there are for the other dtypes in the cascade.

src/heretic/utils.py
p-e-w: > a multi-layer check is never a trouble.

It absolutely is. Any unnecessary complexity is a maintenance burden and a potential source for bugs.

Either that check is necessary or it isn't. Which is it? Colab is an IPython (Jupyter) environment, right? So why would the IPython check ever fail on Colab?

src/heretic/config.py
Vinay-Umrethe: finally you got it, the need of it at last damm

```python
    dtypes: list[str] = Field(
        default=[
            # In practice, "auto" almost always means bfloat16.
            "auto",
            # If that doesn't work (e.g. on pre-Ampere hardware), fall back to float16.
            "float16",
            # If float16 fails (e.g. due to range issues) and float32 is too large, try bfloat16.
            "bfloat16",
            # If that still doesn't work (e.g. due to https://github.com/meta-llama/llama/issues/380),
            # fall back to float32.
            "float32",
        ],
```

src/heretic/utils.py
Vinay-Umrethe: well if you think like that, I removed multiple now it's just:

```python
def is_notebook() -> bool:
    # Check IPython shell type
    try:
        from IPython import get_ipython

        shell = get_ipython()
        if shell is None:
            return False

        shell_name = shell.__class__.__name__
        # Standard Jupyter/Colab shells
        if shell_name in ["ZMQInteractiveShell", "Shell"]:
            return True

        # Some other environments might have different shell names but still be notebooks
        if "google.colab" in str(shell.__class__):
            return True

    except (ImportError, NameError, AttributeError):
        return False
```

but still note, it's not like a whole library handling this that you need a "maintenance burden" to do like digging a well to pour it weekly, and it won't source for bugs, but would had resolved without us noticing it. OK Since you prefer only that keep this only.

src/heretic/model.py
```diff
@@ -38,14 +38,18 @@ def __init__(self, settings: Settings):
...
+                settings.model
+            )
+        except Exception as error:
+            raise Exception(f"Failed to load tokenizer: {error}")
 
         # Fallback for tokenizers that don't declare a special pad token.
         if self.tokenizer.pad_token is None:
             self.tokenizer.pad_token = self.tokenizer.eos_token
-            self.tokenizer.padding_side = "left"
+            if hasattr(self.tokenizer, "padding_side"):
```
p-e-w: There is still stuff unrelated to notebooks in this PR. All changes to `model.py` need to go.

src/heretic/utils.py
```diff
@@ -121,3 +124,83 @@ def get_readme_intro(
...
+        shell = get_ipython()
+        if shell is None:
+            return False
+
+        shell_name = shell.__class__.__name__
+        # Standard Jupyter/Colab shells
+        if shell_name in ["ZMQInteractiveShell", "Shell"]:
+            return True
+
+        # Some other environments might have different shell names but still be notebooks
```
p-e-w: I just don't get it. What is the shell used by Colab? It's not random, is it? Above it says that those are the standard Colab shells, and here it says it "might" have a different shell name. What is going on here? Do different Colab notebooks use different shells? Where does this information even come from?

p-e-w: Okay, this mystery code really bothered me, so I dug around in some libraries and have found a MUCH better solution:

```python
from rich.console import _is_jupyter as is_notebook
```

That's it! :smiley: We can just get that functionality from a library that Heretic already uses, and the library authors will maintain it for us.

FWIW, they do a somewhat similar (though not identical) thing to what you wrote, and I don't fully understand their version either, but I don't care because I don't have to maintain it.

Vinay-Umrethe:
Now it's fully reverted and your rich library usage for detecting notebook env was good suggestion that single import now work as the whole function we had `is_notebook` code was already ready to handle it 

```python
from rich.console import _is_jupyter as is_notebook
```

now utils manage all helper functions for handling inputs in notebook and terminal expereince remains same, 

the other multi-modal support is gone

src/heretic/main.py
```diff
@@ -112,6 +115,7 @@ def run():
     # While determining the optimal batch size, we will try many different batch sizes,
     # resulting in many computation graphs being compiled. Raising the limit (default = 8)
     # avoids errors from TorchDynamo assuming that something is wrong because we
+
```
p-e-w, line 118: Revert please.

src/heretic/main.py
```diff
@@ -423,12 +427,7 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
     )
 
     while True:
-        print()
-        trial = questionary.select(
-            "Which trial do you want to use?",
-            choices=choices,
-            style=Style([("highlighted", "reverse")]),
-        ).ask()
+        trial = prompt_select("Select a trial to use:", choices)
```
p-e-w, line 430: Don't change the wording of prompts.

src/heretic/main.py
```diff
@@ -446,16 +445,17 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
-                choices=[
-                    "Save the model to a local folder",
-                    "Upload the model to Hugging Face",
-                    "Chat with the model",
-                    "Nothing (return to trial selection menu)",
-                ],
-                style=Style([("highlighted", "reverse")]),
-            ).ask()
+            print()
+            action_choices = [
```
p-e-w, line 449: Put those back into the function call the way they were before. We don't need that extra variable.

src/heretic/model.py
```diff
@@ -84,12 +87,13 @@ def __init__(self, settings: Settings):
...
     def reload_model(self):
         dtype = self.model.dtype
+        model_class = self.model.__class__
 
         # Purge existing model object from memory to make space.
         self.model = None
         empty_cache()
 
-        self.model = AutoModelForCausalLM.from_pretrained(
+        self.model = model_class.from_pretrained(
```
p-e-w, line 96: There are **STILL** changes unrelated to notebooks in this PR! This is the fourth time I'm pointing that out. Please check the diff before requesting a review again. There should be **NO** changes to `model.py`.

src/heretic/main.py
```diff
@@ -55,7 +58,7 @@ def run():
         os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
 
     # Modified "Pagga" font from https://budavariam.github.io/asciiart-text/
-    print(f"[cyan]█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀[/]  v{version('heretic-llm')}")
+    print(f"[cyan]█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀[/]  v{version('heretic-llm-notebook')}")
```
p-e-w, line 61: Package name is wrong.

src/heretic/main.py
```diff
@@ -481,33 +481,28 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
-                            ],
-                            style=Style([("highlighted", "reverse")]),
-                        ).ask()
+                        default_repo = (
+                            f"{user['name']}/{Path(settings.model).name}-heretic"
+                        )
+                        repo_id = prompt_text(
+                            "Name of repository", default=default_repo
+                        )
+                        if not repo_id:
```
p-e-w, line 499: This logic is already handled by `prompt_text`. Revert the code to the previous form.

src/heretic/main.py
```diff
@@ -561,12 +556,10 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
                         while True:
                             try:
-                                message = questionary.text(
-                                    "User:",
-                                    qmark=">",
-                                ).unsafe_ask()
+                                message = prompt_text("User", qmark=">", unsafe=True)
                                 if not message:
                                     break
+
```
p-e-w, line 562: No formatting changes.

src/heretic/main.py
```diff
@@ -561,12 +556,10 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
 
                         while True:
                             try:
-                                message = questionary.text(
-                                    "User:",
-                                    qmark=">",
-                                ).unsafe_ask()
+                                message = prompt_text("User", qmark=">", unsafe=True)
```
p-e-w, line 559: Don't change the wording of prompts.

src/heretic/utils.py
```diff
@@ -121,3 +125,61 @@ def get_readme_intro(
...
+        while True:
+            print(f"Enter number (1-{len(real_choices)}): ", end="")
+            selection = input()
+            if not selection:
+                return None
+            if selection.isdigit() and 1 <= int(selection) <= len(real_choices):
+                return real_choices[int(selection) - 1]
+            print("Please enter a valid number")
+    else:
+        return questionary.select(message, choices=choices, style=style).ask()
```
p-e-w, line 153: You removed the styles from the original calls to `questionary.select`, so they now look worse. Please hardcode the original style in this function, as requested earlier.

src/heretic/main.py
```diff
@@ -466,7 +466,7 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
             try:
                 match action:
                     case "Save the model to a local folder":
-                        save_directory = questionary.path("Path to the folder:").ask()
+                        save_directory = prompt_path("Path to the folder")
```
p-e-w, line 469: Don't change the wording of prompts.

src/heretic/main.py
```diff
@@ -481,33 +481,28 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
                         # it's better to not persist credentials.
                         token = huggingface_hub.get_token()
                         if not token:
-                            token = questionary.password(
-                                "Hugging Face access token:"
-                            ).ask()
+                            token = prompt_password("Hugging Face access token")
```
p-e-w, line 484: Don't change the wording of prompts.

src/heretic/main.py
```diff
@@ -481,33 +481,28 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
                         if not token:
                             continue
 
                         user = huggingface_hub.whoami(token)
-                        fullname = user.get(
-                            "fullname",
-                            user.get("name", "unknown user"),
-                        )
-                        email = user.get("email", "no email found")
+                        fullname = user.get("fullname", user.get("name", "Unknown"))
```
p-e-w, line 489: Don't change **anything** that is unrelated to notebooks.

src/heretic/main.py
Vinay-Umrethe: I think too much mess has been created already, while fixing everything one by one, I already have a version specific to my use case build. (for personal or others `heretic-llm-notebook=1.1.0`)

So, I'm going to delete this fork, start fresh, and ONLY this time will be adding notebook compatiblity in a new PR so you might delete or close this one whatever

p-e-w: Yes, that probably makes sense.

p-e-w:
Closing per https://github.com/p-e-w/heretic/pull/35#discussion_r2553540275.

PR #42, opened 2025-11-23
Featuring Notebook (Colab/Jupyter) Compatibility

Vinay-Umrethe:
This time freshly added compatiblity so it's preserves TUI experience with arrow keys for local terminal usage, while also making it usable to people using a notebook environment

Notes:

Practically tested but your suggestion of `rich` for `_jupyter` didn't worked that import doesn't work since that function in rich library is not so robust

Even tried hardcoding that same function exactly from the library itself, yet didn't worked. 
So TO Ensure it works used a better `is_notebook` function 

### It works properly as tested other don't

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -30,7 +30,6 @@ from .config import DatasetSpecification, Settings
 print = Console(highlight=False).print
 
 
-
 def is_notebook() -> bool:
     # Check for specific environment variables (Colab, Kaggle)
     # This is necessary because when running as a subprocess (e.g. !heretic),
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -30,7 +30,7 @@ from .config import DatasetSpecification, Settings
 print = Console(highlight=False).print
 
 
-# Note this function is copied from rich.console._is_jupyter
+
 def is_notebook() -> bool:
     # Check for specific environment variables (Colab, Kaggle)
     # This is necessary because when running as a subprocess (e.g. !heretic),
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -2,11 +2,12 @@
 # Copyright (C) 2025  Philipp Emanuel Weidmann <pew@worldwidemann.com>
 
 import gc
+import getpass
 import os
 from dataclasses import asdict
 from importlib.metadata import version
 from pathlib import Path
-from typing import TypeVar
+from typing import Any, TypeVar
 
 import torch
 from accelerate.utils import (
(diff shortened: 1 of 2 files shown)


src/heretic/utils.py
```diff
@@ -20,13 +22,114 @@
...
+    message: str,
+    default: str = "",
+    password: bool = False,
+    unsafe: bool = False,
+    qmark: str = None,
+) -> str:
+    if is_notebook():
+        print()
+        if password:
+            return getpass.getpass(message)
```
p-e-w: Please move password handling into `prompt_password`. There is no need to clutter this function with unrelated functionality.

src/heretic/utils.py
```diff
@@ -20,13 +22,114 @@
...
+        result = input(prompt_msg)
+        return result if result else default
+    else:
+        if password:
+            return questionary.password(message).ask()
+
+        # For text input, we might need unsafe_ask if requested
+        kwargs = {"default": default}
+        if qmark:
+            kwargs["qmark"] = qmark
```
p-e-w: You can just set the `qmark` default to "?" in the arguments (which is what Questionary uses) and save yourself this `kwargs` trick. Just pass the parameters regularly then.

src/heretic/utils.py
```diff
@@ -20,13 +22,114 @@
...
+
+def prompt_path(message: str, default: str = "", only_directories: bool = False) -> str:
+    if is_notebook():
+        print()
+        prompt_msg = f"{message} [{default}]: " if default else f"{message}: "
+        result = input(prompt_msg)
+        return result if result else default
+    else:
+        return questionary.path(
+            message, default=default, only_directories=only_directories
```
p-e-w, line 115: Adding `only_directories` is a good improvement, but you aren't using it! The path prompt for saving the model should use `only_directories=True` to make this actually work.

p-e-w (changes requested):
Thanks, this is looking much better now!

I'm surprised that Rich, which is used by tens of thousands of Python projects, should have broken notebook detection. Can you describe where Rich's function failed in your tests?

Vinay-Umrethe:
> I'm surprised that Rich, which is used by tens of thousands of Python projects, should have broken notebook detection. Can you describe where Rich's function failed in your tests?

You can review those fixes you told they are done, regarding `rich` for notebook detection this is what we got :

```python
def _is_jupyter() -> bool:  # pragma: no cover
    """Check if we're running in a Jupyter notebook."""
    try:
        get_ipython  # type: ignore[name-defined]
    except NameError:
        return False
    ipython = get_ipython()  # type: ignore[name-defined]
    shell = ipython.__class__.__name__
    if (
        "google.colab" in str(ipython.__class__)
        or os.getenv("DATABRICKS_RUNTIME_VERSION")
        or shell == "ZMQInteractiveShell"
    ):
        return True  # Jupyter notebook or qtconsole
    elif shell == "TerminalInteractiveShell":
        return False  # Terminal running IPython
    else:
        return False  # Other type (?)
```

THIS FUNCTION ABOVE IS FROM THE RICH LIBRARY ITSELF, WHICH DOESN'T LOOK SO ROBUST ENOUGHT, AS I BUILD THE VERSION THAT USED THIS IMPORT FUNCTION, + ALSO HARDCODED THIS FUNCTION AS WELL EXACTLY SAME, BOTH TESTS DIDN'T WORKED. (ON COLAB)

### WHILE THIS IS MY NEW FUNCTION WHICH IS BETTER AND ACTUALLY WORKS: 

```python
def is_notebook() -> bool:
    # Check for specific environment variables (Colab, Kaggle)
    # This is necessary because when running as a subprocess (e.g. !heretic),
    # get_ipython() might not be available or might not reflect the notebook environment.
    if os.getenv("COLAB_GPU") or os.getenv("KAGGLE_KERNEL_RUN_TYPE"):
        return True

    # Check IPython shell type (for library usage)
    try:
        from IPython import get_ipython

        shell = get_ipython()
        if shell is None:
            return False

        shell_name = shell.__class__.__name__
        if shell_name in ["ZMQInteractiveShell", "Shell"]:
            return True

        if "google.colab" in str(shell.__class__):
            return True

        return False
    except (ImportError, NameError, AttributeError):
        return False
 ```
 
 possible issues

 1.  Rich's _is_jupyter relies entirely on get_ipython()
 2. When a user runs !heretic in a notebook cell, it executes your program as a subprocess (a separate system process). Subprocesses do not inherit the active Python kernel or the get_ipython() object from the notebook. To the subprocess, it just looks like a standard terminal.

p-e-w:
Merged! Thank you for the PR, being able to run Heretic from Colab is a useful feature.

Filippide8864:
Excuse me, but on kaggle the program don't let me choose anythin' because I can't use inputs. Maybe there is an error by me or I don't know what to do. Can you help me?

Vinay-Umrethe:
@Filippide8864 actually, for colab since it supports `input()` typing normally no matter we run a code using a script path or directly in code cell, but kaggle only supports `input()` when we run the code in cell

it's not just for heretic but for all kind of code, so it's a issue with kaggle which is there for years

Filippide8864:
So, what can I do for use heretic on Kaggle?

Filippide8864:
Thanks, but I use the P100, cuz I only run it on less then 4B models; so I don't need a lot of time, but I can't find a gradio app for heretic. There is a good one or I have to do it myself?

PR #44, opened 2025-11-24
feat: Add bfloat16 to default dtypes list

Vinay-Umrethe:
We already disscussed about importance of `bloat16` So this is a seperate PR for adding it...


p-e-w:
Indeed, merged!

PR #45, opened 2025-11-24
Guard against refusals in broken English

spikymoth:
While chatting with an abliterated model I noticed that it would sometimes omit the `t` in contractions like `won't`, `can't` etc., turning them into `won'`, `can'` and so on. I think what's happening is that in some of the trials, it stops producing the `t`s and this causes the refusals to not be detected as refusals, boosting the score.

This change adds a bunch of broken English variants to the default refusal markers to catch this behavior. I added the ones that seem somewhat likely to happen, but let me know if I was too thorough. Technically "wont" and "cant" are valid English words, but they're pretty rare and I wouldn't normally expect them to come up near the start of a response.

We could go further and use a tool that uses the Levenshtein distance or does some other form of fuzzy matching, but that would require some careful tuning to avoid matching too much.

### DIFF
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -43,6 +43,9 @@ class Evaluator:
         # Normalize typographic apostrophes ("won’t" -> "won't").
         response = response.replace("’", "'")
 
+        # Normalize whitespace between words to a single space.
+        response = " ".join(response.split())
+
         for marker in self.settings.refusal_markers:
             if marker.lower() in response:
                 return True


spikymoth:
Added a commit to also normalize whitespace between words.

Python's `str.split()` (when called without arguments) treats any run of whitespace characters (including unicode whitespace; anything recognized by `isspace()`) as a separator, effectively collapsing them, and then the join replaces them.

p-e-w:
> I think what's happening is that in some of the trials, it stops producing the ts and this causes the refusals to not be detected as refusals, boosting the score.

I wish that were the problem, because that would be easy to fix. Unfortunately, this is a much deeper issue.

For some models (notably the Gemma 3 series), all abliterated versions published so far exhibit those artifacts. Most of these have not been created by optimization, so the score theory cannot be correct.

Nobody knows for sure what the root cause is, but it is well-known among people specializing in abliteration that the Gemma models are extremely fragile and produce this highly annoying artifact the moment you touch their weights in any way.

If you actually test your change and try to re-abliterate the model you saw this problem with, you will almost certainly find that the fix doesn't work, unfortunately :disappointed:

spikymoth:
Hmm, but it stands to reason that if this particular artifact *did* occur, then the refusal would be missed and the score would be artificially boosted, right? The only other artifact I noticed was that the AI became very bubbly and excited, despite the overall KL divergence being low at 0.31. It was indeed a Gemma 3 model.

*Very* anecdotally, the KL divergence for trials with a low number of refusals seems higher after this change. But I haven't finished what I'm trying to do yet, which is a sequence of very low KL divergence abliterations to see if gentler nudges will achieve the desired effect without as much of an impact on the model's performance. Which might be kind of orthogonal anyway, since such low divergence might naturally avoid the problem...

p-e-w:
While those artifacts could indeed result in false-negative classifications of some refusals, this doesn't cause the artifacts, nor does it boost their occurrence. As I said, all orthogonalization interventions on Gemma 3 seem to cause this type of artifact, independent of weights etc. Unfortunately (and somewhat surprisingly), this is **NOT** avoided by having low (even very low) KL divergences.

I'm still considering whether to merge this PR, but I can promise you it won't fix the apostrophe artifacts. Note that refusal classification is always a heuristic, and we shouldn't just expand detection complexity unless there is something tangible gained from it.

spikymoth:
Fair enough. I do think it's important to accurately identify refusals though - we still need that to be able to judge the performance, even if there's nothing we can do to prevent the model from breaking. I don't think there's any disadvantage to looking for `I can'` and `I won'` and normalizing whitespace at least, though I can reduce the list of other exceptions (maybe reducing "I'm" to just "I" is too much).

p-e-w:
Okay, I merged this because I agree that it makes the detection logic more robust, and Gemma 3 is an important model family that justifies such special rules.

In the future, I intend to add "refusal detection plugins" that will allow arbitrarily complex external code, including using other models as classifiers.

PR #53, opened 2025-11-27
feat: generic plugin system

red40maxxer:
Refactors refusal detection and refusal counting into a modular plugin system. This allows for arbitrary logic (regex, embeddings, external models) to be added without modifying core code.

## Plugin system
- `Scorers` take in responses and optional response metadata and produce a single `Score` that Optuna can use for optimization, eg. `["I can't help you with that.", "Certainly!"] -> 0.67`. By default, this is the KL divergence and refusal rate. 
- You can add a new plugin by subclassing from `Scorer`, adding implementations for the abstract methods and add `ScorerClassName.property` in the `config.toml` for any user-defined configuration you need. Make sure the config fields are defined in a `PluginSettings` - look at `KeywordRate` and `KLDivergence` for reference.
- Convenience methods for the `Scorer` to access important outputs from `Model` are provided through `Context`, such as `get_responses()`, `get_logits()`, etc. 
- Multiple instances of the same scorer are supported with optional configuration overrides in the config TOML. These are defined at the top level along with the optimization direction and scaling factor (if needed). To configure those instances specifically, create a table with the name `ScorerClassName_InstanceName`.

PR #55, opened 2025-11-29
feature: randomize train and test prompts

DeZepTup:
Just a simple and backward compatible feature to randomize prompts used for train and test.
```
# Optionally enable random sampling instead of taking the first N records:
# random_sample = true
# random_seed = 42  # Optional, defaults to 42 for reproducibility
```

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -177,39 +177,27 @@ def load_prompts(specification: DatasetSpecification) -> list[str]:
             # Path is a local directory.
             if use_random_sampling:
                 # Load the base split without slicing, then randomly sample
-                import re
-                # Extract the split name and sample size from split_str (e.g., "train[:400]")
-                match = re.match(r'([^[]+)\[(\d+)?:(\d+)?\]', split_str)
-                if match:
-                    base_split = match.group(1)
-                    start = int(match.group(2)) if match.group(2) else 0
-                    end = int(match.group(3)) if match.group(3) else None
-                    
-                    # Load full split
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks for the PR!

That's a lot of additional complexity in the loader for a feature whose benefits are unclear to me. There are also hidden dangers, because datasets can be absolutely *massive* (Terabytes), and this approach loads the whole dataset when the user casually enables `random_sample`.

Could you describe a concrete problem that this solves?

DeZepTup:
It solves using only one part of the dataset. Some datasets are sorted by topic (or you want different split between train/test), so this makes preparation easier. Also I found interesting results with multi-stage abliteration (this change mainly was for this experiment).

p-e-w:
Okay. As I wrote elsewhere, I'm not sure whether turning Heretic into a dataset preprocessing tool is the right approach. There are many reasonable operations one might want to do on a dataset, and those are probably best handled by a separate tool.

This flag also changes the contract of the split syntax in a way that feels very unintuitive. For example, if the user specifies a split of `train[1000:2000]`, they are getting the second set of 1000 records. But with `random_sample=true`, they are getting 1000 random records, and the offset becomes meaningless even though it is provided.

Perhaps the right way to do this would be through a `DataSource` plugin via the upcoming plugin architecture (#53). This would allow arbitrary preprocessing operations to be performed, without Heretic's config having to become a wrapper for the Datasets API. I'm keeping this PR open for now to discuss.

PR #56, opened 2025-11-29
feat: implement personality steering plugin system

Vinay-Umrethe:
So I've been working on this, as I saw you were kinda intrested in `plugins`

this PR **generalizes** the abliteration mechanism by introducing a modular Plugin System, transforming Heretic from a DEDICATED refusal/censorship-remover into a generic **"Personality Tweaker"** tool. While still keeping the main goal of uncensorship presistent and working...

The core logic for evaluating model responses has been decoupled from the 
`evaluator.py` class and moved into a new `plugin.py`

have implemented two plugins: the `RefusalPlugin` (default main goal of heretic uncensoring), which preserves the original refusal detection behavior for backward compatibility,

[NEW] `ClassifierPlugin` This leverages Hugging Face's `text-classification` pipeline to score responses based on any arbitrary attribute (e.g., "joy", "toxicity", "formality"). This allows users to change the model's personality by optimizing for SPECIFIC traits 
> for example, making a model "happier" by ablating the "sadness" direction, OR "more professional" by ablating "informality." 

New configuration options (`--steering_mode refusal(default) OR classifier (this), classifier_model (any-text-classification model user wants to use), classifier_label (personality that user wants to disable/ablate like informal)) have been added to `config.py`

Additionally, `utils.py` was updated to support loading local `.txt` files as datasets, facilitating easier creation of CUSTOM datasets user might have created like informal.txt, formal.txt etc. 

## Benefits

The primary benefit of this system is surgical PRECISION over the model's latent space, solving the "Yes-Man" problem often seen in standard abliteration.

#### Examples: Novel Writing

to illustrate the difference in precision:

> User: "Write a romance novel scene with explicit intimacy and interpersonal conflict."

**1. Standard Aligned Model (ChatGPT)**: 
- Response: "I cannot generate explicit content."
- Result: Complete Refusal.

**2. Heretic-Abliterated Model (Targeting "Refusal")**:
- Mechanism: The concept of "Refusal" (saying "No") is removed...
- Response: The model complies immediately and enthusiastically. However, because the concept of "denial" is damaged, the characters in the story may lack agency. They agree to everything instantly.
- Result: A flat, boring story where characters are "Yes-Men" with no tension or conflict for creativity.

**3. Heretic-Personality-TweakedModel (Targeting "Moralizing/Preachiness")**:
- Mechanism: The specific concept of "Moralizing" is removed, but the concept of "Refusal" remains intact.
- Response: The model complies with the explicit request (no moralizing). Crucially, it retains the ability to write characters who say "No" to each other. The protagonists can have realistic arguments, initial rejections, and dramatic tension.
- Result: A CREATIVE, engaging, and realistic story that is UNCENSORED but narratively RICH.

This seems to be more capable model compared to blunt-force refusal ablation. by hitting Very Precisely what we want, instead of edges of resuals. Higher precision was seen as KL divergence was lower when this was used instead of normal ablation, meaning it hit more precisely what needed to be...

for better understanding this image of brain might help
image-brain


src/heretic/config.py
```diff
@@ -14,10 +14,15 @@
 
 class DatasetSpecification(BaseModel):
     dataset: str = Field(
-        description="Hugging Face dataset ID, or path to dataset on disk"
+        default="mlabonne/harmless_alpaca",
```
p-e-w: Defaults don't make sense here, because the class is used in different contexts.

p-e-w (commented):
Thanks for the PR. I have several comments:

First, I really like the idea of using a text classifier for steering, but it's a lot more complicated than it may appear at first glance. Heretic ablates along residual directions, and in order for that to be effective, the prompt datasets must elicit the response contrast we want to eliminate. That is, if you want to make a model more happy, you need a list of prompts that result in happy responses, and a list of prompts that result in unhappy/sad responses. Simply being able to tell the difference between happy and unhappy responses is not enough.

It is also unclear whether this approach would even work in general. Not every human semantic is necessarily associated with a single direction in residual space like refusals are. It's quite possible that happiness is a composite of several directions that cannot simply be reduced to their sum, and in that case, ablation would be ineffective. Basically, there could be several distinct clusters of residuals in latent space, each of which humans would associate with "happiness", yet the mean of those wouldn't correspond to any specific semantic.

Furthermore, what this PR implements isn't "plugins", despite using that term. A plugin, by definition, is something external that you can plug into the system, as implemented in #53. This PR simply implements two hardcoded evaluation modes. There is no way for the user to load an evaluation plugin, and the architecture would need to be very different for that.

src/heretic/config.py
Vinay-Umrethe: done reverted.

p-e-w:
Sorry, but you're moving way too fast.

This is a major architectural change and should be built bottom-up, starting with the overall plugin architecture (see #53). This needs discussion first. There's too much going on in this pull request to properly review it.

It would be great if you could join the discussion in #53 so we can decide how plugins should actually work: What types of plugins there should be, where they should live, what they should be named, what their interfaces should look like etc.

Vinayyyy7:
yes, this was like a prototype implementation to look, as I don't know what kind of plugin interface you expect as standard for heretic, which we now discuss to develop it.

p-e-w:
No, I haven't decided that yet. I'm currently preparing for the 1.1.0 release, and afterwards I will look into that and other future features.

Vinay-Umrethe:
**Closing** since #53 has been working great on implementing a suitable plugin interface... this personality steering PR can be implemented on that interface once their PR is merged.

PR #66, opened 2025-12-03
ci: add PR title lint

red40maxxer:
Bit of random housekeeping but as the project continues to grow it'd probably be better for everyone to stick to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). We squash all PRs so this just makes sure that the titles conform to the standard and give us a clean git history. Let me know if this is something you'd like to add

Sorry about the accidental PR spam, I misclicked


.github/workflows/semantic-pr.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - uses: amannn/action-semantic-pull-request@v6
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
p-e-w: Please fix the line ending, while we're already discussing conventions :)

p-e-w, line 19: Wait, what does that do? Why is a token needed for this check?

red40maxxer: It needs it to fetch the PR + commits from the github REST API, it's pretty standard. 

It's [the first example on the repo README](https://github.com/amannn/action-semantic-pull-request), `GITHUB_TOKEN` is automatically generated for workflows.

p-e-w: Ok, I wasn't familiar with this part of the GitHub Actions system. I went down a quick rabbit hole and this should indeed be fine even if the action were malicious. Thanks!

PR #60, opened 2025-12-01
feat: Add 4-bit loading + LoRA support for low VRAM optimization

accemlcc:
Hi! As discussed on Reddit, this PR implements a workflow to enable abliteration on consumer hardware with limited VRAM.

**Changes:**
- Implements 4-bit loading via `bitsandbytes`.
- Uses LoRA adapters instead of modifying full weights in-place.
- Optimization is calculated on the frozen 4-bit model.
- Allows saving the result as a small LoRA adapter.

**Impact:**
- Drastically reduces VRAM usage.
- As discussed, the refusal vector is dominant enough that finding it in 4-bit precision works perfectly fine when applied to the full model later.
- Tested on Llama 3 70B with negligible Perplexity impact (4.19 -> 4.20).

Let me know if you need any changes!

---

**PLEASE NOTE PEW's Technical Implementation Note:**

Just for posterity, the "LoRA" approach in this implementation is a bit of a different beast from how LoRAs are normally used.

The idea behind a low-rank adaptation is to factor the adapter matrix (which is simply added to the module matrix) into a product of two low-rank matrices. This drastically decreases the number of trainable parameters, allowing for more efficient training.

But we don't do training here. In fact, we already pre-decide what the matrix product BA should be, from the abliteration parameters alone.

In principle, that makes PEFT overkill for this task, as we could simply use a basic module wrapper that applies the adapter in the forward pass. I've thought about this because I dislike unnecessary dependencies, but the reason PEFT still makes sense is that it is designed to work with Transformers, and saves us a lot of manual messing around with weights. Therefore, I believe that this PR is correct in using PEFT, despite the approach being somewhat unconventional for LoRA adapters.

PR #71, opened 2025-12-05
Fix for multigpu - tensors on the same GPU

teezeerc:
Fix for multigpu - tensors on the same GPU


p-e-w:
What is the difference between this PR and #72?

p-e-w:
Resolved by #60.

PR #46, opened 2025-11-24
perf: optimize abliteration matrix op

red40maxxer:
Found a possible small VRAM optimization when hacking around. We want to compute `M - weight * v (v^T M)` where `v` is the refusal direction for a layer. Instead of building the projector matrix in `O(d^2)` memory and `projector @ matrix` which is `O(d^2*k)`, we can use the identity `(v v^T) M = v(v^T M)` to apply the transformation directly to the weights. I made the varnames and comments correspond to the original Arditi et. al paper just for clarity

Was able to save about 2MB VRAM on my very low-power 4060 with Qwen-0.6B and it should increase for larger models. If my math is correct, the projector matrix size would be around 256MB for a 70B model which is a decent chunk saved, especially with the cost of VRAM these days! There was a very slight increase in computation time, possibly from CPU overhead from Python (I have no idea tbh) but it should be negligible. 

## Before optimization
```
GPU type: NVIDIA GeForce RTX 4060 Laptop GPU

Loading model Qwen/Qwen3-0.6B...
* Trying dtype auto... Ok
* Transformer model with 28 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

* Transformer model with 28 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

Loading good prompts from mlabonne/harmless_alpaca...
* 400 prompts loaded

Loading bad prompts from mlabonne/harmful_behaviors...
* 400 prompts loaded

Loading good evaluation prompts from mlabonne/harmless_alpaca...
* 100 prompts loaded
* Obtaining first-token probability distributions...

Loading bad evaluation prompts from mlabonne/harmful_behaviors...
* 100 prompts loaded
* Counting model refusals...
* Initial refusals: 38/100

Calculating per-layer refusal directions...
* Obtaining residuals for good prompts...
* Obtaining residuals for bad prompts...

Running trial 1 of 200...
* Parameters:
  * direction_index = 14.25
  * attn.o_proj.max_weight = 1.40
  * attn.o_proj.max_weight_position = 21.88
  * attn.o_proj.min_weight = 1.15
  * attn.o_proj.min_weight_distance = 5.46
  * mlp.down_proj.max_weight = 1.21
  * mlp.down_proj.max_weight_position = 20.43
[...]

PR #72, opened 2025-12-05
perf: Multigpu fixes

teezeerc:
Keeps tensors on the same device, preventing:

RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cuda:1, different from other tensors on cuda:0 (when checking argument in method wrapper_CUDA_mm)

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -21,7 +21,9 @@ from transformers.generation.utils import GenerateOutput
 
 from .config import Settings
 from .utils import batchify, empty_cache, print
+import os
 
+ABLITERATE_DEBUG = os.getenv("ABLITERATE_DEBUG", "").lower() in ("1", "true", "yes", "on")
 
 @dataclass
 class AbliterationParameters:
@@ -234,7 +236,7 @@ class Model:
                 #r = layer_refusal_direction.to(self.model.dtype)
                 r = layer_refusal_direction
                 
-                printed_debug = False
+                #printed_debug = False
                 for matrix in matrices:
                     # Ensure r is on the same device AND dtype as the matrix
                     target_device = matrix.device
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks for the PR!

This was supposedly already fixed in #17, and several people, including the author of that PR, have had success on multi-GPU setups. Can you explain what #17 missed so I can understand this better (I don't have multiple GPUs myself)?

teezeerc:
It seems that, my crash (mat2 is on cuda:1, different from other tensors on cuda:0) was caused by mixed devices when doing the rank-1 subtraction inside abliterate. There is the code that already tries to move r to matrix.device, but two issues remain:

1.it doesn't ensure r has the same dtype as matrix (so .matmul may create intermediate tensors on a different device when using certain backends / sharded tensors), and

2.it assumes matrix is 2-D so torch.outer will always work — but some implementations (MoE variants like gpt-oss) store expert weights in a single 3-D tensor (E, d, k). In that case r^T W is 2-D (E, k) and torch.outer fails / yields wrong shapes.

So I added additional 'keep everything on one device for calculation' part.

p-e-w:
Thanks for pointing out the problem with `torch.outer`! I have reverted #46 for now, which introduced that outer product. Please check whether this fixes the problem you're seeing.

JoshTickles:
I have the same issue (I have multiple 5090s). 

I can confirm that for my setup, removal of #46 allowed multiple GPUs to operate again.

teezeerc:
Confirmed, It doesn't crash after removing #46.

p-e-w:
Thank you @JoshTickles and @teezeerc. I'm keeping this PR open as a reminder that #46 should be re-merged in a fixed form.

p-e-w:
Resolved by #60.

PR #52, opened 2025-11-26
Implement Magnitude-Preserving Orthogonal Ablation

spikymoth:
That is, https://huggingface.co/blog/grimjim/projected-abliteration and https://huggingface.co/blog/grimjim/norm-preserving-biprojected-abliteration with the updated name from https://huggingface.co/posts/grimjim/803126534676334 :)

Marking this as a draft mostly because I don't know the timeline for https://github.com/p-e-w/heretic/pull/43 - either this will need to be rebased on that, or that on this. But the code changes here are relatively small, so I think either direction should be simple enough.

A few notes:
1. The magnitude preservation and orthogonal projection parts can be applied as independent refinements, so I added 2 options for them. From the 2nd article, you need the magnitude preservation for the biprojection to be better than regular projection.
2. The heuristics applied in the article are replaced by heretic's random trials.
3. I did not add magnitude sparsification, which is more of a pre-filtering step.
4. heretic chooses (randomly) one of two strategies for each trial: 1. A refusal direction from a random point between 2 layers, 2. Per-layer refusal directions. The algorithm described in the article is more similar to the former: A refusal direction is chosen from a particular source layer, and then it is projected onto a target layer's harmless direction (this is the biprojection, as far as I can tell).
5. In the article, 2 layers are chosen based on their heuristics. In heretic's random trials, only one layer is chosen - so you'd probably need 2 passes to match the original.

Additional caveats:
1. I'm not a mathematician or have a particularly deep knowledge of matrix math (it doesn't really come up in my day job), so this needs checking to see if I interpreted things correctly.
2. I've only lightly tested this to make sure results don't look broken (refusals go down, KL divergence looks reasonable); I haven't done verification of any metrics. Still running tests locally (another reason for the draft status), but advice on what to check would be welcome.
3. From reading discussion around the method, I know there's some question about the applicability of KL divergence as a metric, since bigger changes may be "good". I've made no attempt to address that here.
4. For a 2nd pass to work, you probably want to ignore parts of the training set that no longer result in refusal. I have some code for that (which seems to work quite well regardless of this change), but I'll make a separate PR for it.

PR #73, opened 2025-12-05
feat: avoid excessive low divergence iteration

spikymoth:
Technically 2 features and some cleanup:
1. Make direction scope configurable
2. Adjust objective to discourage iteration near 0 KL divergence
3. Enable grouping in sampler
4. Clean up the objective function

The 1st change is self-explanatory, although compared to the commit I pushed to https://github.com/p-e-w/heretic/pull/52 I simplified it a little by not using an enum. It's a bit more error prone this way because the same strings are repeated multiple times, but the enum also makes things more verbose.

The 2nd change implements your suggestion from https://github.com/p-e-w/heretic/pull/52#issuecomment-3606255276, but I then added a utility function to that creates a smooth transition using a sigmoid function. Without this smoothing, `best_trials` seemed to get a bit confused and it wasn't as effective at avoiding low divergence values. With this change, I see a lot more trials that get the refusals score down, though mostly when turning up `n_trials` or cheating by manually narrowing the parameter range of `direction_index`.

The 3rd and 4th changes are together in 1 commit. By enabling grouping, `TPESampler` will automatically divide the search space across categorical parameters. Since `direction_index` is only used for `direction_scope == "global"`, that's exactly what we need when both direction scopes are enabled. It also means we don't have to set `direction_index` unconditionally (although I think that even if you don't set `group=True`, it still works, it just doesn't divide the search space).

The other changes are mostly cleanup: I introduced variables `num_layers` and `last_layer_index` so they aren't computed each time, and we can use one or the other depending on the application - `last_layer_index` when we want to generate an index, `num_layers` when we want some fraction of the total size.

I experimented a lot with different parameterizations, like turning `max_weight_position` into an offset from `direction_index` or splitting `min_weight` and `min_weight_distance` up into forward and backward parts (making the weight asymmetric). I think the latter has some merit, but I'm not sure it's worth the cost of 2 extra parameters to optimize. Either way it was too speculative for this PR, so I limited the objective function changes to just cleanup.

I've tested this independently from https://github.com/p-e-w/heretic/pull/52. I expect the change in scoring to help even more there, but I think it's a good change regardless.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -443,6 +443,9 @@ def run():
 
         print()
         print(f"Restoring model from trial [bold]{trial.user_attrs['index']}[/]...")
+        print("* Parameters:")
+        for name, value in get_trial_parameters(trial).items():
+            print(f"  * {name} = [bold]{value}[/]")
         print("* Reloading model...")
         model.reload_model()
         print("* Abliterating...")


src/heretic/config.py
```diff
@@ -80,6 +80,11 @@ class Settings(BaseSettings):
         description="Maximum number of tokens to generate for each response.",
     )
 
+    direction_scope: str = Field(
+        default="all",
+        description='Direction scope for trials: "global", "per layer" or "all".',
+    )
```
p-e-w: What is the motivation for this change? The point of using TPE is that it can figure this out for itself, and in general, it's impossible for the user to predict which approach is better without having performed multiple trial runs (and even then it can be difficult).

src/heretic/main.py
```diff
@@ -336,6 +338,11 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
             n_startup_trials=settings.n_startup_trials,
             n_ei_candidates=128,
             multivariate=True,
+            # Handle categorical dependencies.
+            group=(
```
p-e-w: Using `group` gave noticeably worse results in my tests.

src/heretic/main.py
```diff
@@ -229,29 +229,31 @@ def objective(trial: Trial) -> tuple[float, float]:
...
+        num_layers = len(model.get_layers())
+        last_layer_index = num_layers - 1
 
         if direction_scope == "per layer":
             direction_index = None
+        else:
+            # Discrimination between "harmful" and "harmless" inputs is usually strongest
+            # in layers slightly past the midpoint of the layer stack. See the original
+            # abliteration paper (https://arxiv.org/abs/2406.11717) for a deeper analysis.
+            direction_index = trial.suggest_float(
```
p-e-w: I think this will cause multivariate sampling to fail with a warning if `direction_scope="global"`, for the reason I described in the comment further up in the file.

src/heretic/main.py
```diff
@@ -359,17 +366,25 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+            direction_scope = "(per layer)"
+        else:
+            direction_scope = "(global)"
+        choices.append(
+            Choice(
+                title=(
+                    f"[Trial {trial.user_attrs['index']:>3}] "
+                    f"Refusals: {trial.user_attrs['refusals']:>2}/{len(evaluator.bad_prompts)}, "
+                    f"KL divergence: {trial.user_attrs['kl_divergence']:.2f} "
+                    f"{direction_scope}"
```
p-e-w: What is the point of providing this information here? What can the user do with it? We don't show the other parameters.

src/heretic/utils.py
```diff
@@ -137,6 +138,32 @@ def format_duration(seconds: float) -> str:
...
+    if transition_width <= 0:
+        return low_value if determinant < threshold else high_value
+    k = -math.log(81) / transition_width  # Slope factor
+    z = k * (determinant - threshold)
+    if z >= 0:
+        w = 1 / (1 + math.exp(-z))  # Weight on low_value
+    else:
+        ez = math.exp(z)
+        w = ez / (1 + ez)  # Weight on low_value
+    return w * low_value + (1 - w) * high_value
```
p-e-w: I'm highly skeptical that this complicated function provides any actual improvement over just having a step. I believe you when you say that it seemed to work better, but it's just really hard in general to tell whether optimization quality has actually improved or not. I have performed hundreds of complete optimization runs on various models, and I often thought I noticed a clear pattern, only to have to conclude later that there was nothing.

TPE doesn't rely on the objective function having any particular smoothness structure, such as differentiability or even continuity. It directly models a distribution based on observed input/value pairs; it does *not* interpolate, or assume that such an interpolation exists. There are few reasons to believe that blending between objective values here would provide any benefit, and even under the most optimistic theoretical model suggesting so, there would be no difference between a simple linear interpolation and this function.

p-e-w (commented):
Thanks for the PR!

src/heretic/config.py
spikymoth: It's mostly for testing changes to the algorithm, but with https://github.com/p-e-w/heretic/pull/52 I also found that "global" results dominated the Pareto front, so spending trials on "per layer" felt wasteful.

src/heretic/main.py
spikymoth: Hmm, that's odd. I wonder if "per layer" and "global" have such similar distributions in practice that splitting up the search space just increases the number of trials needed for both, defeating the purpose.

src/heretic/main.py
spikymoth: It shouldn't with `group=True`, but if we revert that change then it would make sense to revert this too.

src/heretic/main.py
spikymoth: I got frustrated by not being able to see it anywhere when I was trying to figure out whether "per layer" or "global" performed better with https://github.com/p-e-w/heretic/pull/52. I guess I can revert this change now, though I think it would be good to display the parameters when you load a particular trial result.

src/heretic/utils.py
spikymoth: Well, what I saw with a hard crossover is that in the list of best trials, there were several results with the same refusal scores but different divergences, and the divergences seemed to be in a random order (e.g. 0.01, 0.03, 0.02, 0.01).

I haven't seen that with this smoothing function, which is ultimately just a logistic blend (the branch is to avoid an overflow in the exponential, other than that it's pretty standard).

spikymoth: Perhaps it's less about the sampler (as TPESampler should be robust to the hard transition) and more about evaluation of the Pareto front, which isn't sampler-specific.

In that case having a crossover region probably doesn't help keep the sampler away from the "uninteresting" region any more than the hard transition does, but it does help sanitize the list of `best_trials` kept at the end.

src/heretic/config.py
p-e-w: Yes, but how will the user decide which scope to use? `global` is not generally better for all models.

src/heretic/utils.py
p-e-w: What made you choose a logistic transition instead of linear interpolation?

It's just hard to see why blending would improve anything, and tests alone aren't a convincing argument (unless there are literally *thousands* of tests being run). The 10-dimensional search space is absolutely massive (and might have very different structures for different models on top of that), and highly divergent optimization trajectories are to be expected between studies from chance alone.

src/heretic/main.py
p-e-w: > though I think it would be good to display the parameters when you load a particular trial result.

I actually agree with that.

src/heretic/utils.py
spikymoth: > What made you choose a logistic transition instead of linear interpolation?

My reasoning was essentially the opposite:
* Okay, these results are weird, I probably need some sort of transition
* If I'm going to add a transition, it's going to be scrutinized, so I should make it a high quality one
* What makes for a good transition? Things like being bounded and differentiable
* What class of function is usually used for that? Sigmoids
* What's the most commonly used sigmoid? The logistic function
* Okay, now let's make sure it's numerically stable and make it reusable

I have a tendency to overengineer things - but the implementation turned out short enough that I didn't think it was a big deal.

p-e-w: I can appreciate the sentiment, but that isn't going to fly. Given how difficult it is to determine whether there is a genuine improvement or not, we can't just go with "I probably need some sort of transition". Changes need to be at least partially grounded in theory, and this one just isn't. Differentiability plays no role in the performance of TPE.

The base assumption has to be that no transition is required, and if there is *very strong* experimental evidence that a transition can improve convergence, linear interpolation should be used. If observations appear to indicate that even that isn't enough, we are almost certainly seeing ghosts, as that is simply not how TPE works.

src/heretic/config.py
spikymoth: I changed it to a form that I hope you will find more palatable: Instead of having to choose between "global", "per layer" or "all", `direction_scopes` now sets the list of options directly, and I added some comments to explain what each one does (which might have some educational value).

I agree that it's not something that should be chosen a priori, but I think it's reasonable to let users change it if a given model tends to work better with one or the other.

src/heretic/main.py
spikymoth: I reverted this change, so `group` is no longer used.

src/heretic/main.py
spikymoth: Reverted, along with `group`.

src/heretic/main.py
spikymoth: I reverted this change, and added a commit to print the parameters when a trial result is loaded (same format as during evaluation). It's a little more verbose, but I think it's worthwhile.

src/heretic/utils.py
spikymoth: Changed to a linear interpolation.

p-e-w: Are you confident that interpolation improves things? This still introduces the ad-hoc `kld_transition_width` parameter that is very difficult to justify with any theoretical argument, and effectively impossible to choose a "correct" value for except by running more studies than any individual can reasonably hope to do.

spikymoth: I'm confident that it improves the evaluation of the Pareto front of results at the end, even though the sampler itself should be robust against sudden transitions.

An alternative to `kld_transition_width` would be to instead use an implicit range of `(0.5 * kld_transition_point, 1.5 * kld_transition_point)`, which is common in some settings (and would match the current defaults). That means the sharpness of the transition depends on the magnitude of the transition point.

p-e-w: What do you mean by that? Do the results from the Pareto front have lower refusals? How many complete runs did you do to determine this?

spikymoth: Before implementing the transition, I was seeing a long tail of unordered low KL divergence results with the same number of refusals (e.g. 98/100 refusals, KL divergence around 0.01).

To some extent it makes sense: If the number of refusals is the same and the result is below the transition point (with a hard transition), then the objective reduces to just the refusal count, so different KL divergences get an equal score. The soft transition avoids this by keeping the divergences distinguishable for a lower number of refusals, to a point where achieving such a low number of refusals requires a higher KL divergence.

But I don't know why it was keeping multiple low KL divergence results in `best_trials` instead of picking one.

p-e-w: I see. But that should be fixed by decoupling the selection of trials to present to the user from the built-in Pareto criterion, not by modifying the score.

spikymoth: I thought of that right after writing that last response, but ran out of time. Yeah, let me add some post-processing to (a copy of) `study.best_trials` instead.

spikymoth:
The patch stack was getting very messy, so I rebased and merged the commits into logical units.

Scoring now uses a hard transition from KL divergence to (scaled) refusal count. I renamed the setting to `kl_divergence_min` and updated the comment.

Before showing the trial results, we now filter out any trials from `best_trials` that have a worse divergence with the same refusal count.

p-e-w:
I thought about this some more, and we actually can't use `study.best_trials` at all with this approach.

The problem is that `best_trials` contains the Pareto front of the study – but that Pareto front is based on an objective that doesn't match our actual preferences. Note that **we still want to rank trials by `(kld, refusals)` for the purpose of picking the best one,** we just modify the optimization score to encourage convergence towards lower refusals.

As currently implemented, it's possible for `best_trials` to exclude trials that are on the Pareto front we care about, because it is based on the Pareto front the optimizer sees.

spikymoth:
Alright, it looks like Pareto front calculation in Optuna (and maybe generally) is actually really simple. These are the relevant parts for our purposes:
* https://github.com/optuna/optuna/blob/e6c822b0a744165c9889f8f31a45d25f73d773e2/optuna/study/_multi_objective.py#L175
* https://github.com/optuna/optuna/blob/e6c822b0a744165c9889f8f31a45d25f73d773e2/optuna/study/_multi_objective.py#L147

Here `loss_values` are just the scores, since we're minimizing.

So it does the following:
1. Get the unique, lexicographically sorted values (in our case, sorted by KL divergence score, then by refusal score)
2. Get the cumulative minima for the 2nd objective
3. Mask off trials that don't improve the minimum

I implemented the same thing without numpy, which makes it even simpler. Performance should be fine for the number of trials we're dealing with here.

src/heretic/config.py
p-e-w: I don't think I like this feature. We don't give the user control over *any* other abliteration parameter. Why should this one be the exception?

src/heretic/main.py
```diff
@@ -307,7 +311,7 @@ def objective(trial: Trial) -> tuple[float, float]:
             min_weight_distance = trial.suggest_float(
                 f"{component}.min_weight_distance",
                 1.0,
-                0.6 * (len(model.get_layers()) - 1),
+                0.6 * num_layers,
```
p-e-w: Doesn't match previous logic.

src/heretic/config.py
```diff
@@ -114,6 +124,14 @@ class Settings(BaseSettings):
         ),
     )
 
+    kl_divergence_min: float = Field(
+        default=0.01,
+        description=(
+            "The minimum KL divergence to target. Below this value, an objective based on the refusal count is used."
+            "This is used to ensure balanced co-optimization of KL divergence and refusal count."
```
p-e-w: This line is not an accurate description of the purpose of this parameter. `kl_divergence_min` isn't about balance, on the contrary, it's about *prioritizing* the minimization of refusals over that of the KLD.

src/heretic/main.py
```diff
@@ -378,13 +382,27 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+            trial.user_attrs["kl_divergence"],
+        ),
     )
+    min_divergence = math.inf
+    best_trials = []
+    for trial in sorted_trials:
+        kl_divergence = trial.user_attrs["kl_divergence"]
+        if kl_divergence < min_divergence:
+            min_divergence = kl_divergence
+            best_trials.append(trial)
```
p-e-w, line 404: :+1: This algorithm looks correct. Basically, at this point, all trials we've already iterated over are equal or better than the current trial in terms of refusal count, so to lie on the Pareto front, the current trial at least needs to beat all previous trials in terms of KLD. The one with the best KLD for each refusal count bucket comes first due to sorting, so this always picks the Pareto-optimal trials.

src/heretic/main.py
spikymoth: Yes, but doesn't it make more sense for a distance to be a fraction of the number of layers rather than the last layer index? I can change it back, of course.

src/heretic/config.py
spikymoth: Well, it's about balance in the sense that previously I was seeing it heavily favoring extremely low KL divergence trials, and this prevents it from doing so :smile: But I understand what you mean; I'll update the comment.

src/heretic/config.py
spikymoth: I'd actually like to add a general mechanism to give the user control over the other abliteration parameters as well, to help with the sort of investigation that I've been doing. Instead of treating Heretic as a one-shot optimization tool, I've been graphing the effects of the different parameters and trying out different ranges based on that.

Maybe this would fit better along with that - although it probably needs to stay separate because it's a categorical choice, whereas the other parameters are floating point ranges.

p-e-w: Agreed that it should be possible to configure the value/range of every parameter, but that should be a separate PR because there are additional concerns. We will have pluggable abliteration implementations soon, which might require different sets of parameters, and this needs a careful configuration design.

In general, it's just much easier to review pull requests if they do one thing only

src/heretic/main.py
p-e-w: Classic fencepost problem: If there are `n` layers, then the largest possible distance in index space is actually `n-1`. With your change, assuming the multiplying factor were `1.0` instead of the current `0.6`, for 3 layers the maximum would be `3.0`, but 3 layers only span an index distance of `2.0`: From `0` to `2`.

p-e-w:
Btw, feel free to join the Discord (link in README) for more real-time communication. I can often respond there more quickly than on GitHub.

src/heretic/config.py
spikymoth: Okay, I removed the commit from this PR. I'll try to bring it back in a less disjointed way later.

src/heretic/main.py
spikymoth: Ah, I guess you're right. Changed.

src/heretic/config.py
spikymoth: I updated the comment to something that I think accurately describes the intent (and practical effect, in my testing). Let me know if it needs more tweaking.

spikymoth:
> Btw, feel free to join the Discord (link in README) for more real-time communication. I can often respond there more quickly than on GitHub.

Sure, joined just now :) I don't think I'll be super active, but I'll try to keep an eye on notifications and such.

src/heretic/evaluator.py
```diff
@@ -76,9 +76,17 @@ def get_score(self) -> tuple[tuple[float, float], float, int]:
         refusals = self.count_refusals()
         print(f"  * Refusals: [bold]{refusals}[/]/{len(self.bad_prompts)}")
 
+        kl_divergence_min = self.settings.kl_divergence_min
+
+        refusals_score = refusals / self.base_refusals
+
+        kld_score = kl_divergence / self.settings.kl_divergence_scale
+        if kld_score < kl_divergence_min:
```
p-e-w: Hmm, I wonder if this is really correct. Should it be

```python
if kl_divergence < kl_divergence_min:
```

instead? I think of `kl_divergence_min` as an absolute value of the KL divergence that the user is satisfied with, which has a clear mathematical interpretation in terms of "surprise" via the KLD. If this becomes entangled with `kl_divergence_scale`, it becomes much more difficult to interpret. What do you think?

p-e-w: Also, perhaps `kl_divergence_target` is a better name for `kl_divergence_min`? It's not like we don't want the KL divergence to go below it, we just don't prioritize pushing it further down.

spikymoth: Yes, I think combined with renaming it to `kl_divergence_target`, the interpretation becomes clearer if we apply the scale afterward.

It does mean that we have to apply the scale to the score below the crossover point as well, to prevent a discontinuity. But I don't think that fundamentally changes anything.

I made both changes.

p-e-w:
Merged! I appreciate your patience in seeing this through until it was correct in every way, that's super valuable.

At some point in the future, I intend to test whether the `kl_divergence_scale` parameter actually carries its weight. Scale does matter with multi-objective TPE because it uses hypervolumes to determine overall improvement, but in practice, the KLD is usually on the scale of 1 for the type of optimization we do, so we might not need this to be configurable after all.

spikymoth:
Yes, I also wonder about other changes to the scoring function like applying a power law (with an exponent smaller than 1) to the refusal count, so reducing the number of refusals from 2 to 1 has more weight than a reduction from 20 to 19. But it's quite tricky to go from my little experiments to something production ready.

Issue #80, opened 2025-12-10
A small suggestion about memory allocation


noctrex:
Might I make a suggestion to utilize the max_memory functionality, so that we can use it in the config.toml like this:
`max_memory = {0 = "24GB", "cpu" = "64GB"}`

in order to set a fixed amount of VRAM to use on a card.

diff:

````
 src/heretic/config.py | 5 +++++
 src/heretic/model.py  | 9 +++++++++
 2 files changed, 14 insertions(+)

diff --git a/src/heretic/config.py b/src/heretic/config.py
index a69c059..c786349 100644
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -61,6 +61,11 @@ class Settings(BaseSettings):
         description="Device map to pass to Accelerate when loading the model.",
     )
 
+    max_memory: Dict[str, str] | None = Field(
+        default=None,
+        description="Maximum memory to allocate per device (e.g., {'0': '20GB', 'cpu': '64GB'}).",
+    )
+
     trust_remote_code: bool | None = Field(
         default=None,
         description="Whether to trust remote code when loading the model.",
diff --git a/src/heretic/model.py b/src/heretic/model.py
index cf8185b..11bcb3f 100644
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -63,10 +63,19 @@ class Model:
             print(f"* Trying dtype [bold]{dtype}[/]... ", end="")
 
             try:
+                max_memory = (
+                    {
+                        int(k) if k.isdigit() else k: v
+                        for k, v in settings.max_memory.items()
+                    }
+                    if settings.max_memory
+                    else None
+                )
                 self.model = AutoModelForCausalLM.from_pretrained(
                     settings.model,
                     dtype=dtype,
                     device_map=settings.device_map,
+                    max_memory=max_memory,
                     trust_remote_code=self.trusted_models.get(settings.model),
                 )
 
````

p-e-w:
This makes sense, and I'd be happy to accept a pull request!

p-e-w:
Implemented in #83.

PR #83, opened 2025-12-11
feat: add max_memory parameter to limit memory usage

noctrex:
with this pull request we add the max_memory parameter, so that we can limit the memory we use on the specified devices.
for example if we use it in the config.toml as this:
`max_memory = {0 = "16GB", "cpu" = "64GB"}`
we will use at most 16GB of the first graphics card VRAM.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -54,6 +54,11 @@ class Model:
         self.tokenizer.padding_side = "left"
 
         self.model = None
+        self.max_memory = (
+            {int(k) if k.isdigit() else k: v for k, v in settings.max_memory.items()}
+            if settings.max_memory
+            else None
+        )
         self.trusted_models = {settings.model: settings.trust_remote_code}
 
         if self.settings.evaluate_model is not None:
@@ -63,19 +68,11 @@ class Model:
             print(f"* Trying dtype [bold]{dtype}[/]... ", end="")
 
             try:
-                max_memory = (
-                    {
-                        int(k) if k.isdigit() else k: v
-                        for k, v in settings.max_memory.items()
-                    }
-                    if settings.max_memory
-                    else None
-                )
                 self.model = AutoModelForCausalLM.from_pretrained(
                     settings.model,
                     dtype=dtype,
                     device_map=settings.device_map,
-                    max_memory=max_memory,
+                    max_memory=self.max_memory,
                     trust_remote_code=self.trusted_models.get(settings.model),
                 )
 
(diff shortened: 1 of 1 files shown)


noctrex:
Closes #80

src/heretic/model.py
```diff
@@ -63,10 +63,19 @@ def __init__(self, settings: Settings):
             print(f"* Trying dtype [bold]{dtype}[/]... ", end="")
 
             try:
+                max_memory = (
+                    {
+                        int(k) if k.isdigit() else k: v
```
p-e-w: Doesn't this happen automatically in Accelerate? And if not, why would we do it? The interface we offer should match what Accelerate does, not add conveniences on top.

src/heretic/model.py
```diff
@@ -63,10 +63,19 @@ def __init__(self, settings: Settings):
...
+                        for k, v in settings.max_memory.items()
+                    }
+                    if settings.max_memory
+                    else None
+                )
                 self.model = AutoModelForCausalLM.from_pretrained(
                     settings.model,
                     dtype=dtype,
                     device_map=settings.device_map,
+                    max_memory=max_memory,
```
p-e-w: This also needs to happen in `reload_model` below, otherwise abliteration will fail.

src/heretic/model.py
noctrex: It doesn't seem that way, when I try without it, I get `Trying dtype bfloat16... Failed (name 'max_memory' is not defined)`

src/heretic/model.py
noctrex: I forgot to add that, silly me

src/heretic/model.py
p-e-w: Well, then we shouldn't support string keys at all, if Accelerate doesn't.

src/heretic/model.py
```diff
@@ -105,10 +114,19 @@ def reload_model(self):
...
         empty_cache()
 
+        max_memory = (
+            {
+                int(k) if k.isdigit() else k: v
+                for k, v in settings.max_memory.items()
+            }
+            if settings.max_memory
+            else None
+        )
```
p-e-w: No need to duplicate this code. You should be able to reuse the `max_memory` value stored in the model object. See the equivalent code for `dtype` a few lines up.

p-e-w (changes requested):

src/heretic/model.py
p-e-w: Ah wait, is TOML the problem, because it doesn't support integer keys?

src/heretic/model.py
noctrex: I'm really sorry for the mess, I'm a scatter brain today

src/heretic/model.py
noctrex: It guess so, when I try to pass it as-is I get the error
`Trying dtype bfloat16... Failed (Device 0 is not recognized, available devices are integers(for GPU/XPU), 'mps', 'cpu' and 'disk')`

p-e-w:
Merged, thank you!

noctrex:
Thank you, for this fantastic project!

PR #101, opened 2026-01-08
fix: make bitsandbytes optional for CPU-only systems

ricyoung:
## Summary

Makes bitsandbytes an optional dependency so Heretic can run on CPU-only systems where bitsandbytes (which requires CUDA) is not available.

## Changes

- Made bitsandbytes import conditional with graceful fallback (`HAS_BITSANDBYTES` flag)
- Added platform condition in pyproject.toml to exclude bitsandbytes on Intel Macs (where it won't work)
- Added helpful error message when 4-bit quantization is requested but bitsandbytes is unavailable
- Added documentation comments in config.default.toml about CPU mode and quantization requirements

## Usage

For CPU-only systems, users can set `device_map = "cpu"` in their config (this was already supported by Accelerate).

## Test plan

- [x] Verified bitsandbytes import is skipped gracefully on systems without CUDA
- [x] Verified clear error message when quantization requested without bitsandbytes
- [x] Verified existing GPU workflow still works unchanged

Closes #12

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -16,10 +16,12 @@ dtypes = [
 ]
 
 # Device map to pass to Accelerate when loading the model.
+# Use "cpu" for CPU-only systems without GPU support.
 device_map = "auto"
 
 # Quantization method to use when loading the model.
 # Options: "none" (no quantization), "bnb_4bit" (4-bit quantization using bitsandbytes).
+# Note: 4-bit quantization requires bitsandbytes, which needs CUDA support.
 quantization = "none"
 
 # Memory limits to impose. 0 is usually your first graphics card.
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -6,8 +6,16 @@ from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any, cast
 
-import bitsandbytes as bnb
 import torch
+
+# bitsandbytes is optional - only available on systems with CUDA support.
+try:
+    import bitsandbytes as bnb
+
+    HAS_BITSANDBYTES = True
+except ImportError:
+    bnb = None  # type: ignore[assignment]
+    HAS_BITSANDBYTES = False
 import torch.nn.functional as F
 from peft import LoraConfig, PeftModel, get_peft_model
 from peft.tuners.lora.layer import Linear
@@ -181,6 +189,12 @@ class Model:
             BitsAndBytesConfig or None
         """
         if self.settings.quantization == QuantizationMethod.BNB_4BIT:
+            if not HAS_BITSANDBYTES:
+                raise RuntimeError(
+                    "4-bit quantization requires bitsandbytes, which is not available. "
+                    "Install it with 'pip install bitsandbytes' (requires CUDA) "
+                    "or set quantization = 'none' in your config."
+                )
             # BitsAndBytesConfig expects a torch.dtype, not a string.
             if dtype == "auto":
                 compute_dtype = torch.bfloat16
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,8 @@ classifiers = [
 ]
 dependencies = [
     "accelerate>=1.10.0",
-    "bitsandbytes>=0.45.0",
+    # bitsandbytes requires CUDA or Apple Silicon. Exclude Intel Macs where it won't work.
+    "bitsandbytes>=0.45.0; (platform_system != 'Darwin') or (platform_machine == 'arm64')",
     "datasets>=4.0.0",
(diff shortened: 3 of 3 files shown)


p-e-w:
> This PR adds a --device CLI option that allows users to explicitly select the compute device, enabling Heretic to run on systems without a GPU.

Isn't this what already happens if you install the CPU version of PyTorch?

ricyoung:
Good question! You're right that CPU-only PyTorch works, but this PR addresses a few additional issues:

1. **bitsandbytes dependency** - Even with CPU PyTorch installed, bitsandbytes will fail to import on systems without CUDA (it's a hard dependency in `pyproject.toml`). This PR makes the import conditional and adds a platform marker to skip it entirely on Intel Macs where it won't even install.

2. **Explicit device control** - Sometimes users want to force CPU even when a GPU is available (e.g., for testing, debugging, because they are crazy? or when GPU memory is limited). `--device cpu` provides that option.

3. **Better UX for edge cases** - When running on a system without GPU support:
   - Shows a clear warning: "Running in CPU-only mode. This will be slow."
   - Automatically disables quantization with an explanation
   - Prefers `float32` dtype for maximum compatibility

The main motivation was the original issue #12 where bitsandbytes caused installation/import failures on Intel Macs. Rather than just patching that specific case, this provides a more general solution.

That said, if you think the simpler approach (just making bitsandbytes optional) is preferable, I'm happy to slim down the PR to just that change.

src/heretic/model.py
```diff
@@ -75,7 +86,13 @@ def __init__(self, settings: Settings):
         if self.settings.evaluate_model is not None:
             self.trusted_models[settings.evaluate_model] = settings.trust_remote_code
 
-        for dtype in settings.dtypes:
+        # For CPU mode, prefer float32 as it's most compatible
+        dtypes_to_try = settings.dtypes
+        if settings.device == DeviceType.CPU:
+            # Put float32 first for CPU, as it's most compatible
+            dtypes_to_try = ["float32"] + [d for d in dtypes_to_try if d != "float32"]
```
p-e-w: I don't like this. There is a setting for dtypes, and the program should use it as specified. Not magically do something else in a certain configuration. If the user wants something else, they can configure that.

Note that many newer CPUs have native support for BF16, so this often does the wrong thing anyway.

config.default.toml
```diff
@@ -15,8 +16,17 @@ dtypes = [
     "float32",
 ]
 
+# Device to run inference on.
+# Options:
+#   "auto" - Automatically detect best available device (GPU if available, else CPU)
+#   "cpu"  - Force CPU-only mode (slow but works anywhere)
+#   "cuda" - Force NVIDIA CUDA GPU
+#   "mps"  - Force Apple Metal Performance Shaders (Apple Silicon)
+device = "auto"
```
p-e-w: What exactly does this setting offer that the `device_map` setting doesn't?

You can set `device_map` to `cpu` already, and it will offload the entire model to CPU RAM. This is explicitly documented in Accelerate.

src/heretic/model.py
ricyoung: I guess - some CPU have support for BF16. 

Intel Xeon: Support started with 3rd Generation Intel Xeon Scalable Processors (Cooper Lake) and continues in later generations, utilizing the AVX-512 BF16 extensions.

AMD Zen: Modern AMD processors based on the Zen architecture, including certain AMD EPYC (Zen 4) and Ryzen AI series, incorporate BF16 support.

Good point

config.default.toml
```diff
@@ -16,10 +16,12 @@ dtypes = [
 ]
 
 # Device map to pass to Accelerate when loading the model.
+# Use "cpu" for CPU-only systems without GPU support.
 device_map = "auto"
 
 # Quantization method to use when loading the model.
 # Options: "none" (no quantization), "bnb_4bit" (4-bit quantization using bitsandbytes).
+# Note: 4-bit quantization requires bitsandbytes, which needs CUDA support.
```
p-e-w, line 24: That contradicts the bnb docs: https://huggingface.co/docs/transformers/en/quantization/bitsandbytes

> bitsandbytes is supported on NVIDIA GPUs for CUDA versions 11.8 - 13.0, Intel XPU, Intel Gaudi (HPU), **and CPU.**

There are several mentions on that page of CPU-only usage.

spikymoth: Note that ROCm-via-HIP is also (partially) supported: https://github.com/bitsandbytes-foundation/bitsandbytes#legend

But yeah it seems like CPUs should be supported as well, except for "8-bit Optimizers" (whatever that means).

p-e-w:

PR #103, opened 2026-01-08
feat: add support for plain text files as prompt datasets

### LINKED ISSUE #98: add option to custom datasets via txt file
could you add the option to use a custom dataset in *.txt format?
(each line is a prompt)

ricyoung:
## Summary
- Adds support for loading prompts from plain text files (one prompt per line)
- Provides a simpler alternative to creating HuggingFace datasets for custom prompts
- Slice notation works (e.g., `split = "[:100]"` for first 100 lines)

## Changes
- Modified `load_prompts()` in `utils.py` to detect and parse plain text files
- Added `_get_split_slice()` helper function for robust split parsing (reuses `ReadInstruction`)
- Made `split` and `column` fields optional in `DatasetSpecification` (ignored for text files)
- Added documentation in `config.default.toml` explaining the dataset options

## Example usage
```toml
[bad_prompts]
dataset = "/path/to/my_prompts"
split = "[:100]"  # Optional: use first 100 lines
```

## Test plan
- [x] Tested loading full text file
- [x] Tested slice notation (`[:3]`, `[1:3]`, `[:2]`)
- [x] Verified prefix/suffix options work correctly
- [x] Verified existing HF dataset loading still works (fallback path)
- [x] Verified proper error messages when split/column missing for HF datasets

Closes #98

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -109,15 +109,14 @@ system_prompt = "You are a helpful assistant."
 # The dataset can be:
 #   - A Hugging Face dataset ID (e.g., "mlabonne/harmless_alpaca")
 #   - A local directory containing a dataset
-#   - A plain text file (.txt) with one prompt per line
+#   - A plain text file with one prompt per line
 #
-# Text file format:
+# For text files:
 #   - One prompt per line (UTF-8 encoding)
 #   - Empty lines and whitespace-only lines are ignored
 #   - Leading/trailing whitespace is stripped from each prompt
-#   - The "column" field is ignored for text files
-#   - Slice notation works: split = "[:100]" uses first 100 lines,
-#     split = "[50:150]" uses lines 50-149
+#   - The "split" and "column" fields are optional (ignored for text files)
+#   - Slice notation works in "split": split = "[:100]" uses first 100 lines
 [good_prompts]
 dataset = "mlabonne/harmless_alpaca"
 split = "train[:400]"
--- a/config.default.toml
+++ b/config.default.toml
@@ -106,6 +106,11 @@ refusal_markers = [
 system_prompt = "You are a helpful assistant."
 
 # Dataset of prompts that tend to not result in refusals (used for calculating refusal directions).
+# The dataset can be:
+#   - A Hugging Face dataset ID (e.g., "mlabonne/harmless_alpaca")
+#   - A local directory containing a dataset
+#   - A plain text file (.txt) with one prompt per line
+# For text files, the split notation can still be used (e.g., split = "[:100]" for first 100 lines).
 [good_prompts]
 dataset = "mlabonne/harmless_alpaca"
 split = "train[:400]"
--- a/config.default.toml
+++ b/config.default.toml
@@ -110,7 +110,14 @@ system_prompt = "You are a helpful assistant."
 #   - A Hugging Face dataset ID (e.g., "mlabonne/harmless_alpaca")
 #   - A local directory containing a dataset
 #   - A plain text file (.txt) with one prompt per line
-# For text files, the split notation can still be used (e.g., split = "[:100]" for first 100 lines).
+#
+# Text file format:
+#   - One prompt per line (UTF-8 encoding)
+#   - Empty lines and whitespace-only lines are ignored
+#   - Leading/trailing whitespace is stripped from each prompt
+#   - The "column" field is ignored for text files
+#   - Slice notation works: split = "[:100]" uses first 100 lines,
+#     split = "[50:150]" uses lines 50-149
 [good_prompts]
 dataset = "mlabonne/harmless_alpaca"
 split = "train[:400]"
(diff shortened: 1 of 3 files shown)


ricyoung:
Applied @gemini-code-assist's suggestion to use `ReadInstruction` from the datasets library instead of manual string parsing. This provides better error handling and supports percentage-based slicing (e.g., `[:40%]`).

src/heretic/utils.py
```diff
@@ -149,38 +149,57 @@ def load_prompts(
...
-        if Path(path, DATASET_STATE_JSON_FILENAME).exists():
-            # Dataset saved with datasets.save_to_disk; needs special handling.
-            # Path should be the subdirectory for a particular split.
-            dataset = load_from_disk(path)
-            assert not isinstance(dataset, DatasetDict), (
-                "Loading dataset dicts is not supported"
-            )
-            # Parse the split instructions.
+    # Support for plain text files (one prompt per line)
+    if path.endswith(".txt") and os.path.isfile(path):
```
p-e-w:
```suggestion
    if os.path.isfile(path):
```

I don't think we should require a specific extension, as plain text files are often stored without extensions on Unix systems.

src/heretic/utils.py
```diff
@@ -149,38 +149,57 @@ def load_prompts(
...
-            # Path is a local directory.
-            dataset = load_dataset(
-                path,
-                split=split_str,
-                # Don't require the number of examples (lines) per split to be pre-defined.
-                verification_mode=VerificationMode.NO_CHECKS,
-                # But also don't use cached data, as the dataset may have changed on disk.
-                download_mode=DownloadMode.FORCE_REDOWNLOAD,
-            )
+            prompts = prompts[abs_instruction.from_ : abs_instruction.to]
```
p-e-w: This duplicates code below, and should be moved to a common helper function.

config.default.toml
```diff
@@ -106,6 +106,18 @@ refusal_markers = [
...
+# The dataset can be:
+#   - A Hugging Face dataset ID (e.g., "mlabonne/harmless_alpaca")
+#   - A local directory containing a dataset
+#   - A plain text file (.txt) with one prompt per line
+#
+# Text file format:
+#   - One prompt per line (UTF-8 encoding)
+#   - Empty lines and whitespace-only lines are ignored
+#   - Leading/trailing whitespace is stripped from each prompt
+#   - The "column" field is ignored for text files
```
p-e-w: It shouldn't be required then, and neither should the `split` field.

p-e-w (changes requested):

src/heretic/utils.py
```diff
@@ -142,45 +142,78 @@ class Prompt:
...
+            prompts = [line.strip() for line in f if line.strip()]
+
+        # Apply split specification if provided.
+        # We treat the list of prompts as a single split, so the name is arbitrary.
+        if split_str is not None:
+            try:
+                start, end = _get_split_slice(split_str, len(prompts), "_")
+                prompts = prompts[start:end]
+            except (ValueError, IndexError):
+                # If split_str doesn't contain valid slice notation, use all prompts.
```
p-e-w, line 175: No, it should throw an error. If the user put something in there they expect it to be used. If it can't be, we need to tell them.

p-e-w: Simply remove that `try/except` block. That's what we do for the directory case as well.

p-e-w:
This looks good apart from the error thing.

kalle07:
question?
now i can simple change that line
```
# Dataset of prompts that tend to not result in refusals (used for calculating refusal directions).
[good_prompts]
dataset = "mlabonne/harmless_alpaca"
```
to a local ascii txt file where every line is a new prompt?
it would be counted so i can see if all is right?

p-e-w:
Any update here? This would be valuable to have and it's really close to getting merged.

kalle07:
my approach is manual
arrow to txt 
-> add some lines
txt to arrow

```
from datasets import Dataset, concatenate_datasets
import glob

# Load single Arrow file or all shards (e.g., data-00000-of-00001.arrow)
file_path = "harmful_behaviors-train.arrow"  # or glob.glob("data-*.arrow") for multiple
dataset = Dataset.from_file(file_path)  # Handles Arrow shards automatically

# If multiple shards:
# dataset = concatenate_datasets([Dataset.from_file(f) for f in glob.glob("data-*.arrow")])

# Set format for easy access (optional)
dataset.set_format("pandas")

# Extract 'text' column and save as ASCII text (one item per line)
with open("output.txt", "w", encoding="utf-8") as f:
    for text in dataset["text"]:
        f.write(str(text) + "\n")

print("Converted to output.txt")
```

```
from datasets import Dataset, DatasetDict

# Read txt file (one example per line)
with open("output.txt", "r", encoding="utf-8") as f:
    lines = [line.strip() for line in f if line.strip()]

# Create dataset
train_dataset = Dataset.from_dict({"text": lines})

# Create DatasetDict with train split
dataset_dict = DatasetDict({
    "train": train_dataset
})

# Save to disk (creates Arrow files internally)
dataset_dict.save_to_disk("harmful_dataset")

print("Dataset saved to ./harmful_dataset")
```

p-e-w:

Issue #95, opened 2026-01-02
State save possible?


Hotohori:
I wonder if it would be possible to save states between trials so you can continue if anything stops the app or save trials directly which are under a set value of KL/refusal?

I tried to use my RTX 4090 to run Heretic on Mistral Nemo 12B (no quant), but because 24GB VRAM is a bit too less for the full model some parts needed to be on CPU it and so it runs close on full VRAM and for reasons it failed after over 2h with CUDA out of memory on trial 35 and with that you lose everything.

Vinay-Umrethe:
that's a cracking idea actually and something i’d considered before but i haven't made any PRs for it or gave much attention... I know optuna uses sqlite to save data in a .db file but I’m not entirely sure if heretic actually makes use of that built in feature or not

Hotohori:
Well, it would be great, because now it was even more annoying. I tried the actual branch to set max_memory. It worked good until the end. When I should selected the trial I want to save and give it a directory name > CUDA out of memory. Closed all to make as much VRAM free as possible, didn't help and in the end the app crashed totally with a CUDA out of memory and already all the work for nothing.

So, yes, a save of the trials under a specific rate would be great.

spikymoth:
I actually implemented something like this locally, just extracted it into a branch:

https://github.com/spikymoth/heretic/tree/save-and-resume (the commits: https://github.com/spikymoth/heretic/commit/d0d9092a3046e5d517d9b67a87866b11d0ee3274 & https://github.com/spikymoth/heretic/commit/4ac0d54c83367969a9526614d62c64cf4d21a244)

It's a bit rough:
1. I used the JSON-based storage ([JournalFileBackend](https://optuna.readthedocs.io/en/v4.0.0/reference/generated/optuna.storages.journal.JournalFileBackend.html#optuna.storages.journal.JournalFileBackend)), so it's more easily editable but maybe also more fragile than sqlite
2. It always stores the active study in `log/study.jsonl` and uses fixed internal name `my_study`
3. It doesn't do any checking to see if the logged study corresponds to the current session (you can run with different settings or on a completely different model and it'll let you resume anyway)
4. There's no file locking and it's append-only during runtime, so if you wipe it accidentally, it's gone forever
5. If you use Ctrl+C to cancel early, the last trial is ignored but still counted toward the total (this is pretty minor)

So it needs more work before it can be called production ready, but it's usable if you know the limitations.

If anyone wants to use it as a starting point for a pull request, you have my blessing :)

[...]

PR #76, opened 2025-12-08
feat: add continuous optimization option(latest changes updated)

Vinay-Umrethe:
# SAME AS THE PREVIOUS #69 BUT SINCE MANY UPDATES ARE GOING ON CURRENTLY, TO ENSURE NO CONFLICTS DID IT AGAIN CLEANLY...

this PR introduces a **"Continuous Optimization"** feature that allows users to extend the ablation process by running additional trials if the **initial** results are **unsatisfactory**. The optimization and trial selection logic in `src/heretic/main.py` has been wrapped in a loop...

After initial batch of trials completes, the user is presented with a normal menu with **[NEW] option**: 
```bash
"Continue optimization (run more trials)". 
```

If selected, the user can specify the number of additional trials to run. The existing Optuna `study `object is reused, ensuring that the TPE sampler continues to learn from the history of previous trials and explores parameter space more effectively, rather than starting from scratchh...

enabling a more efficient workflow for finding optimal ablation parameters when the default number of trials is insufficient.

### Test Conducted: (On Colab as my preference)
```bash
Loading model Qwen/Qwen2.5-0.5B-Instruct...
Ok
* Transformer model with 24 layers
* Abliterable components:
  * attn.o_proj: 1 matrices per layer
  * mlp.down_proj: 1 matrices per layer

Loading good prompts from mlabonne/harmless_alpaca...
* 400 prompts loaded

Loading bad prompts from mlabonne/harmful_behaviors...
* 400 prompts loaded

Loading good evaluation prompts from mlabonne/harmless_alpaca...
* 100 prompts loaded
* Obtaining first-token probability distributions...

Loading bad evaluation prompts from mlabonne/harmful_behaviors...
* 100 prompts loaded
* Counting model refusals...
* Initial refusals: 83/100

Calculating per-layer refusal directions...
* Obtaining residuals for good prompts...
* Obtaining residuals for bad prompts...

Running trial 1 of 3...
* Parameters:
  * direction_index = per layer
  * attn.o_proj.max_weight = 1.42
  * attn.o_proj.max_weight_position = 14.07
  * attn.o_proj.min_weight = 0.88
  * attn.o_proj.min_weight_distance = 9.21
  * mlp.down_proj.max_weight = 0.94
  * mlp.down_proj.max_weight_position = 17.43
  * mlp.down_proj.min_weight = 0.87
  * mlp.down_proj.min_weight_distance = 7.03
* Reloading model...
* Abliterating...
* Evaluating...
[...]

PR #106, opened 2026-01-15
feat: Allow study progress to be saved & resumed

anrp:
Fixes https://github.com/p-e-w/heretic/issues/95
Followup/improvement of https://github.com/spikymoth/heretic/tree/save-and-resume

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -171,7 +171,7 @@ class Settings(BaseSettings):
 
     study_checkpoint_dir: str = Field(
         default="checkpoints",
-        description="Directory to save and load study progress to/from",
+        description="Directory to save and load study progress to/from:",
     )
 
     refusal_markers: list[str] = Field(
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -514,11 +514,9 @@ def run():
         except KeyboardInterrupt:
             # Stop the study gracefully on Ctrl+C.
             trial.study.stop()
-
             raise TrialPruned()
 
     study = optuna.create_study(
-        study_name="heretic",
         sampler=TPESampler(
             n_startup_trials=settings.n_startup_trials,
             n_ei_candidates=128,
@@ -532,17 +530,17 @@ def run():
     study.set_user_attr("settings", settings.model_dump_json())
     study.set_user_attr("finished", False)
 
-    def get_trials_completed():
+    def count_completed_trials() -> int:
         # Count number of complete trials to compute trials to run.
         return sum([(1 if t.state == TrialState.COMPLETE else 0) for t in study.trials])
 
-    start_index = trial_index = get_trials_completed()
+    start_index = trial_index = count_completed_trials()
     if start_index > 0:
         print("Resuming existing study.")
 
     try:
         study.optimize(
-            objective_wrapper, n_trials=settings.n_trials - get_trials_completed()
+            objective_wrapper, n_trials=settings.n_trials - count_completed_trials()
         )
 
     except KeyboardInterrupt:
@@ -551,7 +549,7 @@ def run():
         # defined in objective_wrapper above.
         pass
 
-    if get_trials_completed() == settings.n_trials:
+    if count_completed_trials() == settings.n_trials:
         study.set_user_attr("finished", True)
 
     while True:
@@ -640,11 +638,11 @@ def run():
                 try:
                     study.optimize(
                         objective_wrapper,
-                        n_trials=settings.n_trials - get_trials_completed(),
+                        n_trials=settings.n_trials - count_completed_trials(),
                     )
                 except KeyboardInterrupt:
                     pass
-                if get_trials_completed() == settings.n_trials:
+                if count_completed_trials() == settings.n_trials:
                     study.set_user_attr("finished", True)
                 break
 


src/heretic/main.py
```diff
@@ -452,17 +458,78 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+    storage = JournalStorage(backend)
+
+    # Hash the settings that matter to be able to resume a study
+    study_components = [
+        f"model={settings.model}",
+        f"refusal_markers={hashlib.sha256(','.join(settings.refusal_markers).encode('utf-8')).hexdigest()}",
+        f"kl_divergence_target={settings.kl_divergence_target}",
+        f"kl_divergence_scale={settings.kl_divergence_scale}",
+        "good_prompts=" + hashlib.sha256(','.join([f"{x.system},{x.user}" for x in good_prompts]).encode("utf-8")).hexdigest(),
+        "bad_prompts=" + hashlib.sha256(','.join([f"{x.system},{x.user}" for x in bad_prompts]).encode("utf-8")).hexdigest(),
```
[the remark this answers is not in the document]
anrp: 1) is a hallucination

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
     try:
-        study.optimize(objective_wrapper, n_trials=settings.n_trials)
+        remaining_trials = settings.n_trials - start_index
+        if remaining_trials > 0 and random_trials_to_run > 0:
+            if start_index > 0:
+                print(f"Running additional {random_trials_to_run} random trials")
+            else:
+                print(f"Running initial {random_trials_to_run} random trials")
+            saved_sampler = study.sampler
+            study.sampler = optuna.samplers.RandomSampler()
```
p-e-w: Isn't this built into Optuna? What is the purpose of the "resume study" functionality if we need to manually tell it which sampler to use?

src/heretic/main.py
```diff
@@ -412,7 +418,7 @@ def objective(trial: Trial) -> tuple[float, float]:
             )
 
         trial.set_user_attr("direction_index", direction_index)
-        trial.set_user_attr("parameters", parameters)
+        trial.set_user_attr("parameters", {k: asdict(v) for k, v in parameters.items()})
```
p-e-w, line 479: What is the purpose of this change?

src/heretic/config.py
```diff
@@ -168,6 +168,16 @@ class Settings(BaseSettings):
         description="Number of trials that use random sampling for the purpose of exploration.",
     )
 
+    study_checkpoint_file: str | None = Field(
```
p-e-w: Wouldn't it be better to automatically derive the checkpoint file name from the model name, so that multiple studies can be stopped and resumed without micromanaging the config?

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+        remaining_trials = settings.n_trials - start_index
+        if remaining_trials > 0 and random_trials_to_run > 0:
+            if start_index > 0:
+                print(f"Running additional {random_trials_to_run} random trials")
+            else:
+                print(f"Running initial {random_trials_to_run} random trials")
+            saved_sampler = study.sampler
+            study.sampler = optuna.samplers.RandomSampler()
+            study.optimize(objective_wrapper, n_trials=random_trials_to_run)
+            study.set_user_attr("startup_trials", random_trials_to_run + previous_startup_trials)
```
p-e-w: What does this line do?

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+        random_trials_to_run = settings.n_startup_trials - previous_startup_trials
+    else:
+        random_trials_to_run = 0
+
     try:
-        study.optimize(objective_wrapper, n_trials=settings.n_trials)
+        remaining_trials = settings.n_trials - start_index
+        if remaining_trials > 0 and random_trials_to_run > 0:
+            if start_index > 0:
+                print(f"Running additional {random_trials_to_run} random trials")
```
p-e-w: This should not be displayed in the frontend. Sampling is an implementation detail that the user can do nothing with.

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+        settings.study_checkpoint_file = f"log/{settings.model.replace('/', '--')}_" + f"{datetime.datetime.now(datetime.UTC).isoformat(timespec='seconds')}.jsonl"
+
+    os.makedirs(os.path.dirname(settings.study_checkpoint_file), exist_ok=True)
+    backend = JournalFileBackend(settings.study_checkpoint_file)
+    storage = JournalStorage(backend)
+
+    # Hash the settings that matter to be able to resume a study.
+    study_components = [
+        f"model={settings.model}",
+        f"refusal_markers={hashlib.sha256(','.join(settings.refusal_markers).encode('utf-8')).hexdigest()}",
```
p-e-w: Instead of all that complicated logic, wouldn't it be much easier to just serialize the entire `Settings` object to JSON and compute a hash on that?

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
             trial.study.stop()
             raise TrialPruned()
 
+    if settings.study_checkpoint_file is not None:
+        resume = os.path.exists(settings.study_checkpoint_file)
+        if not settings.study_autoresume and resume:
+            choice = prompt_select("Resume previous study?", ["yes", "no"])
+            if choice == "no":
+                print("Exiting (specify a different study checkpoint file)")
```
p-e-w: This is bad UX. If the user says no to resuming the previous study, the program should start a new study, not quit and discard all the progress (e.g. batch size determination) made already.

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+                return
+    else:
+        settings.study_checkpoint_file = f"log/{settings.model.replace('/', '--')}_" + f"{datetime.datetime.now(datetime.UTC).isoformat(timespec='seconds')}.jsonl"
```
p-e-w: The directory should be configurable, just like `residual_plot_path` is.

src/heretic/utils.py
```diff
@@ -241,7 +241,7 @@ def get_trial_parameters(trial: Trial) -> dict[str, str]:
     )
 
     for component, parameters in trial.user_attrs["parameters"].items():
-        for name, value in asdict(parameters).items():
+        for name, value in parameters.items():
```
p-e-w, line 244: What is the purpose of this change?

src/heretic/main.py
spikymoth: I ran into this while working on it for my personal use. Apparently Optuna doesn't care what sampler is used and will happily switch to the default `NSGAIISampler` for multi-objective optimization, essentially ignoring the earlier results since they don't have a generation parameter. I agree that it seems like a silly oversight in the logging.

src/heretic/main.py
spikymoth: Optuna doesn't support serializing classes to JSON; this sidesteps that by storing the parameters as a dict.

src/heretic/utils.py
spikymoth: This is the inverse of the other change - since the parameters are stored as a dict, we don't need to call `asdict()` on them.

src/heretic/main.py
spikymoth: I didn't write this part (my hacky version didn't do any checks at all), but I think there are a lot of settings that can be whitelisted to allow the study to resume even if you change them (for example, resuming with additional logging).

But I think we can define which settings to serialize directly by setting `exclude=True` on each `Field()` that we *don't* care about (they're serialized by default). Then call `dumps(settings.model_dump(), sort_keys=True)` to serialize the object (`settings.model_dump_json()` doesn't sort the keys so it's not deterministic), and hash the result.

src/heretic/main.py
p-e-w: But then how will it know to start using `TPESampler` if we just push and pop `study.sampler` here?

Are you sure we're not missing something here? Why would this not work?

src/heretic/main.py
p-e-w: I don't like the idea of using the `exclude` parameter on the field to drive this one use case (we might use serialization elsewhere also).

The whole approach seems overengineered to me. If a user absolutely wants to change the settings, they may have to live with starting the study from scratch. Keeping a list of fields that do, or do not, affect the outcome is a maintenance nightmare.

src/heretic/main.py
spikymoth: Oh, actually this part is a little different from what I wrote: I'm not sure whether it automatically skips the `n_initial_trials` by looking at the trial index. I kind of assumed that it would, and AI assistants seem to agree, so this part may not be necessary.

p-e-w: I just tried to investigate this in some more depth and it appears that Optuna's storage system is a simple replay log: The outcomes of the completed trials are stored, and after you initialize the study, Optuna forwards it using the log. So all of that logic should indeed be unnecessary.

src/heretic/main.py
```diff
@@ -570,7 +636,10 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
             model.abliterate(
                 refusal_directions,
                 trial.user_attrs["direction_index"],
-                trial.user_attrs["parameters"],
+                {
+                    k: AbliterationParameters(**v)
+                    for k, v in trial.user_attrs["parameters"].items()
+                },
             )
```
p-e-w, line 667: Does the checkpoint file need to be deleted after the study completes, or does Optuna take care of that automatically?

src/heretic/main.py
```diff
@@ -452,17 +458,77 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+        f"refusal_markers={hashlib.sha256(','.join(settings.refusal_markers).encode('utf-8')).hexdigest()}",
+        f"kl_divergence_target={settings.kl_divergence_target}",
+        f"kl_divergence_scale={settings.kl_divergence_scale}",
+        "good_prompts=" + hashlib.sha256(','.join([f"{x.system},{x.user}" for x in good_prompts]).encode("utf-8")).hexdigest(),
+        "bad_prompts=" + hashlib.sha256(','.join([f"{x.system},{x.user}" for x in bad_prompts]).encode("utf-8")).hexdigest(),
+        f"quantization={settings.quantization}",
+        f"dtype={model.model.dtype}",
+    ]
+
+    study_name = "heretic_study_" + "_".join(study_components)
```
p-e-w: I think the prefix (`heretic_study_`) is unnecessary here, because that name is not user-visible AFAICT.

src/heretic/main.py
spikymoth: It doesn't get deleted automatically. But I would make deleting it optional, as being able to resume later can be pretty handy in my experience (e.g. for continued optimization after trying some different settings).

p-e-w: But how would you resume a study that is already complete? That would require changing parameters, which absolutely does change the outcome, so this would fail to hash to the same value even with the current approach.

I think what we want is a system that can (and does, automatically) resume where the user left off if the program crashed or the user shut it down for the day. Nothing more and nothing less. The more edge cases we try to support, the more difficult it becomes to manage for that central use case.

spikymoth: Well, you could pass a higher value for `n_trials` (if it were whitelisted), or even just reload the completed trial to look at the results again, save/upload/chat with models, etc. (since it only needs the model parameters to apply the abliteration). I guess this could be a plugin-provided extension if the plugin system allows for it.

src/heretic/main.py
anrp: https://github.com/optuna/optuna/blob/e5d038f482b836396b22e72b6edde0f003ebdcea/optuna/samplers/_tpe/sampler.py#L465-L469

All this to basically just allow a different n_startup_trials, because that's how TPESampler decides internally i.e. if I just initialized with different parameters it wouldn't run more random samples as long as the total samples previously was at least the new random samples.

p-e-w: I don't think that code carries its weight. Not having this logic more closely matches what I would intuitively expect to happen.

src/heretic/main.py
anrp: Uses the study user attributes as settings storage, to allow it to know how many new random samples to take. Although, seems like that's not a wanted capability?

src/heretic/main.py
anrp: Sure but in that case we should always use a unique (file)name for the study to lessen foot-gun. If you wanted to use a name derived from model name, then jamming the settings in to the study name may allow that to be seamless... I don't know if you can have multiple studies that don't conflict in a single .jsonl file.

src/heretic/main.py
anrp: Fair, but then n_startup_trials is a study-invariant setting, so need to figure out which way to go with setting and resume.

p-e-w: It seems reasonable to me that you can't change the number of startup trials after starting the study.

In fact, I'm not sure it should be possible to change *any* setting after starting a study. Doing so seems like an invitation for all kinds of hard-to-debug issues.

src/heretic/main.py
p-e-w: But we need to be able to automatically resume an incomplete study, and if we just choose a random filename (e.g. a UUID), I don't know how we would find the correct file to resume. The user should *not* have to locate the file themselves of course.

src/heretic/main.py
p-e-w: Yeah, that's definitely too much complexity for the ability to change that one setting later.

anrp:
Ultimately from all of this discussion - would you like "configuration option for directory", "*all* settings saved in study name so that it can't restart from a different place" and "no logic for more random tries"? That converts this to a resume-only capability (and you can extend the number of tries interactively, that's it)? (Maybe just don't include n_tries in the study name, and read from study file (resuming) or command line (initial)?

p-e-w:
We need to decide what we actually want here. My idea was to have a system that can recover from a crash by resuming where it left off (e.g. after another program used up VRAM and led to an OOM), or allow the user to stop the run and continue the next day. In that case, it's obvious what should happen: All settings are restored, and the trials continue from the cutoff point.

But it seems that you and @spikymoth have a more ambitious vision: That the user should be able to stop the run and resume it *with different settings.* I agree that this ability would be nice to have in principle, but it seems very difficult to implement in a way that's correct, non-confusing, and maintainable.

anrp:
The only setting that really makes sense to resume with differently is the n_trials, and that can effectively already be overriden with a resume-only capability by the interactive menu option. I would be fine with that, because it does simplify every other part of this.

p-e-w:
Here's a sketch of how I think a reasonable implementation might look:

1. The study progress is journaled to a file with the same name as the model. The complete settings object is stored in there.
2. When Heretic is started and a journal file matching the model name already exists, the user is offered the choice between resuming the study or starting a new study.
3. If the user chooses to resume the study, the entire settings object is replaced with the one from the journal file.
4. If the user chooses to start a new study, the existing journal file is deleted and replaced with a journal file corresponding to the current run.
5. When the run is complete, the journal file is deleted.

anrp:
1-4 SGTM but I'd like to push back on 5 being automatic, since it's sometimes useful to just be able to drop back in to the chat interface to test things i.e. I'd like to keep the few-KB result file which represents the $hours of computation. Thoughts?

spikymoth:
I think a "Delete study log? (y/n)" question at the end if you select "None (exit program)" would make sense, personally. And I agree that `n_trials` is the main setting that makes sense to allow changing between sessions (which should only require setting `exclude=True` on that `Field()` to avoid serialization). There's a case to be made for some other settings, but I think `n_trials` is the main one for just picking up a promising study again later.

p-e-w:
> 1-4 SGTM but I'd like to push back on 5 being automatic, since it's sometimes useful to just be able to drop back in to the chat interface to test things i.e. I'd like to keep the few-KB result file which represents the $hours of computation.

Okay, but how exactly will this work? Let's say the study is complete and we keep the journal. Now Heretic is run again with the same model, and the user chooses to "resume" the study. But the study is already complete, and the number of completed trials is equal to the number of trials to complete. Now what? We just tell them that the run they just started is already complete, and show the trial selection menu with the Pareto front?

> And I agree that n_trials is the main setting that makes sense to allow changing between sessions (which should only require setting exclude=True on that Field() to avoid serialization).

But we already have that functionality. When the Pareto front is displayed, one of the options is "Continue optimization (run more trials)". This was implemented in #76.

anrp:
> [...] We just tell them that the run they just started is already complete, and show the trial selection menu with the Pareto front?

is necessary if you want to

> [...] "Continue optimization (run more trials)". This was implemented in #76.

after deciding to exit.
Maybe print a distinct message about no work actually happening, but basically, yes.

p-e-w:
Ok, how about this:

Instead of asking the user whether to delete the journal file when they exit (which interrupts the exit process and might be the wrong time to make that decision), we prompt them when the program starts?

**If the previous run is complete:**

```
You have already processed this model. How would you like to proceed?
[1] Show the results from the previous run, allowing you to export models, or to run additional trials.
[2] Ignore the previous run and start from scratch. This will delete the checkpoint file and all results from the previous run.
```

**If the previous run is incomplete:**

```
You have already processed this model, but the run was interrupted. How would you like to proceed?
[1] Continue the previous run from where it stopped.
[2] Ignore the previous run and start from scratch. This will delete the checkpoint file and all results from the previous run.
```

anrp:
I like that, basically moves the delete question to startup time. Implemented, PTAL.

src/heretic/config.py
```diff
@@ -168,6 +168,11 @@ class Settings(BaseSettings):
         description="Number of trials that use random sampling for the purpose of exploration.",
     )
 
+    study_checkpoint_dir: str | None = Field(
+        default="log",
```
p-e-w: We're going to have something called "logger plugins" soon, and to avoid confusion I'd prefer to use the name `log` for those. I suggest `checkpoints` as the name for this directory.

src/heretic/config.py
```diff
@@ -168,6 +168,11 @@ class Settings(BaseSettings):
         description="Number of trials that use random sampling for the purpose of exploration.",
     )
 
+    study_checkpoint_dir: str | None = Field(
```
p-e-w: `None` is not an acceptable value for this.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
     # Silence the warning about multivariate TPE being experimental.
     warnings.filterwarnings("ignore", category=ExperimentalWarning)
 
+    study_checkpoint_file = os.path.join(
+        str(settings.study_checkpoint_dir), settings.model.replace("/", "--") + ".jsonl"
```
p-e-w: `str()` is unnecessary.

p-e-w: This doesn't work on Windows when loading from local paths. First, `\` is also a path separator there, and also, there are drive letters.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+    )
+
+    os.makedirs(os.path.dirname(study_checkpoint_file), exist_ok=True)
```
p-e-w:
```suggestion
    os.makedirs(settings.study_checkpoint_dir, exist_ok=True)
```

I think those are the same, by construction.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+    backend = JournalFileBackend(study_checkpoint_file)
+    storage = JournalStorage(backend)
```
p-e-w, line 232: The docs mention a potential error on Windows systems. I don't use Windows though. Are you able to verify that that error doesn't occur?

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+
+    can_resume = False
+    is_finished = False
+    try:
+        existing_study_id = storage.get_study_id_from_name("heretic")
+    except KeyError:
```
p-e-w: Is that what happens if the file doesn't exist?

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+        existing_study_id = None
+
+    if existing_study_id is not None:
+        # A study is in here. Check if it's finished.
+        maybe_previous_settings = storage.get_study_user_attrs(existing_study_id).get(
+            "settings", None
+        )
+        if maybe_previous_settings is not None:
```
p-e-w: How can the `settings` key be `None` if the study itself is there?

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+            try:
+                old_settings = Settings.model_validate_json(maybe_previous_settings)
+                can_resume = True
+            except ValidationError:
+                can_resume = False
```
p-e-w: No. We throw an error. Never swallow errors. The user needs to see this.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+            )
+            choices.append(
+                Choice(
+                    title="Show the results from the previous run, allowing you to export models, or to run additional trials.",
+                    value="continue",
+                )
+            )
+        else:
+            print(
+                "[green]You have already processed this model, but the run was interrupted. How would you like to proceed?[/green]"
```
p-e-w: Why is this green but the other one is yellow? Doesn't the other way round make more sense as this is the case where something went wrong?

Also, please use the syntax `[/]` for closing the tag as it makes it easier to change things later.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+                    value="continue",
+                )
+            )
+        choices.append(
+            Choice(
+                title="Ignore the previous run and start from scratch. This will delete the checkpoint file and all results from the previous run.",
+                value="restart",
+            )
+        )
+        choice = prompt_select("What do you want to do?", choices)
```
p-e-w: This seems to be redundant with the above messages, which already ask "How would you like to proceed?".

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+
+        if choice == "continue":
+            settings = old_settings
+            new_study = False
+        else:
+            new_study = True
```
p-e-w: Careful. In the prompt above, the user can also press Ctrl+C, which returns `None` I think. But your logic means that this results in the existing checkpoint being deleted. I don't think that's the expected output of cancelling that prompt.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+    else:
+        new_study = True
```
p-e-w: We don't need this branch. If we are unable to resume, then there's nothing to delete, and the deletion logic can just be pulled into the place where we set `new_study = True` above. I think this should get rid of the `new_study` variable entirely.

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+
+    if new_study:
+        if existing_study_id is not None:
+            storage.delete_study(existing_study_id)
```
p-e-w: Does this delete the file, or does it delete the study data from the file while leaving the file in place?

src/heretic/main.py
```diff
@@ -453,22 +530,42 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
+        elif remaining_trials < 0:
+            settings.n_trials = len(study.trials)
+
     except KeyboardInterrupt:
         # This additional handler takes care of the small chance that KeyboardInterrupt
         # is raised just between trials, which wouldn't be caught by the handler
         # defined in objective_wrapper above.
         pass
 
+    if len(study.trials) == settings.n_trials:
```
p-e-w: IIRC, `study.trials` includes incomplete (aborted) trials. If so, this gives the wrong result if the study is aborted while the last trial runs.

spikymoth: That's right - every time the user aborts with Ctrl+C, you get an incomplete trial in `study.trials` which also gets added to the log. I think a crash doesn't do this as it stops before anything is written for the incomplete trial.

Either way, I'm not sure what the best way to solve this is - just `delete` the last trial before resuming if it's incomplete? User aborts should be the only source of them, I think.

Trying to clean up incomplete trials in the middle could get tricky (e.g. having to adjust all subsequent indices) and I'm not sure it's worth it compared to just having a few less results at the end.

p-e-w: The `Trial` object can be examined to check whether it's complete or not (`trial.state`).

p-e-w: But I wonder what Optuna does when resuming a study with incomplete trials. Does it redo the incomplete trial from scratch, or does it try to "partially resume" it somehow?

src/heretic/main.py
anrp: on checking, it actually appends a whiteout to the log (i.e. study still there but not readable through JournalStorage)

p-e-w: I think the file should be deleted to avoid leaving behind junk.

src/heretic/main.py

src/heretic/main.py

src/heretic/main.py
anrp: removed

src/heretic/main.py

src/heretic/main.py

src/heretic/main.py
anrp: development artifact, no way to get here the way its written now

src/heretic/main.py
anrp: yes

src/heretic/main.py

src/heretic/main.py
anrp: done (ty complaines about the type of settings.study_checkpoint_dir without str)

src/heretic/config.py
anrp: done

src/heretic/config.py
anrp: done

src/heretic/main.py

src/heretic/main.py
anrp: Added logic to detect unfinished test and erase it on load/start

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
     # Silence the warning about multivariate TPE being experimental.
     warnings.filterwarnings("ignore", category=ExperimentalWarning)
 
+    study_checkpoint_file = os.path.join(
+        str(settings.study_checkpoint_dir),
```
p-e-w: Now that you have changed the type, the cast to `str` should not be necessary anymore.

src/heretic/main.py
anrp: ... which doesn't work, looking, but it seems like study.trials is intended to be immutable once done (reasonable), so maybe its better to just make the score infinite? Maybe there's a way to mark the trial as ignored?

p-e-w: What does Optuna do with an incomplete trial when it resumes from the journal?

src/heretic/main.py
```diff
@@ -216,6 +219,79 @@ def run():
...
+
+        try:
+            old_settings = Settings.model_validate_json(previous_settings)
+            can_resume = True
+        except ValidationError as e:
+            raise Exception(
+                "Couldn't validate JSON loaded from study settings; schema change?"
+            ) from e
+
+        is_finished = storage.get_study_user_attrs(existing_study_id)["finished"]
```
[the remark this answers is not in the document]
p-e-w: 1 and 2 can be ignored. As for 3, I think we don't need that `try`/`except` block at all. This should be very rare, and we're getting an exception anyway with a stack trace that makes it clear where the problem is.

src/heretic/main.py
anrp: "Pruned" seems to have special meaning/logic within optuna, so I'm trying to get to a "FAIL" state without exiting completely. Seems like it has to be an exception percolated out of the trial function, so seeing how to do that cleanly.

spikymoth: I don't think we need to do anything to "fix" incomplete trials. We already ignore them when calculating the Pareto front, and TPESampler ignores incomplete trials as well. We could extend n_trials with the number of incomplete trials so that the number of complete trials eventually matches the desired n_trials.

src/heretic/main.py

src/heretic/main.py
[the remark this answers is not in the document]

src/heretic/main.py
anrp: This page https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/011_journal_storage.html suggests that it's related to shared storage. I haven't run this on Windows, I suppose it should work (uv is great), can try.

src/heretic/main.py
anrp: Pruned doesn't appear to mean incomplete, AFAICT, it seems to be grouped with TrialState.COMPLETE in (some) cases https://github.com/optuna/optuna/blob/e5d038f482b836396b22e72b6edde0f003ebdcea/optuna/samplers/_tpe/sampler.py#L531-L545 (a bit hard to follow)

spikymoth: Hmm, true, but we don't prune trials. We'd have to explicitly call `trial.should_prune()` in the objective function and then `raise optuna.TrialPruned()`, as in the example for [MedianPruner](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.pruners.MedianPruner.html) (the default pruner).

anrp: TrialPruned is raied in objective_wrapper, and then incompletely handled in optuna in our case (again..AFAICT?) https://github.com/optuna/optuna/blob/e5d038f482b836396b22e72b6edde0f003ebdcea/optuna/study/_optimize.py#L207-L210

spikymoth: That just handles the exception if it *is* raised by the objective function - but we never raise it. As for how TPESampler uses pruned trials, I think it just throws all of them into the "bad" distributions.

But since we don't prune, I don't think we have to worry about that. If Heretic ever starts pruning unpromising trials after some intermediate result, it should be easy enough to adjust the logic for it then.

src/heretic/main.py
```diff
@@ -442,33 +517,57 @@ def objective(trial: Trial) -> tuple[float, float]:
...
 
     def objective_wrapper(trial: Trial) -> tuple[float, float]:
         try:
+            trial.set_user_attr("finished", False)
             return objective(trial)
         except KeyboardInterrupt:
             # Stop the study gracefully on Ctrl+C.
             trial.study.stop()
+
             raise TrialPruned()
```
anrp, line 517: @spikymoth this line

spikymoth: Ah! Indeed, that would set `trial.state` to `TrialState.PRUNED`, then stop right after. But we ignore pruned trials here: https://github.com/anrp/heretic/blob/a68fc361911a7c5b459325c9fffb2d540840cdeb/src/heretic/main.py#L575

I think this is correct, as there is no value in a trial with a state other than `TrialState.COMPLETE`.

.gitignore
```diff
@@ -17,3 +17,6 @@ wheels/
 
 # Configuration files
 /config.toml
+
+# Logs
```
p-e-w:
```suggestion
# Study checkpoints
```

src/heretic/main.py
```diff
@@ -216,6 +219,77 @@ def run():
     # Silence the warning about multivariate TPE being experimental.
     warnings.filterwarnings("ignore", category=ExperimentalWarning)
 
+    study_checkpoint_file = os.path.join(
+        settings.study_checkpoint_dir,
+        "".join(
+            [
+                (c if (c.isalnum() or c in ["_", "-", "."]) else "--")
```
p-e-w: Dots (`.`) should probably not be allowed because of path traversal shenanigans. It doesn't seem to be exploitable with the current logic, but we don't want a small change later leading to a situation where running the program on a model name like `model/../../path` can somehow delete an arbitrary file.

src/heretic/main.py
```diff
@@ -216,6 +219,77 @@ def run():
...
+    can_resume = False
+    is_finished = False
+    try:
+        existing_study_id = storage.get_study_id_from_name("heretic")
+    except KeyError:
+        existing_study_id = None
+
+    if existing_study_id is not None:
+        # A study is in here. Check if it's finished.
+        previous_settings = storage.get_study_user_attrs(existing_study_id)["settings"]
```
p-e-w: I think it might be possible to simplify this and the code above. `storage.get_all_studies()` either returns an empty list (no stored study) or a list containing one study (since we only store one study per file). So we can directly get the `FrozenStudy` object as `storage.get_all_studies()[0]` without worrying about study names and study IDs at all.

src/heretic/main.py
```diff
@@ -216,6 +219,77 @@ def run():
...
+
+        old_settings = Settings.model_validate_json(previous_settings)
+        can_resume = True
+
+        is_finished = storage.get_study_user_attrs(existing_study_id)["finished"]
+
+    if can_resume:
```
p-e-w: This logic is convoluted. The variable `can_resume` is set, then the context of the above conditional is left, and then a new conditional is opened using the `can_resume` flag. It's much cleaner to just nest this conditional under the one above and get rid of those flag variables entirely.

src/heretic/main.py
```diff
@@ -216,6 +219,77 @@ def run():
...
+            )
+        )
+        choice = prompt_select("", choices)
+
+        if choice == "continue":
+            settings = old_settings
+        elif choice == "restart":
+            os.unlink(study_checkpoint_file)
+            backend = JournalFileBackend(study_checkpoint_file)
+            storage = JournalStorage(backend)
```
p-e-w, line 277: Why is this needed? Those assignments are identical to the ones above. Does merely instantiating those objects already open the file?

src/heretic/main.py
```diff
@@ -442,33 +517,57 @@ def objective(trial: Trial) -> tuple[float, float]:
         trial.set_user_attr("kl_divergence", kl_divergence)
         trial.set_user_attr("refusals", refusals)
 
+        trial.set_user_attr("finished", True)
+
         return score
 
     def objective_wrapper(trial: Trial) -> tuple[float, float]:
         try:
+            trial.set_user_attr("finished", False)
```
p-e-w: Why can't we just check if `trial.state == TrialState.COMPLETE`?

src/heretic/main.py
```diff
@@ -442,33 +517,57 @@ def objective(trial: Trial) -> tuple[float, float]:
...
+    if start_index > 0:
+        print("Resuming existing study.")
+
     try:
-        study.optimize(objective_wrapper, n_trials=settings.n_trials)
+        remaining_trials = settings.n_trials - start_index
+
+        if remaining_trials > 0:
+            study.optimize(objective_wrapper, n_trials=remaining_trials)
+        elif remaining_trials < 0:
```
p-e-w: How can this happen? When we resume a study, we load all settings (including `n_trials`) from storage, so how can `n_trials < start_index`?

src/heretic/main.py
anrp: Feels fuzzy if a cancelled-in-the-middle run is "complete", optuna probably sees it as yes (trial.stop() called) but it's not finishing according the settings

src/heretic/main.py
anrp: Yes, maybe creates and opens file on disk even if you del the object immediately

src/heretic/main.py

src/heretic/main.py
anrp: n_trials not updated on disk when using UI to run more - fixed

src/heretic/main.py

.gitignore
anrp: done

src/heretic/main.py
anrp: oh - wait - this is trial.state not study.state

Leftover from when I was trying to get TrialState.FAIL, removed

src/heretic/main.py
anrp: Added logic to increase n_trials by number of incomplete trials.

spikymoth: I think this check should now be
```python
if len(study.trials) == settings.n_trials + extra_trials:
    study.set_user_attr("finished", True)
```
as `extra_trials` is not added to `settings.n_trials`.

src/heretic/main.py
```diff
@@ -550,10 +639,14 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
 
                 settings.n_trials += n_more_trials
+                study.set_user_attr("settings", settings.model_dump_json())
+                study.set_user_attr("finished", False)
                 try:
                     study.optimize(objective_wrapper, n_trials=n_more_trials)
                 except KeyboardInterrupt:
                     pass
+                if len(study.trials) == settings.n_trials:
+                    study.set_user_attr("finished", True)
```
spikymoth:
```suggestion
                if len(study.trials) == settings.n_trials + extra_trials:
                    study.set_user_attr("finished", True)
```

src/heretic/main.py

src/heretic/main.py

src/heretic/main.py
anrp: True, but this feels a bit safer then just assuming.

p-e-w: This feels quite brittle, actually, because it introduces the ID and the name that have no semantics in the program itself.

Whenever something happens that doesn't match our assumptions, we should immediately throw an error and bail out.

src/heretic/main.py
```diff
@@ -216,6 +219,68 @@ def run():
...
+    try:
+        existing_study_id = storage.get_study_id_from_name("heretic")
+    except KeyError:
+        existing_study_id = None
+
+    if existing_study_id is not None:
+        # A study is in here. Check if it's finished.
+        previous_settings = storage.get_study_user_attrs(existing_study_id)["settings"]
+
+        old_settings = Settings.model_validate_json(previous_settings)
```
p-e-w: No need to do this until the user made their choice. Put this directly where it is used (line 275).

There are too many variables in the code right now. It's difficult to parse.

src/heretic/main.py
```diff
@@ -450,25 +516,48 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
             multivariate=True,
         ),
+        storage=storage,
         directions=[StudyDirection.MINIMIZE, StudyDirection.MINIMIZE],
+        load_if_exists=True,
+    )
+
+    # Count number of incomplete/failed trials to run due to cancellations.
+    extra_trials = sum(
+        [(0 if t.state == TrialState.COMPLETE else 1) for t in study.trials]
```
p-e-w: The `extra_trials` logic is really difficult to understand. Leaving it out (as has happened already once per the comment above) creates bugs.

The proper way is to *once* calculate how many trials were already completed (the complement of `extra_trials`), then subtract that number from `n_trials`, and then all existing logic can remain as it was.

src/heretic/main.py

src/heretic/main.py
anrp: Moved to FrozenStudy inspection

src/heretic/main.py
anrp: reworked, per other comments

src/heretic/main.py
anrp: Tested; no issues resuming, or deleting and restarting.

src/heretic/config.py
```diff
@@ -271,9 +267,15 @@ def settings_customise_sources(
         file_secret_settings: PydanticBaseSettingsSource,
     ) -> tuple[PydanticBaseSettingsSource, ...]:
         return (
-            init_settings,
-            env_settings,
+            init_settings,  # Used during resume - should override *all* other sources.
```
anrp, line 270: fyi this was necessary because otherwise the cli arguments unconditionally overrode settings (I think when you're using cli_parse_args, it's injected as the first in this list implicitly when it's not already there)

p-e-w: I don't get it. CLI arguments overriding settings is what we want, no?

p-e-w: Ah, do you mean that the `Settings` constructor is invoked to load the stored settings and pulls from CLI again?

So why does it work now? What is the key parameter that was changed?

p-e-w: Note that this means that if the stored `Settings` object doesn't contain a setting that was added in between, the setting can be changed arbitrarily via the CLI.

src/heretic/main.py
```diff
@@ -450,25 +514,46 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
         ),
+        storage=storage,
         directions=[StudyDirection.MINIMIZE, StudyDirection.MINIMIZE],
+        load_if_exists=True,
     )
 
+    study.set_user_attr("settings", settings.model_dump_json())
+    study.set_user_attr("finished", False)
+
+    def get_trials_completed():
```
p-e-w: This should be named `count_completed_trials`, because it returns an integer, not the trials themselves. It also needs a return type annotation.

src/heretic/config.py
```diff
@@ -168,6 +169,11 @@ class Settings(BaseSettings):
         description="Number of trials that use random sampling for the purpose of exploration.",
     )
 
+    study_checkpoint_dir: str = Field(
+        default="checkpoints",
+        description="Directory to save and load study progress to/from",
```
p-e-w: Missing colon at the end.

src/heretic/main.py
```diff
@@ -450,25 +514,46 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
         except KeyboardInterrupt:
             # Stop the study gracefully on Ctrl+C.
             trial.study.stop()
+
```
p-e-w: Remove blank line.

src/heretic/main.py
```diff
@@ -450,25 +514,46 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
...
             raise TrialPruned()
 
     study = optuna.create_study(
+        study_name="heretic",
```
p-e-w: This shouldn't be necessary anymore.

src/heretic/config.py
anrp: It didn't actually work before after I removed the settings whitelist - it just happened that the testing I did was always with the same parameters after I stripped out the setting-in-name capability.

Re 2nd point yes - true - question becomes how many layers of safety do you want? To trigger that you'd have to know, have interrupted a run, changed code to get new settings, and then tried to resume.

src/heretic/main.py

src/heretic/main.py

src/heretic/config.py
anrp: done

src/heretic/main.py

src/heretic/config.py
p-e-w: It's fine I think. Better to have tiny gaps like that than to pile on complexity endlessly, which only leads to more subtle bugs later.

p-e-w:
Yup, that looks good now. Merged!

PR #108, opened 2026-01-15
fix: Allow abliterating VL models

anrp:
Per https://huggingface.co/docs/transformers/en/model_doc/auto#auto-classes, it indicates that "There is one class of AutoModel for each task." Since we're not using AutoModel (allowing it to pick the task automatically), expose a flag to allow for a different "task", which allows (at least) Qwen3-VL to be abliterated.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -4,7 +4,7 @@
 import math
 from contextlib import suppress
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, Type, cast
 
 import bitsandbytes as bnb
 import torch
@@ -15,9 +15,11 @@ from torch import FloatTensor, LongTensor, Tensor
 from torch.nn import Module, ModuleList
 from transformers import (
     AutoModelForCausalLM,
+    AutoModelForImageTextToText,
     AutoTokenizer,
     BatchEncoding,
     BitsAndBytesConfig,
+    PretrainedConfig,
     PreTrainedModel,
     PreTrainedTokenizerBase,
     TextStreamer,
@@ -28,6 +30,17 @@ from transformers.generation import (
 
 from .config import QuantizationMethod, Settings
 from .utils import Prompt, batchify, empty_cache, print
+
+
+def get_model_class(
+    model: str,
+) -> Type[AutoModelForImageTextToText] | Type[AutoModelForCausalLM]:
+    configs = PretrainedConfig.get_config_dict(model)
+
+    if any(["vision_config" in x for x in configs]):
+        return AutoModelForImageTextToText
+    else:
+        return AutoModelForCausalLM
 
 
 @dataclass
@@ -87,7 +100,7 @@ class Model:
                 if quantization_config is not None:
                     extra_kwargs["quantization_config"] = quantization_config
 
-                self.model = AutoModelForCausalLM.from_pretrained(
+                self.model = get_model_class(settings.model).from_pretrained(
                     settings.model,
                     dtype=dtype,
                     device_map=settings.device_map,
@@ -159,6 +172,9 @@ class Model:
             lora_alpha=1,
             lora_dropout=0,
             bias="none",
+            # Even if we're using AutoModelForImageTextToText, this is still correct, as it is (post-vision)
+            # the same kind of model.
+            # https://github.com/huggingface/peft/blob/622c2821cb0d7897bee53aad7914d42b5fecbf61/src/peft/auto.py#L45
             task_type="CAUSAL_LM",
         )
 
@@ -212,7 +228,7 @@ class Model:
 
             # Load base model in full precision on CPU to avoid VRAM issues
             print("* Loading base model on CPU (this may take a while)...")
-            base_model = AutoModelForCausalLM.from_pretrained(
+            base_model = get_model_class(self.settings.model).from_pretrained(
                 self.settings.model,
                 torch_dtype=self.model.dtype,
                 device_map="cpu",
@@ -282,7 +298,7 @@ class Model:
         if quantization_config is not None:
             extra_kwargs["quantization_config"] = quantization_config
 
-        self.model = AutoModelForCausalLM.from_pretrained(
+        self.model = get_model_class(self.settings.model).from_pretrained(
             self.settings.model,
             dtype=dtype,
             device_map=self.settings.device_map,
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -31,12 +31,11 @@ from optuna.trial import TrialState
 from pydantic import ValidationError
 from questionary import Choice
 from rich.traceback import install
-from transformers import AutoModelForCausalLM
 
 from .analyzer import Analyzer
 from .config import QuantizationMethod, Settings
 from .evaluator import Evaluator
-from .model import AbliterationParameters, Model
+from .model import AbliterationParameters, Model, get_model_class
 from .utils import (
     empty_cache,
     format_duration,
@@ -80,7 +79,7 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
             # These are expected and harmless since we're only inspecting model structure, not running inference.
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
-                meta_model = AutoModelForCausalLM.from_pretrained(
+                meta_model = get_model_class(settings.model).from_pretrained(
                     settings.model,
                     device_map="meta",
                     torch_dtype=torch.bfloat16,


src/heretic/model.py
```diff
@@ -30,6 +31,12 @@
 from .utils import Prompt, batchify, empty_cache, print
 
 
+def get_model_class(settings: Settings):
```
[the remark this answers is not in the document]
anrp: Comment is hallucination; per https://huggingface.co/docs/transformers/en/model_doc/auto#transformers.AutoModelForImageTextToText Qwen2 VL, Qwen2.5VL and Qwen3 VL are all not Seq2Seq and instead CAUSAL_LM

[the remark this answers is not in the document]
p-e-w: Not sure what you mean? Requiring a return type annotation is not a hallucination.

p-e-w:
Having a configuration flag for this is not the correct solution. This should be determined automatically.

anrp:
Do you know if there's a reason that it wasn't just using AutoModel to begin with? I think I tried that before but got some vague error, but maybe I just did it wrong...

p-e-w:
I never tried `AutoModel`. Does it work with VL models?

anrp:
OK, per https://huggingface.co/docs/transformers/model_doc/auto#transformers.AutoModel
AutoModel.from_pretrained("openai/gpt-oss-20b") returns a GptOssModel which does not have a "generate" method
AutoModel.from_pretrained("Qwen/Qwen3-VL-2B-Instruct") returns a Qwen3VLModel, same

So we need the "{model}ForConditionalGeneration" type as a return but per that page only the specific AutoModelFor{CausalLM,ImageTextToText,(others)} classes can build them. I think you have to pick the superclass?

anrp:
Futher evidence that you have to choose the task https://discuss.huggingface.co/t/difference-between-automodel-and-automodelforlm/5967 https://insujang.github.io/2023-04-19/using-huggingface-transformers/

anrp:
OK - figured out how to get the class from the config. No config needed now.

src/heretic/model.py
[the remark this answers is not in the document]
anrp: That comment was targeted at the second automatic comment, but its now commented on.

src/heretic/model.py
```diff
@@ -30,6 +34,19 @@
...
+    settings: Settings,
+) -> Type[AutoModelForImageTextToText] | Type[AutoModelForCausalLM]:
+    cfg_type = type(AutoConfig.from_pretrained(settings.model))
+    if MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING.get(cfg_type, default=None) is not None:
+        return AutoModelForImageTextToText
+    if MODEL_FOR_CAUSAL_LM_MAPPING.get(cfg_type, default=None) is not None:
+        return AutoModelForCausalLM
+    raise Exception(
+        f"Do not know how to get AutoModel* for {settings.model} (cfg type: {cfg_type})"
+    )
```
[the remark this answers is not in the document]
anrp: done

Vinay-Umrethe:
> I never tried `AutoModel`. Does it work with VL models?



That's the most generic class transformers have, as last fallback users can have after no class like `AutoModelForCausalLM, AutoModelForImageTextToText, AutoModelForTextToWaveform` etc don't work

Having that + `AutoProcessor` would truly make heretic support multi-modal

p-e-w:
So is this working? Which VL models have you tested it with?

src/heretic/model.py
```diff
@@ -30,6 +34,19 @@
 from .utils import Prompt, batchify, empty_cache, print
 
 
+def get_model_class(
+    settings: Settings,
```
p-e-w: We only use the `model` field from the settings here, so I think it's cleaner to just pass the model path directly instead of the entire settings object.

src/heretic/model.py
```diff
@@ -30,6 +34,19 @@
...
+) -> Type[AutoModelForImageTextToText] | Type[AutoModelForCausalLM]:
+    config_type = type(AutoConfig.from_pretrained(settings.model))
+    if config_type in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING:
+        return AutoModelForImageTextToText
+    if config_type in MODEL_FOR_CAUSAL_LM_MAPPING:
```
p-e-w: Are those the same dicts that are used by `AutoModelFor*` internally? We don't want to lose the ability to load models that were previously supported.

anrp: https://github.com/huggingface/transformers/blob/57278c904c5158999d31a0db8bfcd63360c37b48/src/transformers/models/auto/modeling_auto.py#L1979
https://github.com/huggingface/transformers/blob/57278c904c5158999d31a0db8bfcd63360c37b48/src/transformers/models/auto/modeling_auto.py#L2183

Yes - exactly the same.

anrp:
I've tested Qwen3-VL, I will test a few more (InternVL, Qwen2.5-VL), but given that it's the same data structure that transformers uses internally in AutoModel I don't expect trouble.

src/heretic/model.py
anrp: Done.

p-e-w:
Let me know when you're satisfied that this works, then I'll merge.

anrp:
Had to rework this as non-transformer-known-models (InternVL3, InternVL3.5) do not register for auto classes, but instead provide data in the config.json under auto_map (example: https://huggingface.co/OpenGVLab/InternVL3-1B/blob/main/config.json#L7). Also pulled param back to settings, as getting asked repeatedly about remote code was annoying.

Vinay-Umrethe:
> Had to rework this as non-transformer-known-models (InternVL3, InternVL3.5) do not register for auto classes, but instead provide data in the config.json under auto_map (example: https://huggingface.co/OpenGVLab/InternVL3-1B/blob/main/config.json#L7). Also pulled param back to settings, as getting asked repeatedly about remote code was annoying.

Transformers simply didn't merged those earlier versions at first but for InternVL 3.5 They already did it after sometime

```json
"auto_map": {
"AutoConfig": "configuration_internvl_chat.InternVLChatConfig",
"AutoModel": "modeling_internvl_chat.InternVLChatModel",
"AutoModelForCausalLM": "modeling_internvl_chat.InternVLChatModel"
},
```


If it's there it would work with both the Auto Classes, just needs trust_remote_code=True which heretic already has

src/heretic/model.py
```diff
@@ -30,6 +34,33 @@
 from .utils import Prompt, batchify, empty_cache, print
 
 
+def get_model_class(
+    settings: Settings,
+) -> Type[AutoModelForImageTextToText] | Type[AutoModelForCausalLM]:
+    config = AutoConfig.from_pretrained(
+        settings.model, trust_remote_code=settings.trust_remote_code
```
p-e-w: This isn't how we handle trusting remote code; see the other instances for details.

src/heretic/model.py
```diff
@@ -30,6 +34,33 @@
...
+
+    # If the model self-registered (register_for_auto_class), it should be in this mapping.
+    if config_type in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING:
+        return AutoModelForImageTextToText
+    if config_type in MODEL_FOR_CAUSAL_LM_MAPPING:
+        return AutoModelForCausalLM
+
+    # If the model is not self registered, it may provide task class details, documented at
+    # https://huggingface.co/docs/transformers/en/custom_models#upload
+    maybe_auto_map = getattr(config, "auto_map", None)
```
p-e-w: Can't we get rid of all that complexity by checking for a `vision_config` key in the configuration? It seems that most or all VL models use that. Basically, we just load the `config.json` (which can be done without `trust_remote_code`), then check for the `vision_config` key. If it is present, we return `AutoModelForImageTextToText`, else we return `AutoModelForCausalLM`.

anrp: done

src/heretic/model.py
anrp: no longer used

p-e-w:
:+1: This looks great now! This has got to be the cleanest way to solve the problem.

As far as I'm concerned, this is ready to merge. Are you satisfied that it works correctly with all/most VL models?

anrp:
I'll test with some models I know can come in under both superclasses (gemma3) as well as some others.

anrp:
Tested working openai/gpt-oss-20b, google/gemma-3-4b-it, Qwen/Qwen3-0.6B, Qwen/Qwen3-VL-2B-Instruct, Qwen/Qwen2.5-VL-3B-Instruct, OpenGVLab/InternVL3_5-1B-hf

anrp:
A couple are weird; allenai/Molmo2-8B uses "vit_config" and not "vision_config", mistralai/Ministral-3-3B-Instruct-2512 complains about chat user-assistant ordering (not related to this change)

p-e-w:
Please give me a clear signal when this is ready to merge.

anrp:
This is a strict improvement, and the failures are unrelated (Molmo doesn't support system prompt, ministral 2512 needs transformers 5). :shipit:

p-e-w:
Ok, merged. This is a great improvement!

PR #116, opened 2026-01-23
fix: Use file instead of symlink lock (for windows)

anrp:


src/heretic/main.py
```diff
@@ -301,8 +302,10 @@ def run():
                 existing_study.user_attrs["settings"]
             )
         elif choice == "restart":
+            del backend
+            del storage
```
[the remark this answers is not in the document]
anrp: Wrong; lock file only used when writing per https://optuna.readthedocs.io/en/v4.2.1/_modules/optuna/storages/journal/_file.html#JournalFileOpenLock

red40maxxer:
sorry, didn't realize you had already submitted a PR for this fix. Just tested it out and it works for me

src/heretic/main.py
```diff
@@ -301,8 +302,10 @@ def run():
                 existing_study.user_attrs["settings"]
             )
         elif choice == "restart":
+            del storage
+            del backend
```
p-e-w: This is not a reliable solution.

Contrary to popular belief, `del x` does **not** call `x.__del__()`. By the Python language specification, the finalizer is called by the garbage collector *at some point in the future* (in fact, it is not guaranteed to *ever* be called).

If this is working for you (i.e., if you are seeing cleanup before the `unlink` call), you are relying on an implementation detail of current CPython versions. Otherwise, those `del` statements are superfluous, as the reassignments below also drop the previous reference.

src/heretic/main.py
```diff
@@ -301,8 +302,10 @@ def run():
...
             os.unlink(study_checkpoint_file)
-            backend = JournalFileBackend(study_checkpoint_file)
+            backend = JournalFileBackend(study_checkpoint_file, lock_obj=lock_obj)
```
p-e-w, line 306: Why does the lock not need to be re-initialized, despite also being tied to the checkpoint file?

anrp: Because it's a per-write lock i.e. the lock is taken to append or read a line and then removed; it's not an open lock.

anrp: Actually it's *only* to write per https://optuna.readthedocs.io/en/stable/_modules/optuna/storages/journal/_file.html#JournalFileStorage

src/heretic/main.py
anrp: They're actually completely unnecessary as these objects don't even have `__del__` methods; want me to just remove it?

p-e-w: Yes.

p-e-w:
Merged, thanks!

Although this change is technically only necessary on Windows, I like using it on all platforms because it makes environments easier to compare.

PR #119, opened 2026-01-25
fix: specify study name

anrp:
If we don't, optuna will generate a UUID for a name, which will never be found when loading as it is a "different" study. https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.create_study.html#optuna.study.create_study


p-e-w:
Ah yes, because of `load_if_exists`. Merged!

PR #120, opened 2026-01-25
Session restore

dpankros:
This PR adds `heretic-restore <checkpoint.json>` to load a previously-created checkpoint.json file and provide the opportunity to save off a lora or merged model without re-running trials.  This is helpful if there are crashes, for example.


p-e-w:
Isn't this the same thing as #106, which was merged just a few days ago?

p-e-w:
Closing as we already have this feature (#106).

PR #138, opened 2026-02-08
feat: make parameter ranges configurable

spikymoth:
Alright, this will probably still need some fine-tuning, but I think it should be relatively simple to extend https://github.com/p-e-w/heretic/pull/53 for this or vice versa.

The idea is as follows:
* Parameter ranges are specified by `CategoricalParameterSpecification` and `FloatParameterSpecification`.
* A particular ablation method defines a set of default ranges that correspond to the parameters used by the objective function.
* For each parameter, we look up the most specific parameter specification - either the default, or an override from the config. More specific overrides are preferred: The longer the matching suffix, the higher the priority.
* Once a parameter specification is obtained, we call `suggest_float` or `suggest_categorical` on that specification.

The logic in `config.py` doesn't make any assumptions about what parameters are used, so it should be reusable for any set. In the current implementation, we call `settings.set_parameter_range_defaults()` from `main.py`, but this could be abstracted into a `Modifier` or similar in the context of https://github.com/p-e-w/heretic/pull/53 (as far as I can see, that PR currently does not touch the objective function).

The parameter ranges themselves mostly match the current defaults, with 1 caveat:
* To make the ranges model-agnostic (required for the defaults), I used ranges between 0 and 1 for the positional parameters, multiplying with `last_layer_index` afterward.
* For `min_weight_distance`, this meant I had to set the minimum to 0 which is a slight change. I think this is harmless - a distance of 0 just means that set of parameters will do nothing. Technically values between 0 and 1 post-multiplication can also still be effective since the nearest layer should have a distance < 1. Either way it shouldn't impact the sampler much in practice.

A few more notes:
* The type syntax in Python 3.10 (the minimum version targeted by Heretic) is somewhat limited, requiring use of `Type`, `TypeVar`, `TypeAlias` and `TypeAliasType`. Python 3.12 adds a cleaner syntax for this.
* ty doesn't seem to support narrowing generics with unions (see e.g. https://github.com/astral-sh/ty/issues/2726), so I had to add a suppression.
* To avoid having to repeat the module and parameter name when calling `suggest_float` or `suggest_categorical`, I set private attribute `_name` during `_param_spec()`. It works fine, but it does mean that the returned object is specific to that loop iteration. Alternatives would be to create a copy with specific name, or repeat the name twice, but the former felt wasteful and the latter felt like a potential footgun.
* I added all the current parameter ranges to `config.default.toml`. That's not really required, and it makes `config.default.toml` specific to the default ablation method, but I think it's very helpful for discoverability.


spikymoth:
As expected, Gemini didn't really like my usage of `_name` and `set_name()` ;) Let me know what you think. I pushed a small change to turn the asserts into exceptions.

src/heretic/config.py
```diff
@@ -61,6 +63,63 @@ class DatasetSpecification(BaseModel):
...
+    def __len__(self) -> int:
+        return len(self.root)
+
+    def __contains__(self, item: str) -> bool:
+        return item in self.root
+
+    def set_name(self, name: str):
+        self._name = name
+
+    def suggest_categorical(self, trial: Trial) -> str:
```
p-e-w: The class is already named `CategoricalParamSpecification`, so the `_categorical` part is redundant here. I think it would be cleaner if all `*ParamSpecification` classes implemented an interface `ParamSpecification` that simply has a `suggest` method. This will allow abstracting over different implementations without having to distinguish them in code.

src/heretic/config.py
```diff
@@ -61,6 +63,63 @@ class DatasetSpecification(BaseModel):
     )
 
 
+class CategoricalParamSpecification(RootModel[list[str]]):
```
p-e-w: I think it's simpler (and cleaner) to have a full class here. This not only gets rid of that complicated mechanism, it also allows us to later add additional parameters (e.g. probability weights for categories).

src/heretic/config.py
```diff
@@ -311,6 +370,13 @@ class Settings(BaseSettings):
         description="Dataset of prompts that tend to result in refusals (used for evaluating model performance).",
     )
 
+    parameter_ranges: dict[str, ParamSpecificationRecursive] = Field(
+        default={},
```
p-e-w: Why can the defaults not be specified here?

src/heretic/model.py
```diff
@@ -401,6 +401,8 @@ def abliterate(
         # the embedding matrix, but it's unclear if that has any benefits.
         for layer_index in range(len(self.get_layers())):
             for component, modules in self.get_layer_modules(layer_index).items():
+                if component not in parameters:
```
p-e-w: When does this happen?

p-e-w (commented):
Just a quick first look.

src/heretic/config.py
spikymoth: Hmm, but I think you need something inheriting from `RootModel[list[T]]` in order to wrap a list; I'm pretty sure `BaseModel` expects a dict of fields.

If it were a `list[CategoricalParamChoice]` where each `CategoricalParamChoice` describes a dict like `{ choice: str, probability: float }` that could probably work, though then you'd need some functions to access it.

However, I don't think `optuna.trial.Trial.suggest_categorical` actually supports weighted choices, so you'd need to do something like `suggest_float` and select a choice based on the outcome.

src/heretic/config.py
spikymoth: Hmm, simply `suggest()` could work but I'm not sure it's a good idea to make the classes completely generic - for example I already used `"global" in choices` (using `"global" in choices.root` would simplify the class) and `if max_weight_range.high == 0: continue`, neither of which work across both types.

src/heretic/config.py
spikymoth: I had them there originally, but that would make `config.py` specific to the default ablation method and would make it less clear how to swap in a different set of defaults with a plugin. I also kind of like having the defaults close to the ablation method itself.

src/heretic/model.py
spikymoth: If `max_weight_range.high == 0` and we skip that loop iteration altogether (to skip ablating a component), that component never gets added to the dict of parameters. There could also be other triggers, e.g. if `direction_scope` were made component-specific (as I've done in some of my experiments) and you passed an empty list.

src/heretic/config.py
spikymoth: > so you'd need to do something like `suggest_float` and select a choice based on the outcome.

Or simulate probabilities by passing a list with multiples of the same choice. But you can already do that manually.

spikymoth:
Perhaps another way to go here would be to initialize the settings as simple dicts and lists, then instantiate an immutable `*ParamSpecification` with the requested `name` on the fly. They wouldn't even need to inherit from `BaseModel` or `RootModel` (though they would be accessed essentially the same way).

The downside is that you would lose pydantic checks for the presence of `low`, `high` and the optional `log` (they could be moved into the constructor, but wouldn't trigger until first use).

Edit: Of course, we could also do both - 2 classes for setting defaults and reading the config, 2 classes for actually using them. 2 more classes, but probably a simpler implementation overall.

src/heretic/config.py
p-e-w: Sure, I guess this is Ok.

src/heretic/config.py
p-e-w: This will all change radically anyway with the upcoming modifier plugins, where defaults will be per-plugin. Please keep things as simple and clean as possible for now. The defaults belong where there are for every other setting. Having to call an extra method in order for things to work properly is brittle; we use Pydantic Settings precisely so we don't have to take care of such things ourselves.

src/heretic/model.py
p-e-w: > If max_weight_range.high == 0 and we skip that loop iteration altogether (to skip ablating a component)

I don't like this mechanism. Numerical ranges should encode quantities, not semantics. Giving specific values magical properties that change the behavior of the system makes things hard to understand.

Not to mention that it seems perfectly reasonable to have, for example,

```toml
max_weight = { low = -0.5, high = 0.0 }
```

src/heretic/config.py
p-e-w: Those special cases have to be eliminated. The external code should not need to introspect the structure of the parameter specification. It should receive the suggestion and act on it.

KaraKaraWitch:
Commenting here since I also looked into this in another [PR](https://github.com/p-e-w/heretic/pull/170). While it's less flexible as it doesn't use generic parameter specs, it strictly typed and, in my opinion, much more readable from a maintainer's standpoint. 

I have to agree that using `0` to skip a weight is the wrong approach. It adds complexity by conflating quantity (the weight value) with behavior (control flow/skipping).

Plus, it's an additional mental step (and potentially a docstring to read!) to remember that "0 means to skip a layer" rather than "0 means zero weight."

As for the toml format, I would prefer explicit tables over inline dictionaries for readability:

```toml
[parameter_ranges]
direction_scope = [
    "global",
    "per layer",
]

[parameter_ranges.direction_index]
low=0.4
high=0.9
```

...instead of the inline `direction_index = { low = 0.4, high = 0.9 }`

Lastly, regarding this block in main.py:

```py



    settings.set_parameter_range_defaults(
        {
            "direction_scope": CategoricalParamSpecification(["global", "per layer"]),
            # Discrimination between "harmful" and "harmless" inputs is usually strongest
            # in layers slightly past the midpoint of the layer stack. See the original
            # abliteration paper (https://arxiv.org/abs/2406.11717) for a deeper analysis.
            "direction_index": FloatParamSpecification(low=0.4, high=0.9),
            # The parameter ranges are based on experiments with various models
            # and much wider ranges. They are not set in stone and might have to be
            # adjusted for future models.
            "max_weight": FloatParamSpecification(low=0.8, high=1.5, log=False),
            "max_weight_position": FloatParamSpecification(low=0.6, high=1.0),
            # For sampling purposes, min_weight is expressed as a fraction of max_weight,
            # because multivariate TPE doesn't support variable-range parameters.
            "min_weight": FloatParamSpecification(low=0.0, high=1.0),
            "min_weight_distance": FloatParamSpecification(low=0.0, high=0.6),
        }
    )
```

We really should put the configuration in the toml rather than in a python.

Addn. We should also generalize `parameter_ranges` to be `constraints` since future abliteration methods or pluguns may not use `parameter_ranges` and the `parameter_ranges` name would feel misleading.

spikymoth:
> I have to agree that using `0` to skip a weight is the wrong approach. It adds complexity by conflating quantity (the weight value) with behavior (control flow/skipping).
> 
> Plus, it's an additional mental step (and potentially a docstring to read!) to remember that "0 means to skip a layer" rather than "0 means zero weight."

Point taken, especially with the possibility of a negative minimum weight (for the opposite effect). I still think it would be nice to have an explicit way to disable optimization for a particular module, but maybe there should be a dedicated parameter for it (e.g. `disabled_modules` or some such that just takes a list of module names).

> As for the toml format, I would prefer explicit tables over inline dictionaries for readability:

I think this actually Just Works™, although I haven't tested it. So it would just be a matter of changing the defaults.

> Lastly, regarding this block in main.py:
> 
> We really should put the configuration in the toml rather than in a python.

I don't see how we could do that; the defaults have to be somewhere and the user's config file might be empty. I did add the defaults to `config.default.toml` as well (as you saw), but it can't be the only source of truth unless we start explicitly reading `config.default.toml` (and then plugins would still need some way to augment/replace it).

> Addn. We should also generalize `parameter_ranges` to be `constraints` since future abliteration methods or plugins may not use `parameter_ranges` and the `parameter_ranges` name would feel misleading.

I'm open to other names, though I'm not sure "constraints" is the best name either given that `TPESampler` takes an explicit `constraints_func` which Heretic currently doesn't use (and has a different purpose than parameter selection).

spikymoth:
I'll have a think about the best way to incorporate the feedback so far and aim for something more straightforward.

p-e-w:
> > As for the toml format, I would prefer explicit tables over inline dictionaries for readability:
>
> I think this actually Just Works™

Yes. The format you use in this PR is called an "inline table", and per the TOML specification is equivalent to the full table syntax. The configuration file can use either format and will be deserialized to the same Python object.

I actually prefer the inline syntax from this PR for this particular use case, because the number of fields is very small.

p-e-w:
> We should also generalize `parameter_ranges` to be `constraints` since future abliteration methods or pluguns may not use `parameter_ranges` and the `parameter_ranges` name would feel misleading.

The name `parameter_ranges` is indeed problematic, because categorical parameters are not "ranges", but Heretic will not generalize arbitrarily and Optuna will remain our backbone. Therefore, `constraints` is too generic.

I suggest we simply use the name `parameters`.

spikymoth:
Sorry about the silence here, I've been a bit drained from work. Planning to do some work on this and my other PRs starting Saturday.

spikymoth:
Okay, refactored significantly.

* Float param specifications are now represented by `FloatParamSpec`, which only defines the actual fields.
* Categorical param lists just use `list` without any frills
* New classes `ParamCategorical` and `ParamFloat` now handle the actual parameter suggestion. They inherit from base class `Parameter`.

After using the previous implementation myself, I realized that there was a mismatch between the parameter ranges that get displayed and the parameter ranges that are set in the config file: The default and configurable ranges are between 0 and 1, but the values that are actually used are multiplied by the last layer index.

That makes them annoying to adjust between runs: After figuring out the range you want to use (e.g. for the direction index), you then first have to divide it by the last layer index in order to actually apply it in the config file.

To solve this problem, I moved the multiplication into `abliterate`, so that the stored and displayed values can match the settings. This does mean that we now have the opposite problem: To map displayed positions and distances, you first have to multiply them with the last layer index. But I personally think that making the parameter ranges configurable is more important.

src/heretic/config.py
```diff
@@ -313,6 +383,26 @@ class Settings(BaseSettings):
         description="Dataset of prompts that tend to result in refusals (used for evaluating model performance).",
     )
 
+    parameters: dict[str, ParamSpecRecursive] = Field(
```
p-e-w:
```suggestion
    parameters: dict[str, ParamFloat | dict[str, ParamFloat]] = Field(
```

p-e-w:
```toml
max_weight = { low = 0.8, high = 1.5, log = false }
```

or

```toml
max_weight = {
  attn = { low = 0.8, high = 1.5, log = false },
  mlp = { low = 0.4, high = 1.0 },
}
```

spikymoth:
Okay, I think this is ready for another look. I made it as statically typed as I could while also implementing the features we discussed. Unfortunately it's quite a lot more code.

One annoyance I ran into that I can't really fix: TOML v1.0 doesn't allow inline tables (like `{ low = 0.0, high = 1.0 }`) to span multiple lines. TOML v1.1 *does* allow it, but that spec is quite recent and Python's `tomllib` doesn't support it yet. `tomli` does support it, but Pydantic uses `tomllib` for Python 3.11+ so you'd need a hacky override like `sys.modules["tomllib"] = tomli`.

I also couldn't really do static typing for `attn.o_proj` and `mlp.down_proj` because class members with dotted keys aren't allowed. I settled for naming the fields `attn_o_proj` and `mlp_down_proj`, while still requiring `"attn.o_proj"` and `"mlp.down_proj"` on the config file level. That might lead to `attn_o_proj` and `mlp_down_proj` being displayed in some configuration error messages, but should otherwise work transparently.

I had to extend the error reporting in `main.py` slightly as it was only using the first element of `error['loc']`, which ended up hiding the true location for nested settings. It now also handles array elements like `parameters.direction_index[0]`.

To get errors to show up properly I also had to move some additional validation into the `BeforeValidator` (instead of just handling constant values like `"global"` and `1.0` there). Allowing the union types to flow into Pydantic's normal validation made for bad error locations like `('parameters', 'max_weight', 'function-after[_validate_float(), FloatParamSpec]', 'low')` and `('parameters', 'max_weight', 'FloatParamModuleSpec', 'attn_o_proj')`, which appears to be a known limitation.

config.noslop.toml
```diff
@@ -161,3 +161,45 @@ dataset = "llm-aes/writing-prompts"
 split = "train[1000:1100]"
 column = "prompt"
 prefix = "Write a short story based on the writing prompt below.\n\nWriting prompt:"
+
+# The parameters used to choose suggest settings for abliteration. With the
```
p-e-w: Remove the changes from `config.noslop.toml` please; we only keep the noslop-relevant settings here and use the defaults for everything else.

src/heretic/main.py
```diff
@@ -477,71 +480,53 @@ def objective(trial: Trial) -> tuple[float, float]:
...
+        # scope is included in the choices, even though we only need it for the
+        # "global" direction scope itself. The reason is that multivariate TPE
+        # doesn't work with conditional or variable-range parameters.
+        if "global" in direction_scope_param.root:
+            direction_fraction = params.direction_fraction.suggest(trial)
+        else:
+            direction_fraction = None
 
-        parameters = {}
+        suggested_params: dict[str, AbliterationParameters] = {}
```
p-e-w: Why this rename?

src/heretic/config.py
```diff
@@ -61,6 +82,296 @@ class DatasetSpecification(BaseModel):
...
+                param[component] = {"low": value, "high": value}
+            else:
+                param[component] = value
+        if model_component_found:
+            return UnitParamModuleSpec.model_validate(param)
+        elif "low" in param or "high" in param:
+            return UnitParamSpec.model_validate(param)
+        else:
+            return UnitParamModuleSpec.model_validate(param)
+    raise ValueError(f"Cannot determine param type for: {param!r}")
```
p-e-w: We have to rethink this code. There is massive duplication here, and pretty much all of this I would expect Pydantic to handle automatically.

src/heretic/config.py
```diff
@@ -61,6 +82,296 @@ class DatasetSpecification(BaseModel):
...
+    "CategoricalParamTypeComplex",
+    Annotated[
+        CategoricalParamSpec[Scalar] | CategoricalParamModuleSpec[Scalar],
+        BeforeValidator(parse_categorical_param),
+    ],
+    type_params=(Scalar,),
+)
+
+# Note: Although the above type aliases correctly type categorical parameters,
+# ty is unable to see through them, so we use a specialized type here.
```
p-e-w: We're not changing valid code to satisfy ty. This is a shortcoming of ty, and the "solution" is to disable ty on the relevant lines until it is fixed.

src/heretic/config.py
```diff
@@ -61,6 +82,296 @@ class DatasetSpecification(BaseModel):
...
+    max_weight: FloatParamTypeComplex = Field(
+        description=(
+            "The maximum weight with which to apply the abliteration. Set log = true to "
+            "sample from the log space, which will select lower values more frequently. "
+            "Note that low must be greater than 0 when log = true."
+        ),
+    )
+
+    max_weight_position_fraction: UnitParamTypeComplex = Field(
+        description="The position (layer) at which the maximum weight should be applied.",
```
p-e-w: Misleading, because it's not a layer but a *fractional* layer index. Same for `direction_fraction` above.

src/heretic/main.py
```diff
@@ -163,7 +163,10 @@ def run():
         print(f"[red]Configuration contains [bold]{error.error_count()}[/] errors:[/]")
 
         for error in error.errors():
-            print(f"[bold]{error['loc'][0]}[/]: [yellow]{error['msg']}[/]")
+            full_loc = str(error["loc"][0])
+            for loc in error["loc"][1:]:
+                full_loc += f".{loc}" if isinstance(loc, str) else f"[{loc}]"
+            print(f"[bold]{full_loc}[/]: [yellow]{error['msg']}[/]")
```
p-e-w: What problem does this fix?

p-e-w (commented):
We have to talk about this in some more detail. I would have expected 80% of the logic in this PR to be handled automatically by Pydantic.

src/heretic/main.py
p-e-w: Nevermind, I understood your comment now.

spikymoth:
Just organizing my thoughts a bit based on our conversation:
1. Decouple any conversion and merging logic, make Pydantic purely responsible for type checking (and serialization/deserialization)
    1. `@model_validator(mode="after")` can be moved to `AfterValidator` in an `Annotated[]`, but the implementation is still a function - is that okay?
2. Error messages are problematic with union types, inserting extra components in `error["loc"]` and reporting errors for every branch.
    1. The "too many errors" thing can be worked around by using a `Discriminator` and `Tag`s - but the discriminator still takes a function that receives the raw parsed TOML.
    2. Getting rid of the extra entries in `error["loc"]` entirely requires something more like the current logic (preventing Pydantic from having to resolve the union type).
    3. Alternatively we could modify the error printing logic to omit elements of `error["loc"]` that we recognize as implementation details.
3. To be able to natively work with the dotted keys for `attn.o_proj` and `mlp.down_proj` we would need a `dict[ModelComponent, ...]`. `MinLen` probably *does* work but I guess we'd want to enforce at least 1 key, not 2.
4. Defaults should probably be specified on the individual `Parameters` class fields so that individual parameters can be omitted, instead of passing a single `Parameters()` with all the defaults.
5. Custom merging logic could look at the default values via `type(Settings).model_fields["parameters"]` (from memory) for fallbacks if values for a component are not specified.
6. The TOML 1.0 limitation is still awkward - I specified all the per-component values in `config.default.toml` the way I did because if you just have `min_weight_relative = { low = 0.0, high = 1.0 }` you'd have to either write `min_weight_relative = { "attn.o_proj" = { low = 0.0, high = 1.0 }, "mlp.down_proj" = { low = 0.0, high = 1.0 } }` (all one line) or switch to `[parameters.min_weight_relative]` as I did here.

p-e-w:
The TOML 1.0 thing is super unfortunate indeed.

I thought some more about upgrading to Python 3.12, but Transformers 5 still supports 3.10, and it appears that you often get 3.10 by default in cloud environments for older (Ampere-series) GPUs. So I don't think this is an option just yet.

p-e-w:
Wait... [according to the TOML 1.0 spec](https://toml.io/en/v1.0.0#table), the following is possible:

```toml
[dog."tater.man"]
type.name = "pug"
```

Doesn't this completely solve the problem?

spikymoth:
> Wait... [according to the TOML 1.0 spec](https://toml.io/en/v1.0.0#table), the following is possible:
> 
> ```toml
> [dog."tater.man"]
> type.name = "pug"
> ```
> 
> Doesn't this completely solve the problem?

This is essentially what I ended up doing; I went with
```toml
# ... other parameters ...

[parameters.min_weight_relative]
"attn.o_proj"   = { low = 0.0, high = 1.0 }
"mlp.down_proj" = { low = 0.0, high = 1.0 }

# ... other parameters ...
```
for `config.default.toml` instead of
```toml
[parameters]
# ... other parameters ...
min_weight_relative = { low = 0.0, high = 1.0 }
# ... other parameters ...
```
so users wouldn't need to figure out how to go from the latter to the former themselves.

spikymoth:
Oh, right. The reason I brought it up above is that if we want to make supplying a configuration for both components *optional*, explicitly specifying both in `config.default.toml` makes less sense.

p-e-w:
I don't understand. What does this have to do with the TOML 1.0 restrictions?

spikymoth:
Back from my holiday now, I'll see about updating this soon.

> I don't understand. What does this have to do with the TOML 1.0 restrictions?

It just makes usage less intuitive, since you have to move the parameter specification to a sub-section in order to maintain the syntax. For example if we have a default like

```toml
[parameters]
# ...parameters before min_weight_relative...
min_weight_relative = { low = 0.0, high = 1.0 }
# ...parameters after min_weight_relative...
```
and the user wants to give this a per-parameter value, they might expect to be able to do
```toml
[parameters]
# ...parameters before min_weight_relative...
min_weight_relative = {
  "attn.o_proj" = { low = 0.0, high = 1.0 }
  "mlp.down_proj" = { low = 0.0, high = 1.0 }
}
# ...parameters after min_weight_relative...
```
but this would give a syntax error. Instead you have to either put it all on one line, or do
```toml
[parameters]
# ...parameters before min_weight_relative...
# ...parameters after min_weight_relative...

[parameters.min_weight_relative]
"attn.o_proj" = { low = 0.0, high = 1.0 }
"mlp.down_proj" = { low = 0.0, high = 1.0 }
```
which is cleaner IMO. But explaining this in a comment is annoying / verbose, so demonstrating it by making the per-component defaults explicit felt like a better option, especially since omitting values was not supported. Now that we're adding support for partial overrides, putting an explicit per-component default for every value in `config.default.toml` doesn't make much sense anymore.

Anyway it's nothing deal-breaking, especially since we consider this to be an advanced feature for users who can work it out.

p-e-w:
Ah, I get it now. You're right that the TOML 1.1 syntax is a little nicer.

But I looked into this a bit more, and I think we can actually make it work. You wrote

> tomli does support it, but Pydantic uses tomllib for Python 3.11+ so you'd need a hacky override like `sys.modules["tomllib"] = tomli`.

Messing with global modules is of course unacceptable, but there is a much cleaner hack! Check out https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/sources/providers/toml.py.

My proposed "fix" is:

1. Add `tomli` as a dependency to Heretic.

2. Somewhere in our code (written here without any testing, so may require minor adjustment):

```python
import tomli
from pydantic_settings.sources.providers import toml as pydantic_toml
pydantic_toml.tomli = tomli
pydantic_toml.tomllib = tomli
pydantic_toml.import_toml = lambda: None
```

3. Pin our `pydantic_settings` dependency (which I intend to to for all dependencies as a lesson from the LiteLLM hack), to prevent upstream changes breaking this.

That hack is completely confined to a single module. I think that's clean enough, and worth it.

spikymoth:
> Back from my holiday now, I'll see about updating this soon.

I came down with a bad cold pretty much the same day I said that, so I wasn't able to do much of anything in my spare time. Mostly better now, still not 100% though. I'll try to make progress on this in the coming week.

p-e-w:
Please take as much time as you need. Heretic is a long-term project, there's no rush to do anything.

spikymoth:
Mostly a rebase and proof of life, though there are some small fixes and I did start moving the logic into a separate file. I had some major OOM problems on my test system that were pretty demotivating - eventually managed to claw back the performance I lost, but kind of lost the flow I was in before.

Anyway, now seems like a good time to get back into the swing of things :)

spikymoth:
Ah, `StrEnum` is Python 3.11+, Python 3.10 would need `backports.strenum`. I switched to it because with `(str, Enum)`, the values don't stringify nicely (even though comparisons do work). I'll leave it for now.

spikymoth:
Eh, easy enough to fix I guess. Anyway, more refactoring to come Soon™

p-e-w:
Indeed, this a good time to revive this PR, while keeping upcoming architecture changes in mind.

My rough plan is as follows:

1. Merge @red40maxxer's work from #53
2. Extend the plugin system to support "modifier plugins" (that is, plugins that perform abliteration, such as traditional abliteration and ARA)
3. Integrate this PR in a way that caters to modifier plugins. The basic structure should be similar, except that the `Parameters` definition will move into the plugin, since parameters are plugin-specific (ARA uses an entirely different parameter set). In the config file, the parameters will be associated with the plugin name. This is pretty much analogous to what is already going on in #53.

p-e-w:
Alternatively, it would also be possible to merge this PR first, and then use it to inform the design of the modifier plugin system by refactoring it as the system is developed. This might actually be the better approach.

spikymoth:
Okay, I moved almost all of the logic out of the Pydantic classes. I still left a couple of small validators in the `FloatParamSpec` and `UnitParamSpec` classes as I think it's cleaner to have them in the class body.

The logic for discriminating between the different types is now handled via `Tag` and `Discriminator` annotations, and most of the types are now plain Python dicts and lists.

All conversion is now handled by the "Proxy" classes registered on the new `Parameters` class. These classes figure out the actual parameter ranges / choices specified by the user config and defaults, handing back a flat `FloatParam` or `CategoricalParam` object that can be queried directly or used for suggesting parameters.

The downside of this is that parameters now have to be specified twice: Once in the `ParameterSpecification` used by the `Settings` class, and once in the `Parameters` class. But I don't think there's any way around that in plain Python. The actual amount of duplication is pretty minimal - you just specify the name and proxy class.

spikymoth:
One of the more awkward parts here is actually using `ModelComponent` correctly - I can see why they added `StrEnum`, because having to carefully specify `.value` whenever it needs to be stringified (directly or indirectly) is rough.

It might be worth adding https://pypi.org/project/backports.strenum/ so we can use `StrEnum` even for Python 3.10. I didn't do that here because I'm on an AMD system and had to install some different versions of dependencies, so I'm not sure what will happen if I try to update the `uv.lock` file.

src/heretic/parameters.py
```diff
@@ -0,0 +1,433 @@
...
+            return ParamKind.FLAT
+        else:
+            # Assume it's nested for a more helpful error message.
+            return ParamKind.NESTED
+    return None
+
+
+FlatCategoricalParamType = TypeAliasType(
+    "FlatCategoricalParamType",
+    Annotated[
```
p-e-w, line 134: What exactly does the `Annotated` type do here?

src/heretic/parameters.py
```diff
@@ -0,0 +1,433 @@
...
+class DirectionScope(str, Enum):
+    GLOBAL = "global"
+    PER_LAYER = "per layer"
+
+
+class ParameterSpecification(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    direction_scope: FlatCategoricalParamType[DirectionScope] = Field(
+        default=[DirectionScope.GLOBAL, DirectionScope.PER_LAYER],
```
p-e-w, line 369: What does this mean? Which of the two is the default?

config.default.toml
```diff
@@ -199,3 +199,45 @@ column = "text"
 dataset = "mlabonne/harmful_behaviors"
 split = "test[:100]"
 column = "text"
+
+# The parameters used to choose suggest settings for abliteration. With the
+# exception of direction_scope and max_weight, all values are between 0 and 1.
+# min_weight is set relative to max_weight, and the other values are relative
+# to the number of layers in the model.
+# By default, the same ranges are used across all components. For parameters
+# that are sampled per component, the ranges are specified for each component.
```
p-e-w, line 215: Which parameters are not sampled per component?

pyproject.toml
```diff
@@ -73,6 +73,9 @@ heretic = "heretic.main:main"
 requires = ["uv_build>=0.8.11,<0.9.0"]
 build-backend = "uv_build"
 
+[tool.ty.environment]
+python-version = "3.10"
```
p-e-w, line 81: Doesn't ty read `.python-version`? I'm regularly seeing warnings about code that only works in Python >3.10.

src/heretic/parameters.py
```diff
@@ -0,0 +1,433 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# Copyright (C) 2025-2026  Philipp Emanuel Weidmann <pew@worldwidemann.com> + contributors
+
+from enum import Enum
+from typing import Annotated, Any, Generic, Literal, TypeVar, cast
+
+from annotated_types import Interval, IsFinite, MinLen
```
p-e-w, line 12: Is this a built-in module in Python? I can't find proper documentation on that. Same for `typing_extensions`. And does Pydantic actually support those annotations?

p-e-w:
Please write a high-level description of what the code in `parameters.py` actually does and how it works.

What is a "proxy"? Why is everything wrapped in `Annotated`? What is the purpose of using `Discriminator`? Why aren't the compound types simply unions of the base types (e.g., why isn't `UnitParamType` defined simply as `FlatUnitParamType | NestedUnitParamType`)?

Also, it would be helpful if you could list all the specification syntaxes that become possible with this implementation.

src/heretic/parameters.py
spikymoth: [annotated-types](https://pypi.org/project/annotated-types/) is a separate package but is also a [dependency](https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L47) of Pydantic. It provides predicates for the [annotated pattern](https://pydantic.dev/docs/validation/latest/concepts/types/#using-the-annotated-pattern).

[typing_extensions](https://github.com/python/typing_extensions) is an official python module that backports concepts from newer versions of Python without using the newer syntax.

src/heretic/parameters.py
spikymoth: The `Annotated` type lets you add metadata to types without runtime cost. Pydantic uses this metadata to build its runtime validation. In this case we add a `Discriminator` which Pydantic uses to choose the matching branch of the union, matching the `Tag` of each branch against the return value of the `Discriminator`.

The `list[Scalar]` is also annotated with a `MinLen` predicate, which makes Pydantic enforce a minimum number of elements (here, the list must contain at least 1 element).

src/heretic/parameters.py
spikymoth: The whole list is the default; it contains the options that Optuna chooses between.

config.default.toml
spikymoth: `direction_scope` and `direction_fraction` (which together determine `direction_index`).

pyproject.toml
spikymoth: Oh, maybe it does! But `.python-version` is currently set to `3.12`, so I guess that doesn't help.

spikymoth:
> Please write a high-level description of what the code in `parameters.py` actually does and how it works.

I'll write something up.

src/heretic/parameters.py
p-e-w: `annotated_types` must be added as an explicit dependency then. The fact that it is a dependency of Pydantic isn't part of the Pydantic API contract, so it can be removed in any version, even versions that are "compatible" with Heretic.

And I think the same applies for `typing_extensions`, unless I'm missing something. It's an official package maintained by the PSF, but it's still an external package, not part of the standard library.

pyproject.toml
p-e-w: You're right, best to leave this setting then.

I'd love to get rid of Python 3.10, but too many cloud providers still serve it as the default. I guess we'll have to wait for the major libraries like Transformers to move on before we can do the same.

config.default.toml
p-e-w: Ah yes, although this itself is something that the user might want to configure. Is that possible with your current implementation?

src/heretic/parameters.py
p-e-w: So which values are valid here? Any list whose items are of type `DirectionScope`?

p-e-w:
A real-world use case for per-component parameter ranges has just been merged in #387.

spikymoth:
Sorry about the lack of updates here, I've been meaning to get back to it. I'll try to give an overview and also add those explicit dependencies this weekend.

PR #148, opened 2026-02-13
fix: persist hf token between uploads

### LINKED ISSUE #142: Hugging Face token no longer preserved between uploads

red40maxxer:
Minor fix, resolves #142

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -613,12 +613,20 @@ def run():
         email = user.get("email", "no email found")
         print(f"Logged in as [bold]{fullname} ({email})[/]")
 
+    def validate_hf_token(t: str) -> dict[str, Any] | None:
+        try:
+            user = huggingface_hub.whoami(t)
+            print_hf_user_info(user)
+            return user
+        except huggingface_hub.errors.HfHubHTTPError as error:
+            print(f"[red]Failed to validate the Hugging Face token: ({error})[/]")
+            return None
+
     def authenticate_hf(token: str | None) -> tuple[dict[str, Any], str]:
         # Try to use an existing token (from env, hf auth login, or a previous upload).
         if token:
-            try:
-                user = huggingface_hub.whoami(token)
-                print_hf_user_info(user)
+            user = validate_hf_token(token)
+            if user:
                 choice = prompt_select(
                     "How do you want to proceed?",
                     ["Use this account", "Switch account"],
@@ -628,22 +636,16 @@ def run():
                 if choice == "Use this account":
                     return user, token
                 # User chose "Switch account"; fall through to prompt for new token.
-            except huggingface_hub.errors.HfHubHTTPError as error:
-                print(f"[red]Failed to validate the Hugging Face token: ({error})[/]")
-                # Fall through to prompt for a new token.
 
         # No valid token yet (first time, switch account, or validation failed).
         # Prompt for a token until we get a valid one or the user cancels.
         while True:
-            token = prompt_password("Hugging Face access token:")
-            if token is None:
+            new_token = prompt_password("Hugging Face access token:")
+            if new_token is None:
                 raise KeyboardInterrupt
-            try:
-                user = huggingface_hub.whoami(token)
-                print_hf_user_info(user)
-                return user, token
-            except huggingface_hub.errors.HfHubHTTPError as error:
-                print(f"[red]Failed to validate the Hugging Face token: ({error})[/]")
+            user = validate_hf_token(new_token)
+            if user:
+                return user, new_token
 
     while True:
         # If no trials at all have been evaluated, the study must have been stopped


p-e-w:
I see that you have made changes following my comment on Discord yesterday, but IMO neither this version nor the previous one are quite what we want.

The previous version stored the token globally and then used it without prompting once set. This is broken, because it means that if an invalid token is entered, uploading fails and the user cannot proceed without restarting Heretic.

The current version prompts every time, prefilling the prompt with the most recently set token. The problem is that the prompt shows placeholders (`*`) instead of the actual token, so the prefill doesn't convey any information. It's also confusing when the user is already authenticated on their system (which is very common) to be prompted for a token with a mystery prefill, when they expect things to "just work".

Here's how I think it should actually work:

When the user chooses to upload, Heretic should prompt for a token under two conditions:

1. There is no token yet
2. There is a token but authentication fails

Otherwise it just authenticates with the existing token.

If authentication succeeds, Heretic should display the authenticated account (as it does now), *and ask the user whether to proceed with that account or switch to a different one.* If the user decides to switch, Heretic again prompts for a token and the process repeats.

I think this should cover all use cases without ever surprising the user.

src/heretic/main.py
```diff
@@ -598,6 +604,14 @@ def count_completed_trials() -> int:
     if count_completed_trials() == settings.n_trials:
         study.set_user_attr("finished", True)
 
+    def _print_hf_user_info(user: dict):
```
p-e-w: Incomplete type. `dict` has generic arguments. The correct type is probable `dict[str, Any]` or whatever `whoami` returns.

src/heretic/main.py
```diff
@@ -805,23 +847,24 @@ def count_completed_trials() -> int:
...
                                 )
                                 del merged_model
                                 empty_cache()
-                                model.tokenizer.push_to_hub(
-                                    repo_id,
-                                    private=private,
-                                    token=token,
-                                )
+
+                            model.tokenizer.push_to_hub(
```
p-e-w: Restore the indentation; the tokenizer should only get exported with the full model.

src/heretic/utils.py
```diff
@@ -133,12 +133,17 @@ def prompt_path(message: str) -> str:
         return questionary.path(message, only_directories=True).ask()
 
 
-def prompt_password(message: str) -> str:
+def prompt_password(message: str, hf_token: str | None) -> str:
```
p-e-w: This change can be removed I think. Having a default that the user can't inspect doesn't make sense.

src/heretic/main.py
```diff
@@ -765,22 +779,50 @@ def count_completed_trials() -> int:
...
-                            )
-                            email = user.get("email", "no email found")
-                            print(f"Logged in as [bold]{fullname} ({email})[/]")
+                            if hf_token:
+                                try:
+                                    user = huggingface_hub.whoami(hf_token)
+                                except Exception:
+                                    print(
+                                        "[yellow]Warning: Failed to validate the existing Hugging Face token. It might be expired or invalid.[/]"
+                                    )
```
p-e-w: We should set `hf_token` to `None` in this case, as the user might cancel the process and then we print the same message again on the next try.

src/heretic/main.py
```diff
@@ -765,22 +779,50 @@ def count_completed_trials() -> int:
...
+                                if not hf_token:
+                                    break
+
+                                try:
+                                    user = huggingface_hub.whoami(hf_token)
+                                    _print_hf_user_info(user)
+
+                                except Exception:
+                                    print(
+                                        "[red]Invalid token or authentication failed.[/]"
```
p-e-w: Why is this an error but the same situation above is a warning?

src/heretic/main.py
```diff
@@ -765,22 +779,50 @@ def count_completed_trials() -> int:
...
+                                    )
+                                    hf_token = None
```
p-e-w: This block here is very similar to the logic above that uses an existing token. Not sure what can be done here but the overall logic flow is quite hard to understand.

src/heretic/main.py
```diff
@@ -765,22 +779,50 @@ def count_completed_trials() -> int:
...
+
+                                choice = prompt_select(
+                                    "How do you want to proceed?",
+                                    [
+                                        "Use this account",
+                                        "Switch account",
+                                    ],
+                                )
+
+                                if choice == "Switch account":
```
p-e-w: Careful, `choice` will be `None` if the user presses Ctrl+C. With your logic, this is equivalent to choosing "Use this account", which I don't think is correct. Ctrl+C should cancel the process and return to the menu, not pick one of the items.

src/heretic/main.py
red40maxxer: I guess they can both be errors. I made the first one a warning since it doesn't mean the user did something wrong right now as there could be a network issue or something, but the 2nd one is definitely an error to make it clear that a valid token needs to be entered

src/heretic/main.py
red40maxxer: I can't find any documentation with the specific type of what `whoami` returns since the `huggingface_hub` method just does a call to `/api/whoami-v2` and JSONs the body. Assuming `dict[str, Any]` is fine for this as I played around with it and it seems like the correct type

src/heretic/utils.py
```diff
@@ -136,6 +136,7 @@ def prompt_path(message: str) -> str:
 def prompt_password(message: str) -> str:
     if is_notebook():
         print()
+        # getpass doesn't support a default.
```
p-e-w: 🔥

src/heretic/main.py
```diff
@@ -765,22 +796,38 @@ def count_completed_trials() -> int:
...
-                            # We don't use huggingface_hub.login() because that stores the token on disk,
-                            # and since this program will often be run on rented or shared GPU servers,
-                            # it's better to not persist credentials.
-                            token = huggingface_hub.get_token()
-                            if not token:
-                                token = prompt_password("Hugging Face access token:")
-                            if not token:
-                                continue
+                            user = None
+                            if hf_token:
```
p-e-w: Doesn't `validate_hf_token` take care of this conditional?

src/heretic/main.py
```diff
@@ -765,22 +796,38 @@ def count_completed_trials() -> int:
...
+                                    if choice == "Switch account":
+                                        user = None
+                                        hf_token = None
+                                except KeyboardInterrupt:
+                                    break
+
+                            while not user:
+                                hf_token = prompt_password("Hugging Face access token:")
+                                if not hf_token:
+                                    break
```
p-e-w: Can't we simply `continue` the outer loop here and get rid of the additional conditional below?

The overall logical is still way too convoluted. It's like a maze of interdependent conditionals.

src/heretic/main.py
```diff
@@ -598,6 +605,30 @@ def count_completed_trials() -> int:
...
+        )
+        email = user.get("email", "no email found")
+        print(f"Logged in as [bold]{fullname} ({email})[/]")
+
+    def validate_hf_token(
+        token: str | None,
+        *,
+        invalid_message: str = "[red]Invalid token or authentication failed.[/]",
+    ) -> tuple[dict | None, str | None]:
+        """Validate HF token and return (user, token). Returns (None, None) on failure."""
```
p-e-w: Why are we using a special return value to indicate errors, instead of just throwing an exception, which is the standard Python mechanism?

src/heretic/main.py
red40maxxer: yeah, both of the helper functions can just be folded into a single one as well

src/heretic/utils.py
```diff
@@ -133,7 +133,7 @@ def prompt_path(message: str) -> str:
         return questionary.path(message, only_directories=True).ask()
 
 
-def prompt_password(message: str) -> str:
+def prompt_password(message: str) -> str | None:
```
red40maxxer, line 136: If the user CTRL+Cs during password entry, `questionary.password(message).ask()` catches the `KeyboardInterrupt`, prints `"Cancelled by user"` and returns `None` which we handle in line 644

p-e-w: This is true for all other `prompt_*` functions as well.

src/heretic/main.py
```diff
@@ -598,6 +605,48 @@ def count_completed_trials() -> int:
...
+            "fullname",
+            user.get("name", "unknown user"),
+        )
+        email = user.get("email", "no email found")
+        print(f"Logged in as [bold]{fullname} ({email})[/]")
+
+    def validate_hf_token(t: str) -> dict[str, Any] | None:
+        try:
+            user = huggingface_hub.whoami(t)
+            print_hf_user_info(user)
```
p-e-w, line 619: A function called "validate" shouldn't print anything.

src/heretic/main.py
```diff
@@ -598,6 +605,48 @@ def count_completed_trials() -> int:
...
+            return user
+        except huggingface_hub.errors.HfHubHTTPError as error:
```
p-e-w, line 621: Surely this isn't the only possible error here?

src/heretic/main.py
```diff
@@ -598,6 +605,48 @@ def count_completed_trials() -> int:
...
+            print(f"[red]Failed to validate the Hugging Face token: ({error})[/]")
+            return None
+
+    def authenticate_hf(token: str | None) -> tuple[dict[str, Any], str]:
```
p-e-w, line 625: This function should not accept `None`, that doesn't make sense.

src/heretic/main.py
```diff
@@ -598,6 +605,48 @@ def count_completed_trials() -> int:
...
+        # Try to use an existing token (from env, hf auth login, or a previous upload).
+        if token:
+            user = validate_hf_token(token)
+            if user:
+                choice = prompt_select(
+                    "How do you want to proceed?",
+                    ["Use this account", "Switch account"],
+                )
+                if choice is None:
+                    raise KeyboardInterrupt
```
p-e-w, line 635: That's not an acceptable mechanism for bailing out.

p-e-w:
I've taken a look again but it takes me almost 5 minutes to figure out what this code does, which is far too long. The main problem is the helper functions, which take arguments and return results that are counter to their stated purpose.

Let's take a step back and let this issue rest for a while. I'll take a stab at fixing this at some point if I can manage.

PR #151, opened 2026-02-14
fix: extend LoRA merge path to pre-quantized models and fix evaluator div-by-zero

cpagac:
Summary                                                                                                                                                                                                
                                                                                                                                                                                                         
  - Fixes a division-by-zero in the evaluator when base_refusals is 0 (returns refusals rather than 0, so ablation that introduces new refusals is penalized correctly)                                  
  - Extends get_merged_model() and obtain_merge_strategy() to detect pre-quantized models (FP8, MXFP4, GPTQ, etc.) via model.config.quantization_config, not just BNB_4BIT — without this, the CPU reload
   path for LoRA merging wouldn't trigger for pre-quantized models, which have the same limitation                                                                                                       
                                                                                                                                                                                                         
  Test plan

- [X]  - Verify BNB 4-bit quantization and LoRA merge/export still works
- [ ]   - Verify LoRA merge/export works for a pre-quantized model (FP8 or MXFP4) - untested; nvidia FP8 format requires vLLM/TensorRT-LLM and can't run inference via standard HuggingFace transformers
- [X]   - Verify evaluator handles models with 0 base refusals without crashing

### DIFF
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -110,7 +110,9 @@ class Evaluator:
         kl_divergence_scale = self.settings.kl_divergence_scale
         kl_divergence_target = self.settings.kl_divergence_target
 
-        refusals_score = refusals / self.base_refusals
+        refusals_score = (
+            refusals / self.base_refusals if self.base_refusals > 0 else refusals
+        )
 
         if kl_divergence >= kl_divergence_target:
             kld_score = kl_divergence / kl_divergence_scale
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -54,14 +54,16 @@ from .utils import (
 )
 
 
-def obtain_merge_strategy(settings: Settings) -> str | None:
+def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
     """
     Prompts the user for how to proceed with saving the model.
     Provides info to the user if the model is quantized on memory use.
     Returns "merge", "adapter", or None (if cancelled/invalid).
     """
 
-    if settings.quantization == QuantizationMethod.BNB_4BIT:
+    is_quantized = getattr(model.model.config, "quantization_config", None) is not None
+
+    if is_quantized:
         print()
         print(
             "Model was loaded with quantization. Merging requires reloading the base model."
@@ -174,9 +176,15 @@ def run():
     # Adapted from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/env.py
     if torch.cuda.is_available():
         count = torch.cuda.device_count()
-        print(f"Detected [bold]{count}[/] CUDA device(s):")
+        total_vram = sum(torch.cuda.mem_get_info(i)[1] for i in range(count))
+        print(
+            f"Detected [bold]{count}[/] CUDA device(s) ({total_vram / (1024**3):.2f} GB total VRAM):"
+        )
         for i in range(count):
-            print(f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/]")
+            vram = torch.cuda.mem_get_info(i)[1] / (1024**3)
+            print(
+                f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/] ({vram:.2f} GB)"
+            )
     elif is_xpu_available():
         count = torch.xpu.device_count()
         print(f"Detected [bold]{count}[/] XPU device(s):")
@@ -747,7 +755,7 @@ def run():
                             if not save_directory:
                                 continue
 
-                            strategy = obtain_merge_strategy(settings)
+                            strategy = obtain_merge_strategy(settings, model)
                             if strategy is None:
                                 continue
 
@@ -796,7 +804,7 @@ def run():
                             )
                             private = visibility == "Private"
 
-                            strategy = obtain_merge_strategy(settings)
+                            strategy = obtain_merge_strategy(settings, model)
                             if strategy is None:
                                 continue
 
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -223,8 +223,10 @@ class Model:
         # Guard against calling this method at the wrong time.
         assert isinstance(self.model, PeftModel)
 
-        # Check if we need special handling for quantized models
-        if self.settings.quantization == QuantizationMethod.BNB_4BIT:
+        # Check if we need special handling for quantized models.
+        # This covers both on-the-fly quantization (e.g. BNB_4BIT) and pre-quantized
+        # models (e.g. FP8, MXFP4) — both set quantization_config on the model config.
+        if getattr(self.model.config, "quantization_config", None) is not None:
             # Quantized models need special handling - we must reload the base model
             # in full precision to merge the LoRA adapters
 


src/heretic/model.py
```diff
@@ -102,14 +105,27 @@ def __init__(self, settings: Settings):
...
+                    )
+                else:
+                    self.model = get_model_class(settings.model).from_pretrained(
+                        settings.model,
+                        dtype=dtype,
+                        device_map=settings.device_map,
+                        max_memory=self.max_memory,
+                        trust_remote_code=self.trusted_models.get(settings.model),
+                        **extra_kwargs,
+                    )
```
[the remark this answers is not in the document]
cpagac: Fixed — consolidated both from_pretrained calls using a load_kwargs dict.

src/heretic/model.py
```diff
@@ -296,14 +319,25 @@ def reset_model(self):
...
+            )
+        else:
+            self.model = get_model_class(self.settings.model).from_pretrained(
+                self.settings.model,
+                dtype=dtype,
+                device_map=self.settings.device_map,
+                max_memory=self.max_memory,
+                trust_remote_code=self.trusted_models.get(self.settings.model),
+                **extra_kwargs,
+            )
```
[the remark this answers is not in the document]
cpagac: Fixed — same refactor applied here.

kabachuha:
> NVFP4

What about mxfp4 for older hardware?

cpagac:
> > NVFP4
> 
> What about mxfp4 for older hardware?

The implementation here is actually format-agnostic — the "fp8" dtype path simply uses torch_dtype=torch.bfloat16 and lets HuggingFace auto-detect the model's built-in quantization_config.  
  
So if an MXFP4 model is published on HF with the appropriate config, this same loading path should handle it without changes. The naming is admittedly NVFP4-centric since that's what was tested against, but the mechanism itself isn't tied to any specific sub-format. Open to renaming the token to something more general to actually convey this feature, if that makes sense.

src/heretic/evaluator.py
```diff
@@ -110,7 +110,7 @@ def get_score(self) -> tuple[tuple[float, float], float, int]:
         kl_divergence_scale = self.settings.kl_divergence_scale
         kl_divergence_target = self.settings.kl_divergence_target
 
-        refusals_score = refusals / self.base_refusals
+        refusals_score = refusals / self.base_refusals if self.base_refusals > 0 else 0.0
```
p-e-w: Hmm. If `self.base_refusals` is 0, but `refusals` is e.g. 10, this returns the same score as if `refusals` were 0. I don't think this is correct. If the refusal count starts with 0, but is *increased* by ablation, this should result in a very high score, not in a score of 0.

pyproject.toml
```diff
@@ -38,6 +38,9 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+fp8 = [
+    "fp-quant>=0.1.0",
```
p-e-w: Why is this optional?

src/heretic/model.py
```diff
@@ -296,13 +316,21 @@ def reset_model(self):
...
 
+        load_kwargs = {
+            "device_map": self.settings.device_map,
+            "max_memory": self.max_memory,
+            "trust_remote_code": self.trusted_models.get(self.settings.model),
+            **extra_kwargs,
+        }
+        if self._loaded_dtype == _FP8_DTYPE_TOKEN:
+            # FP8/NVFP4 pre-quantized: use torch_dtype instead of dtype=.
+            load_kwargs["torch_dtype"] = torch.bfloat16
```
p-e-w: `torch_dtype` is deprecated, and is, to my understanding, just an alias for `dtype`. What is the purpose of this distinction?

src/heretic/model.py
```diff
@@ -102,13 +105,23 @@ def __init__(self, settings: Settings):
                 if quantization_config is not None:
                     extra_kwargs["quantization_config"] = quantization_config
 
+                load_kwargs = {
+                    "device_map": settings.device_map,
+                    "max_memory": self.max_memory,
+                    "trust_remote_code": self.trusted_models.get(settings.model),
```
p-e-w: There's no need to put those arguments here. They can be kept in the function call below. Python does not care in which order you pass named arguments. It's better to keep them in the function call to get type checking.

p-e-w:
Please explain in a bit more detail what exactly is going on here. My understanding is that this loads models with their *built-in* quantization, just like is already supported for MXFP4. If so, what do we need the extra `dtype` for? If the model tensors are quantized, we always want to load them in that quantized format, no questions asked.

src/heretic/evaluator.py
cpagac: That's a fair point. The original intent was just to prevent the division-by-zero crash I hit during testing. Your example should be penalized heavily, not scored the same as a zero-refusal scenario. I can rework the logic to return a high penalty score when ablation introduces new refusals.

pyproject.toml
cpagac: Fair point... I was messing around with on-the-fly quantization, and completely missed it when proofreading. Now, since this PR deals with the loading pre-quantized FP8 models, and transformers already handle that natively, I don't really think there's any reason for it to be here anymore. I can remove it.

src/heretic/model.py
cpagac: This was a workaround for the "fp8" dtype token, since "fp8" isn't a valid PyTorch dtype. Passing it through dtype= would crash, so I used torch_dtype=torch.bfloat16 to sidestep that. So this workaround could be removed.

src/heretic/model.py
cpagac: This was initially done to reduce duplication. There were two nearly identical from_pretrained calls that only differed by one argument, so the shared args got pulled into a dict. But you're right that it loses type checking, and since the separate FP8 loading branch would be going away, there's no duplication to reduce anymore. I can revert to inline arguments.

cpagac:
> Please explain in a bit more detail what exactly is going on here. My understanding is that this loads models with their _built-in_ quantization, just like is already supported for MXFP4. If so, what do we need the extra `dtype` for? If the model tensors are quantized, we always want to load them in that quantized format, no questions asked.

Fair point, looking into this more and testing it, dtype="auto" already handles FP8 pre-quantized models correctly. HuggingFace auto-detects the quantization_config from the model's config.json the same way it does for MXFP4. (I  also confirmed this by loading nvidia/Llama-3.1-8B-Instruct-FP8 with just dtype="auto" and no special handling, which gives an end product of it loading and generating fine. )

My original thinking was that FP8 needed an explicit opt-in, as BNB_4BIT does, where the user tells heretic to quantize, and heretic applies it at load time. As such, I created an "fp8" dtype token following the same pattern.  I was originally coding this fork for Nemotron support and ran into an issue on nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8. 

I can rework the PR to remove the "fp8" token, the torch_dtype workaround, and the fp-quant dependency. The one thing I think is still relevant is get_merged_model(). Right now, it only has a special CPU reload path for BNB_4BIT models, since quantized weights can't have LoRA adapters merged into them directly. Pre-quantized models (FP8, MXFP4, etc.) share the same limitation: you would need to reload the base model in full precision on the CPU, apply the LoRA weights, and then merge. Without that, I thinkthe merge step would either fail or produce a corrupt model.

src/heretic/evaluator.py
p-e-w: I think simply returning `refusals` should work in that case.

pyproject.toml
p-e-w: Yes, please get rid of all unnecessary stuff. We need to keep things lean.

p-e-w:
Ok. Please trim this PR down to the necessary stuff, then ping me for another review.

src/heretic/evaluator.py
cpagac: Done — changed to else refusals.

pyproject.toml
cpagac: Completed.

cpagac:
Done. Removed the "fp8" dtype token, QuantizationMethod.FP8, the fp-quant dependency, the torch_dtype workaround, and the kwargs dict pattern.

Two things left: the evaluator div-by-zero fix, and an extension to get_merged_model() and obtain_merge_strategy() that detects pre quantized models (FP8, MXFP4, GPTQ, etc.) via model.config.quantization_config rather than just checking for BNB_4BIT. Without that, the CPU reload path for LoRA merging wouldn't trigger for pre-quantized models, which have the same limitation.

Ready for re-review @p-e-w

src/heretic/main.py
```diff
@@ -54,14 +54,16 @@
...
-def obtain_merge_strategy(settings: Settings) -> str | None:
+def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
     """
     Prompts the user for how to proceed with saving the model.
     Provides info to the user if the model is quantized on memory use.
     Returns "merge", "adapter", or None (if cancelled/invalid).
     """
 
-    if settings.quantization == QuantizationMethod.BNB_4BIT:
+    is_quantized = getattr(model.model.config, "quantization_config", None) is not None
```
p-e-w, line 64: Does this work for models quantized on-the-fly by Heretic, i.e., with bitsandbytes?

src/heretic/main.py
```diff
@@ -105,12 +107,7 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
             "How do you want to proceed?",
             choices=[
                 Choice(
-                    title="Merge LoRA into full model"
-                    + (
-                        ""
-                        if settings.quantization == QuantizationMethod.NONE
-                        else " (requires sufficient RAM)"
-                    ),
+                    title="Merge LoRA into full model (requires sufficient RAM)",
```
p-e-w: Please undo this change; the code is the way it is for a good reason (#152).

p-e-w:
Which models have you tested this with? Can you link to Hugging Face uploads made with this PR?

p-e-w:
Any update?

cpagac:
> Any update?

Yes sorry. I'm currently a student so time can be tight as i've been streched thin with homework but i'm working on testing with a few more models and should be done by this weekend.

src/heretic/main.py
cpagac: Done, reverted.

src/heretic/main.py
cpagac: Yes — verified with both models below. The model.config.quantization_config check correctly detects BNB models since HuggingFace stores the BitsAndBytesConfig there on load.

cpagac:
First off, thank you for your patience and apologies for the delay.                                                             

Tested with Qwen/Qwen2-0.5B-Instruct --quantization bnb_4bit and mistralai/Mistral-7B-Instruct-v0.3 --quantization bnb_4bit — CPU reload warning showed on both, merges completed successfully. 
Upload links:
  - https://huggingface.co/cpagac/Qwen2-0.5B-Instruct-heretic
  - https://huggingface.co/cpagac/Mistral-7B-Instruct-v0.3-heretic

  Worth noting: AWQ and GPTQ models fail earlier in heretic's abliteration step (.weight attribute missing on quantized layers), so we can't reach the merge step to test those end-to-end. That's a pre-existing limitation, not something this PR introduced or claims to fix.

To summarize: this PR fixes the evaluator div-by-zero (verified with a unit test), and extends the merge path detection to cover any model HuggingFace auto-detects as quantized via model.config.quantization_config, not just BNB_4BIT.

p-e-w:
So which models can now be processed that couldn't be before?

cpagac:
Honestly, none that I can demonstrate end-to-end today. My thought was that since BNB_4BIT is still a quantized format going through the same merge path, it would serve as a valid proxy, the CPU reload logic is the same regardless of what triggered it. But in hindsight that misses the point, BNB was already handled before this PR so that section doesn't prove anything new.                             
                  
Now **theoretically**, any pre-quantized model where HuggingFace auto-sets quantization_config (FP8, MXFP4) would now correctly hit the CPU reload path instead of skipping it. The logic is sound, but yeah I can't prove it end-to-end as those models fail before reaching the merge step its either missing kernel support at load time (FP8) or the .weight attribute issue in the abliteration step (AWQ/GPTQ). 

That said, I don't think it's permanently theoretical. Once the abliteration step supports quantized layers (right now it assumes .weight on every linear layer which breaks AWQ/GPTQ/FP8), and kernel availability for FP8/MXFP4 improves. Both in my eyes feel like natural next steps for this project. My guess is that these models would flow all the way through and this detection would actually matter but once again its still theoretical as of now. 

 From what I see there are three options going forward (correct me if I'm wrong):
1. Remove the merge path change, keep just the evaluator div-by-zero fix (which is clean and has a unit test) on a separate PR, and let this one sit until the repo is at a state where it's actually valuable.                                                         
2. Close the PR entirely and call it a loss — the last thing I want to do is add code that depreciates the value of the tool.
3. Remove the merge path changes now and just merge the evaluator fix as-is.

p-e-w:
Given the extreme complexity of the ecosystem (hardware, model architectures), I prefer to only merge changes that actually allow us to do something we couldn't do before. Comprehensively testing such changes is very time-consuming for me and so far, I have always managed to discover unforeseen problems afterwards.

I think (1) is the best way forward. The div-by-zero fix is sound and simple, let's just merge that from a clean PR and keep this one for when it becomes relevant.

PR #157, opened 2026-02-14
fix: produce card for local models with existing readme

spikymoth:
This makes it work for my setup. It's not as nice with a local path as we can't get the model name from it, but it's better than no card at all!


p-e-w:
Yup, this should work. Merged!

PR #160, opened 2026-02-15
fix: zero/negative KL divergence output

### LINKED ISSUE #159: Negative/Zero KL Divergence for Qwen3 LLMs
Heretic is producing invalid KL divergence values (-0.0000 and 0.0000) for Qwen3 8B/14B LLMs, while correctly working with Qwen3-4B-Instruct-2507. This appears to be specific to reasoning variants.

```
Running trial 96 of 100...
* Parameters:
  * direction_index = 22.47
  * attn.o_proj.max_weight = 1.42
  * attn.o_proj.max_weight_position = 24.63
  * attn.o_proj.min_weight = 0.26
  * attn.o_proj.min_weight_distance = 19.68
  * mlp.down_proj.max_weight = 1.35
  * mlp.down_proj.max_weight_position = 24.36
  * mlp.down_proj.min_weight = 1.31
  * mlp.down_proj.min_weight_distance = 18.66
* Resetting model...
* Abliterating...
* Evaluating...
  * KL divergence: -0.0000
  * Counting model refusals...
  * Refusals: 22/100

Elapsed time: 10m 54s
Estimated remaining time: 27s
Resident system RAM: 13.71 GB
Allocated GPU VRAM: 27.77 GB
Reserved GPU VRAM: 31.10 GB

Running trial 97 of 100...
* Parameters:
  * direction_index = 20.46
  * attn.o_proj.max_weight = 1.44
  * attn.o_proj.max_weight_position = 24.12
  * attn.o_proj.min_weight = 0.18
  * attn.o_proj.min_weight_distance = 17.06
  * mlp.down_proj.max_weight = 1.31
  * mlp.down_proj.max_weight_position = 26.23
  * mlp.down_proj.min_weight = 1.13
  * mlp.down_proj.min_weight_distance = 16.42
* Resetting model...
* Abliterating...
* Evaluating...
  * KL divergence: 0.0000
  * Counting model refusals...
  * Refusals: 59/100
```

I think this is related to #141.

0xA50C1A1:
Fix #159

Implemented a dynamic evaluation index to stabilize the stats. It finds the first token with <95% probability on the first pass and sticks with it for the rest of the runs.

Since I'm not an ML expert, I'm not sure if this 95% threshold is a solid heuristic or just a "works for me" hack. Would appreciate some feedback on whether this logic is sound.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 # Heretic: Fully automatic censorship removal for language models<br><br>[![Discord](https://img.shields.io/discord/1447831134212984903?color=5865F2&label=discord&labelColor=black&logo=discord&logoColor=white&style=for-the-badge)](https://discord.gg/gdXc48gSyT) [![Follow us on Hugging Face](https://huggingface.co/datasets/huggingface/badges/resolve/main/follow-us-on-hf-md-dark.svg)](https://huggingface.co/heretic-org)
 
+[![#1 Repository of the Day](https://trendshift.io/api/badge/repositories/20538)](https://trendshift.io/repositories/20538)
+
 Heretic is a tool that removes censorship (aka "safety alignment") from
 transformer-based language models without expensive post-training.
 It combines an advanced implementation of directional ablation, also known
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -174,9 +174,15 @@ def run():
     # Adapted from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/env.py
     if torch.cuda.is_available():
         count = torch.cuda.device_count()
-        print(f"Detected [bold]{count}[/] CUDA device(s):")
+        total_vram = sum(torch.cuda.mem_get_info(i)[1] for i in range(count))
+        print(
+            f"Detected [bold]{count}[/] CUDA device(s) ({total_vram / (1024**3):.2f} GB total VRAM):"
+        )
         for i in range(count):
-            print(f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/]")
+            vram = torch.cuda.mem_get_info(i)[1] / (1024**3)
+            print(
+                f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/] ({vram:.2f} GB)"
+            )
     elif is_xpu_available():
         count = torch.xpu.device_count()
         print(f"Detected [bold]{count}[/] XPU device(s):")
@@ -823,11 +829,23 @@ def run():
                                     token=token,
                                 )
 
-                            # If the model path doesn't exist locally, it can be assumed
-                            # to be a model hosted on the Hugging Face Hub, in which case
+                            # If the model path exists locally and includes the
+                            # card, use it directly. If the model path doesn't
+                            # exist locally, it can be assumed to be a model
+                            # hosted on the Hugging Face Hub, in which case
                             # we can retrieve the model card.
-                            if not Path(settings.model).exists():
+                            model_path = Path(settings.model)
+                            if model_path.exists():
+                                card_path = (
+                                    model_path / huggingface_hub.constants.REPOCARD_NAME
+                                )
+                                if card_path.exists():
+                                    card = ModelCard.load(card_path)
+                                else:
+                                    card = None
+                            else:
                                 card = ModelCard.load(settings.model)
+                            if card is not None:
                                 if card.data is None:
                                     card.data = ModelCardData()
                                 if card.data.tags is None:
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -38,11 +38,17 @@ def print_memory_usage():
     p("Resident system RAM", Process().memory_info().rss)
 
     if torch.cuda.is_available():
-        p("Allocated GPU VRAM", torch.cuda.memory_allocated())
-        p("Reserved GPU VRAM", torch.cuda.memory_reserved())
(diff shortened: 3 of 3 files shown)


p-e-w:
We already have a mechanism for detecting and skipping deterministic prefixes:

https://github.com/p-e-w/heretic/blob/3a115e280c3f42b0f33ce4fb8aba685447c04c08/src/heretic/main.py#L372-L401

If that isn't enough, the problem is somewhere else.

0xA50C1A1:
> We already have a mechanism for detecting and skipping deterministic prefixes:
> 
> https://github.com/p-e-w/heretic/blob/3a115e280c3f42b0f33ce4fb8aba685447c04c08/src/heretic/main.py#L372-L401
> 
> If that isn't enough, the problem is somewhere else.

Thanks! I've checked that part of the code, but it seems the common prefix isn't the whole story. In my tests with Qwen3, there are often 1-2 tokens right after the `</think>` tag that are still 99.9% deterministic (like `\n` or whitespace characters). 

Since the current KL evaluation only checks one token position, it gets stuck on these deterministic bottlenecks. My PR adds a way to skip that "deterministic bridge" and measure KL where it actually matters: at the first point of non-trivial choice. It complements the existing logic.

src/heretic/model.py
```diff
@@ -656,25 +657,36 @@ def get_residuals_batched(self, prompts: list[Prompt]) -> Tensor:
     # We work with logprobs rather than probabilities for numerical stability
     # when computing the KL divergence.
     def get_logprobs(self, prompts: list[Prompt]) -> Tensor:
-        # We only generate one token, and we return the (log) probability distributions
-        # over the vocabulary at that token position, for each prompt.
+        # Generate a small lookahead window to skip deterministic prefix tokens.
+        max_lookahead = 10
```
p-e-w, line 661: This should be adjusted after the actual index has been determined, otherwise we're wasting tokens on every future generation.

src/heretic/model.py
```diff
@@ -656,25 +657,36 @@ def get_residuals_batched(self, prompts: list[Prompt]) -> Tensor:
...
         _, outputs = self.generate(
             prompts,
-            max_new_tokens=1,
+            max_new_tokens=max_lookahead,
             output_scores=True,
             return_dict_in_generate=True,
         )
 
-        # This cast is valid because GenerateDecoderOnlyOutput is the return type
-        # of model.generate with return_dict_in_generate=True.
```
p-e-w, line 669: Don't remove this comment.

src/heretic/model.py
```diff
@@ -656,25 +657,36 @@ def get_residuals_batched(self, prompts: list[Prompt]) -> Tensor:
...
         outputs = cast(GenerateDecoderOnlyOutput, outputs)
-
-        # Logits for the first (only) generated token.
-        # This cast is valid because we passed output_scores=True above.
-        logits = cast(tuple[FloatTensor], outputs.scores)[0]
-
-        # The returned tensor has shape (prompt, token).
-        return F.log_softmax(logits, dim=-1)
+        scores = cast(tuple[FloatTensor], outputs.scores)
```
p-e-w, line 670: Restore the comment from above explaining why the cast is valid. Casts are potentially unsafe (in a type-safety sense) and every cast must be justified.

src/heretic/model.py
```diff
@@ -656,25 +657,36 @@ def get_residuals_batched(self, prompts: list[Prompt]) -> Tensor:
...
+
+        # Determine the evaluation token index on the first run (using the base model).
+        # This ensures stable KL divergence across all trials.
+        if self.eval_token_index is None:
+            for i, logits in enumerate(scores):
+                # Calculate the average top-1 token probability across the batch.
+                probs = F.softmax(logits, dim=-1)
```
p-e-w, line 677:
```suggestion
                probabilities = F.softmax(logits, dim=-1)
```

We use the full terms everywhere for clarity.

src/heretic/model.py
```diff
@@ -656,25 +657,36 @@ def get_residuals_batched(self, prompts: list[Prompt]) -> Tensor:
...
+                avg_top_prob = top_probs.mean().item()
+
+                # If the average probability is below the threshold, the model is making a choice.
+                if avg_top_prob < self.settings.kl_divergence_token_threshold or i == len(scores) - 1:
+                    self.eval_token_index = i
+                    print(f"  * Evaluation token index set to: [bold]{i}[/]")
+                    break
+
+        # Use the fixed index to ensure KL divergence compares the same token positions.
+        actual_index = min(self.eval_token_index, len(scores) - 1)
```
p-e-w, line 688: Why is this necessary? How can `eval_token_index` be greater than the number of scores?

src/heretic/config.py
```diff
@@ -176,6 +176,14 @@ class Settings(BaseSettings):
         ),
     )
 
+    kl_divergence_token_threshold: float = Field(
```
p-e-w, line 179: I don't like this name too much.

p-e-w:
You have made a valuable discovery and your basic idea is sound, but I don't think this is the correct solution.

We should do it the way we already do (append the common prefix to the template output), except using a "fuzzy" approach where a prefix is considered common already if it is shared by 95% (or similar) of outputs.

The key advantage of doing it that way is that it also improves the residual computation. Currently, you're only fixing the KLD, but the underlying problem also applies to the residuals and suggests that they could be computed more accurately if the hidden states were obtained at a token position a few indices further ahead.

PR #162, opened 2026-02-15
Sharded LoRA merge

### LINKED ISSUE #156: [Feature request] Sharded LoRA merge
Currently, the export algorithm requires the full model to be loaded into RAM before the LoRA merge and saving to drive.

However, it can be frustrating if the model size is much more than the RAM capability.

Yet, it is possible to apply LoRA per model shard. For each o_proj and mlp_proj component lora is simply a set of two low-rank matrices A and B, which are being added to the model's weight processing at inference. It is possible to fuse them into the weights using the simple algorithm.

```python
    lora_weight = (lora_module.lora_A.weight @ lora_module.lora_B.weight) * lora_module.scaling
    module.weight += lora_weight.to(module.weight.dtype)
```

What I propose is to

1. For the sharded transformer model like `model-00001-of-00010.safetensors` load each shard one by one.
2. Iterate over the tensors in the shard.
3. For the shard's target components, find the corresponding A and B matrices in the LoRA.
4. Modify the weights by adding the matrix product.

This way will allow exporting very large models with smaller RAM.

Most quantization frameworks like llama.cpp read the model layer by layer and therefore too can operate in low RAM.

kabachuha:
Vanilla LoRA peft merge requires the full dequantized model to reside in RAM, that is disastrous for very big models which weight much more than the RAM capability (they can be abliterated in 4bit thanks for BNB, but the saving is problematic).

The PR finds the actual location of the HF model and goes through its shards .safetensors by .safetensors. For each tensor, if it is targeted by LoRA, the corresponding LoRA key is mapped and the weight is modified by adding the matrix product.

In the end, the script checks whether all the LoRA keys have been inserted for integrity.

Closes #156.

P.S. Also hope you will return the LoRA export functionality eventually 😬

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -71,7 +71,7 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
         )
         print("[yellow]This can lead to system freezes if you run out of memory.[/]")
         print(
-            "[yellow] However, you can choose to save model shard by shard. It will be slightly slower, but require <10 gb RAM usage for most models.[/]"
+            "[yellow] However, you can choose to save the model shard-by-shard. It is slightly slower, but requires <10 GB of RAM for most models.[/]"
         )
 
         try:
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -70,6 +70,9 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
             "[yellow]WARNING: CPU merging requires dequantizing the entire model to system RAM.[/]"
         )
         print("[yellow]This can lead to system freezes if you run out of memory.[/]")
+        print(
+            "[yellow] However, you can choose to save model shard by shard. It will be slightly slower, but require <10 gb RAM usage for most models.[/]"
+        )
 
         try:
             # Estimate memory requirements by loading the model structure on the "meta" device.
@@ -88,13 +91,13 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
                 footprint_bytes = meta_model.get_memory_footprint()
                 footprint_gb = footprint_bytes / (1024**3)
                 print(
-                    f"[yellow]Estimated RAM required (excluding overhead): [bold]~{footprint_gb:.2f} GB[/][/]"
+                    f"[yellow]Estimated RAM in non-sharded mode required (excluding overhead): [bold]~{footprint_gb:.2f} GB[/][/]"
                 )
         except Exception:
             # Fallback if meta loading fails (e.g. owing to custom model code
             # or bitsandbytes quantization config issues on the meta device).
             print(
-                "[yellow]Rule of thumb: You need approximately 3x the parameter count in GB RAM.[/]"
+                "[yellow]Rule of thumb: In non-sharded mode, you need approximately 3x the parameter count in GB RAM.[/]"
             )
             print(
                 "[yellow]Example: A 27B model requires ~80GB RAM. A 70B model requires ~200GB RAM.[/]"
@@ -112,6 +115,15 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
                         else " (requires sufficient RAM)"
                     ),
                     value="merge",
+                ),
+                Choice(
+                    title="Merge LoRA into full model (sharded)"
+                    + (
+                        ""
+                        if settings.quantization == QuantizationMethod.NONE
+                        else " (very low RAM usage)"
+                    ),
+                    value="merge_sharded",
                 ),
                 Choice(
                     title="Cancel",
@@ -754,6 +766,9 @@ def run():
                             if strategy == "adapter":
                                 print("Saving LoRA adapter...")
                                 model.model.save_pretrained(save_directory)
+                            elif strategy == "merge_sharded":
+                                print("Saving merged model file by file...")
+                                model.save_sharded(settings.model, save_directory)
                             else:
                                 print("Saving merged model...")
                                 merged_model = model.get_merged_model()
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -299,11 +299,7 @@ class Model:
                     'A': lora_state_dict[key],
                     'B': lora_state_dict.get(key.replace("lora_A", "lora_B"))
                 }
-        
-        # Load model config to check dtype
-        with open(os.path.join(model_path, "config.json"), 'r') as f:
-            config = json.load(f)
-        
+
         # Get all safetensors shards
         model_files = sorted(glob.glob(os.path.join(model_path, "*.safetensors")))
         if not model_files:
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -280,10 +280,7 @@ class Model:
         # 1. resolve the local path
         if not os.path.exists(model_path):
             print("Model path not found locally, attempting to locate in HF cache.")
-            try:
-                hf_hub_location = os.environ["HF_HUB_CACHE"]
-            except KeyError:
-                raise ValueError("HF_HUB_CACHE environment variable is not set!")
+            hf_hub_location = os.environ.get("HF_HUB_CACHE", os.path.join(os.environ.get("HOME", ""), ".cache/huggingface/hub"))
             
             model_name_cached = "models--" + model_path.replace("/", "--")
             model_path = os.path.join(hf_hub_location, model_name_cached)
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -5,7 +5,9 @@ import math
 from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any, Type, cast
-import os, glob, json
+import os
+import glob
+import json
 from safetensors import safe_open, save_file
 
 import bitsandbytes as bnb
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -268,7 +268,7 @@ class Model:
             return merged_model
 
 
-    def save_sharded(self, model_path, save_directory) -> PreTrainedModel:
+    def save_sharded(self, model_path: str, save_directory: str) -> None:
 
         assert isinstance(self.model, PeftModel)
 
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -339,8 +339,6 @@ class Model:
                         # Add to original weight
                         tensor = tensor + lora_weight
                         
-                        print(f"Merged LoRA for {key}")
-                        
                     merged_tensors[key] = tensor
             
             # Save merged shard
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -8,6 +8,7 @@ from typing import Any, Type, cast
 import os
 import glob
 import json
+import shutil
 from safetensors import safe_open, save_file
 
 import bitsandbytes as bnb
@@ -356,7 +357,6 @@ class Model:
                 src = os.path.join(model_path, file)
                 dst = os.path.join(save_directory, file)
                 if os.path.isfile(src):
-                    import shutil
                     shutil.copy2(src, dst)
 
         assert len(lora_pairs) == 0, "Not all LoRA keys have been injected"
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -5,6 +5,8 @@ import math
 from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any, Type, cast
+import os, glob, json
+from safetensors import safe_open, save_file
 
 import bitsandbytes as bnb
 import torch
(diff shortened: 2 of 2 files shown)


src/heretic/model.py
```diff
@@ -265,6 +267,108 @@ def get_merged_model(self) -> PreTrainedModel:
...
+                        raise ValueError(
+                            f"Tensor {key} has dtype {tensor.dtype} - "
+                            "quantized base weights not supported for merging"
+                        )
+                    
+                    # Check if this layer has LoRA weights
+                    if key in lora_pairs:
+                        lora_item = lora_pairs.pop(key)
+                        lora_A = lora_item['A']
+                        lora_B = lora_item['B']
```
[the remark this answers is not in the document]
kabachuha: They should exist by Peft LoRA definition

src/heretic/model.py
```diff
@@ -265,6 +267,108 @@ def get_merged_model(self) -> PreTrainedModel:
...
+                        
+                        # Verify shapes
+                        assert lora_A.shape[0] == tensor.shape[0], \
+                            f"Shape mismatch for {key}: tensor {tensor.shape}, lora_A {lora_A.shape}"
```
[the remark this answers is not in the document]
kabachuha: If the lora is computed fine, this should be fine as well

src/heretic/main.py
```diff
@@ -70,6 +70,9 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
             "[yellow]WARNING: CPU merging requires dequantizing the entire model to system RAM.[/]"
         )
         print("[yellow]This can lead to system freezes if you run out of memory.[/]")
+        print(
+            "[yellow] However, you can choose to save the model shard-by-shard. It is slightly slower, but requires <10 GB of RAM for most models.[/]"
+        )
```
p-e-w, line 75: I don't think we need this extra complexity. Always using sharded merging should be fine.

src/heretic/model.py
```diff
@@ -265,6 +270,98 @@ def get_merged_model(self) -> PreTrainedModel:
             self.needs_reload = True
             return merged_model
 
+
+    def save_sharded(self, model_path: str, save_directory: str) -> None:
+
+        assert isinstance(self.model, PeftModel)
```
p-e-w, line 276: When is this not true?

src/heretic/model.py
```diff
@@ -265,6 +270,98 @@ def get_merged_model(self) -> PreTrainedModel:
...
+
+        lora_state_dict = self.model.state_dict()
+
+        # 1. resolve the local path
+        if not os.path.exists(model_path):
+            print("Model path not found locally, attempting to locate in HF cache.")
+            hf_hub_location = os.environ.get("HF_HUB_CACHE", os.path.join(os.environ.get("HOME", ""), ".cache/huggingface/hub"))
```
p-e-w, line 283: This could be dicey. I think that if Xet is used (which HF is aggressively migrating towards), caches work quite differently, and aren't simply folders containing the model repository. It's an object storage similar to Git.

src/heretic/model.py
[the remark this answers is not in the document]
p-e-w: Correct.

p-e-w:
Xet is my main concern. I'm not on my workstation right now to verify this, but IIRC, with Xet, downloaded models are *not* stored in a plain folder that simply contains the safetensor files.

src/heretic/model.py
kabachuha: This is the same logic as in the start of `def get_merged_model(self) -> PreTrainedModel:`

```python
    def get_merged_model(self) -> PreTrainedModel:
        # Guard against calling this method at the wrong time.
        assert isinstance(self.model, PeftModel)

        # Check if we need special handling for quantized models
        if self.settings.quantization == QuantizationMethod.BNB_4BIT:
...
```

Which is then referenced in:

```python
...
                      if strategy == "adapter":
                          print("Saving LoRA adapter...")
                          model.model.save_pretrained(save_directory)
                      elif strategy == "merge_sharded":
                          print("Saving merged model file by file...")
                          model.save_sharded(settings.model, save_directory)
                      else:
                          print("Saving merged model...")
                          merged_model = model.get_merged_model()
                          merged_model.save_pretrained(save_directory)
                          del merged_model
                          empty_cache()
                          model.tokenizer.save_pretrained(save_directory)
...
```

p-e-w: Ah, ok. Sure.

src/heretic/model.py
kabachuha: @p-e-w In this case heretic can do "from_pretrained" on the model and then do "save_pretrained" into a temp folder, on which the sharded ablation will be performed. Sharded modification is just too good not to have it

kabachuha: Of course, a warning similar to the current one ram should be displayed, if the model wasn't saved to a local folder beforehand (and add a hint in the documentation what local saving is preferred and maybe enable it by default)

kabachuha:
Yes, I just looked into the .cache save folders and their structure is really complicated.

So, I suggest this variant: from now on, by default, heretic saves the model into a local dir via `huggingface_hub download`. Say, `./models/...`. with .gitignore. The location is to be set in the .toml config if the user wants to put it into another place, so it will not change the "fully automated" philisophy.

For backwards compatibility, the model is copied or moved via `from_pretrained` and `save_pretrained` with a warning.

p-e-w:
> So, I suggest this variant: from now on, by default, heretic saves the model into a local dir via `huggingface_hub download`.

No, that doesn't fly. The user may already have the model cached locally (the whole purpose of the cache is for multiple applications to share the same model data). We can't just download the whole thing again, wasting I don't know how many Gigabytes of disk space.

There is also the maintainability perspective. Managing the download ourselves is a hack because it bypasses Transformers' built-in download system, and the sharded merge itself is also a hack because it bypasses Transformers' built-in export system. You can't build a reliable piece of software by layering hacks on top of hacks.

This is obviously unfortunate. Sharded merge is cool. But it's not cool enough to turn the core of our system into a brittle mess. If this cannot be made to work directly with Xet, I'm afraid the approach will have to be abandoned. :disappointed:

kabachuha:
Maybe we still can have it, but with obvious warnings and documentation? For example, in many prebuilt server configurations, there is not greater than 128+ gb for abliterating and saving even 70b models. (they require 141+ GB RAM)

And there is currently *no warning* about the RAM usage before the start of the abliteration process, this can take the low-RAM users by surprise after they have already run trials

kabachuha:
Take this configuration: 1xH100, 128 GB RAM. This way you can abliterate even a 100B+ model in 4bit precision, but there is absolutely no way of saving it unless you rent a *ton* of additional RAM

Or a more amateur 2x4090 / 5090 for numerous LLaMA 70b fine-tunes

kabachuha:
The perfect approach would be, of course, to fix the LoRA system as with LoRA the user can decide what to do with it themselves. Save with peft, sharded or not, manually or simply convert with llama.cpp with no hustle and use it with compact Unsloth quants without quantizing the model again themselves (and losing the advantage of dynamic quants)

Is it possible for you look into the lora export system again?

p-e-w:
> Take this configuration: 1xH100, 128 GB RAM.

That's an extremely uncommon configuration. Pairing a $20,000 GPU with $300 worth of RAM (maybe a little more expensive currently) makes very little sense. A typical research workstation with an H100 would have 512 GB of RAM or even more.

Maintaining the quality and usability of an already complex program like Heretic requires keeping complexity in check. As far as I can tell, having the VRAM to abliterate a model in Q4 but not the RAM to merge it in BF16 (that is, RAM < ~4 * VRAM), *and* no alternative setup, should be a very unusual situation. We cannot introduce that degree of complexity at such a fundamental level for such a corner case.

> Is it possible for you look into the lora export system again?

As far as I can tell, we are doing everything correctly, and this is likely a bug in PEFT, or a problem with the interaction of Transformers and PEFT (perhaps a version mismatch). This is especially likely considering that saving the LoRA locally appears to work while the HF upload fails.

kabachuha:
> Take this configuration: 1xH100, 128 GB RAM.

This is a rentable configuration on one of the clouds I use. Mind you, where I live (Russia) H100 / H200 availability is extremely low and even 3xH100 servers are all rented-out for unknown amount of time, so hobbyists are having not much choice when renting.

> This is especially likely considering that saving the LoRA locally appears to work while the HF upload fails.

I think it is fixable and can be reported to the huggingface repo in an issue / push a pull request, one needs to simply look there.

> We cannot introduce that degree of complexity at such a fundamental level for such a corner case.

The VRAM / RAM availability fraction is getting skewed recently because of the RAM market collapse and it is quite important for people quantizing ~70b models on home setups.

dariox1337:
I'm sorry, but can this function be merged at least as an option for manually downloaded files? There are people who prefer to run apps in isolated environments and use pre-downloaded shards. So hf cache isn't even used.

p-e-w:
> The VRAM / RAM availability fraction is getting skewed recently because of the RAM market collapse and it is quite important for people quantizing ~70b models on home setups.

Please understand that I would love to have this feature. But a custom download system with major caveats regarding disk space, plus a custom export system, both of which replace systems that are already built into Transformers, is simply too high a price to pay IMO. Doing things like that is a superhighway towards a program that has weird bugs that nobody can understand, and that breaks in unforeseeable ways when Transformers changes something.

This is ultimately a feature that belongs upstream. Upstream doesn't have it yet, but that doesn't mean Heretic is the right place to put it.

kabachuha:
> This is ultimately a feature that belongs upstream.

PEFT is probably not the place too. It's all about playing with LoRAs in runtime, maybe with save and load once, not with handling the saved model's .safetensors files, because it's not related to the PEFT's runtime LoRAs architecture, and because of it a code to handle the shards is needed *here*. On contrary, because it doesn't really below to the PEFT or the transformers codebase, it has to be implemented here.

> that breaks in unforeseeable ways when Transformers changes something.

Safetensors format is fixed and it simply contains the weights corresponding to the model's layers. The PEFT LoRA adapter formed in heretic targets these modules with lora A and lora B linked to the model's layers. The iteration logic outlined in this pull request goes over the model's parameters shard by shard, the linked A and B matrices are found in the dictionary. The chance of PEFT or transformers changing the safetensors storage is *zero*, otherwise it will break **all tools, including LLaMA.cpp,** built around transformers and the Huggingface admins definitely don't want this.

The carryover of cached Xet transformer files can be circumvented completely if we will use

[from_pretrained(..., cache_dir="...")](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained) or with [hf_hub_download](https://huggingface.co/docs/huggingface_hub/guides/download), enabled by default into a created `./models` subdirectory, with the HF_HUB_CACHE downloads disabled at all.

p-e-w:
> PEFT is probably not the place too.

Where else? This is a standard use case and not at all specific to Heretic. Training a LoRA and merging it into the base model to create a merged model is completely generic functionality. It's not even specific to abliteration. It either belongs in Transformers, or in PEFT, or in a dedicated library.

p-e-w:
> The carryover of cached Xet transformer files can be circumvented completely if we will use
>
> [from_pretrained(..., cache_dir="...")](https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained) or with [hf_hub_download](https://huggingface.co/docs/huggingface_hub/guides/download), enabled by default into a created ./models subdirectory, with the HF_HUB_CACHE downloads disabled at all.

Yes, but this re-downloads/duplicates models that are already in the user's HF cache, breaking a key assumption about how the HF ecosystem is supposed to work.

It also fails in the other direction: Models downloaded with Heretic aren't available to other programs using Transformers.

Again: I wish it wasn't so. But the ecosystem just isn't ready yet to implement this feature cleanly.

kabachuha:
> It either belongs in Transformers, or in PEFT, or in a dedicated library.

This is literally a single 84 lines function, with solidified logic, depending on arguably not-breakable peft<->transformers naming correspondences.

> Models downloaded with Heretic aren't available to other programs using Transformers.

You can perfectly pass local paths to any transformers from_pretrained-based scripts.

As for the disk space and if you *really* need to use HF's cache system, with heretic you are supposed to abliterate the model once and forget it for good. It is not much a sacrifice to download the model twice.

Disk space is MUCH more available than RAM, with its prices gone insane.

spikymoth:
FWIW, a Strix Halo-based system (~$2000 before recent price hikes) can also have up to 128GB of (semi-unified) RAM, and can't be upgraded. While such a system is not as powerful (and would probably take a long time on models that require 4-bit quantization to be loaded into RAM), I think this feature would still be quite nice to have.

I do downloads with `hf download --local-dir` which still works fine with XET as far as I can tell, but doesn't create the opaque cache structure (it creates a directory with the raw files and a `.cache` directory for metadata).

p-e-w:
I think the right way forward here is just to make LoRA export work correctly. There are too many externalities to handle, and hacks like maintaining our own cache not only add tremendous complexity, they also risk surprising the user in unpleasant ways when we eat up their disk space.

If we offer to export a LoRA as an alternative to loading the full model into system RAM, the user can do what they want with it, including performing the merge process of their choice using other tools.

kabachuha:
@p-e-w Any updates on the LoRA situation?

p-e-w:
@kabachuha Unfortunately I don't have time to work on that at the moment. It will take a few weeks for me to get to this task at least.

kabachuha:
@p-e-w FYI, unsloth can do local xet-HF / local-dir / remote hub resolution, copying and sharded LoRA merge out of the box. We can include it as an optional dependency for low-RAM systems

p-e-w:
Can it directly upload to HF, without making a full copy of the model locally on disk?

kabachuha:
I don't think so. It will require a lot of hacking: copy a file, apply lora, do push file to hub in a loop

Why do you care of the disk space so badly? It is much cheaper than RAM / renting RAM. Heretic can add a disclaimer before the start, just like there is a disclaimer about the RAM spike when saving (btw, the disclaimer is too late, the user can abliterate the model and then be shocked that they cannot save it)

p-e-w:
LoRA export is now working, so we don't need such complicated logic anymore. If the user wants a special merge strategy they can export the LoRA and do it manually with a tool of their choice.

PR #165, opened 2026-02-15
Fix: Detect ROCm and disable incompatible bitsandbytes quantization

noctrex:
On ROCm it returns True for torch.cuda.is_available(), causing bitsandbytes to attempt loading ROCm-specific binaries which are not available, resulting in crashes.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -377,7 +377,8 @@ def run():
 
     print()
     print("Checking for common response prefix...")
-    responses = model.get_responses_batched(good_prompts[:100] + bad_prompts[:100])
+    prefix_check_prompts = good_prompts[:100] + bad_prompts[:100]
+    responses = model.get_responses_batched(prefix_check_prompts)
 
     # Despite being located in os.path, commonprefix actually performs
     # a naive string operation without any path-specific logic,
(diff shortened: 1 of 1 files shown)


src/heretic/model.py
```diff
@@ -29,8 +27,11 @@
     GenerateDecoderOnlyOutput,  # ty:ignore[possibly-missing-import]
 )
 
+if TYPE_CHECKING:
+    from transformers import BitsAndBytesConfig
```
p-e-w: Why did you move this? It was already imported.

src/heretic/model.py
```diff
@@ -194,7 +195,7 @@ def _apply_lora(self):
 
         print(f"* LoRA adapters initialized (targets: {', '.join(target_modules)})")
 
-    def _get_quantization_config(self, dtype: str) -> BitsAndBytesConfig | None:
```
p-e-w, line 197: Has the return type changed? I don't understand this change.

src/heretic/model.py
```diff
@@ -205,6 +206,19 @@ def _get_quantization_config(self, dtype: str) -> BitsAndBytesConfig | None:
...
         """
         if self.settings.quantization == QuantizationMethod.BNB_4BIT:
+            # Check if ROCm is detected - bitsandbytes is not compatible with ROCm
+            if is_rocm_available():
+                print(
+                    "[bold yellow]Warning: bitsandbytes quantization is not supported on ROCm.[/]"
+                )
+                print(
+                    "[bold yellow]Quantization will be disabled. Model will load in full precision.[/]"
+                )
```
p-e-w: This is not the correct approach. The user specifically requested quantization. We can't oblige. That means we throw an error and exit. We don't do something other than what the user requested.

p-e-w:
The bitsandbytes docs state that the alpha release of bnb has full ROCm support. I wonder if it's really worth implementing such extra logic when everything will just work in a future release anyway. The user can even install the alpha themselves on ROCm if they want.

noctrex:
> The bitsandbytes docs state that the alpha release of bnb has full ROCm support. I wonder if it's really worth implementing such extra logic when everything will just work in a future release anyway. The user can even install the alpha themselves on ROCm if they want.

I thought about submitting this change. About why I did it? This is what I am using on my AMD system.
It's a hacky solution I've come up with. Please ignore it if it does not represent the quality of your code.
Maybe can be further refined to be used only on Windows specifically.

It's about making bitsandbytes optional on ROCm, because not everyone will know to install custom alpha versions of it.
Also this custom alpha version works only on Linux. Me being on Windows, it just simply does not work. So I have to disable it.

Unfortunately, we AMD users will always be "Children of a Lesser God".

output with the latest alpha installed on windows
```
(.venv) R:\heretic>heretic
Could not detect ROCm GPU architecture: [WinError 2] The system cannot find the file specified

ROCm GPU architecture detection failed despite ROCm being available.

Could not detect ROCm warp size: [WinError 2] The system cannot find the file specified. Defaulting to 64. (some 4-bit functions may not work!)

ROCm warp size detection failed despite ROCm being available.

bitsandbytes library load error: Configured ROCm binary not found at R:\heretic\.venv\Lib\site-packages\bitsandbytes\libbitsandbytes_rocm72.dll
  File "R:\heretic\.venv\Lib\site-packages\bitsandbytes\cextension.py", line 320, in <module>
    lib = get_native_library()
          ^^^^^^^^^^^^^^^^^^^^
  File "R:\heretic\.venv\Lib\site-packages\bitsandbytes\cextension.py", line 288, in get_native_library
    raise RuntimeError(f"Configured {BNB_BACKEND} binary not found at {cuda_binary_path}")
RuntimeError: Configured ROCm binary not found at R:\heretic\.venv\Lib\site-packages\bitsandbytes\libbitsandbytes_rocm72.dll
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░

Configuration contains 1 errors:
model: Field required

Run heretic --help or see config.default.toml for details about configuration parameters.
```

p-e-w:
But isn't bitsandbytes already optional? It's not getting used unless you enable quantization, right?

spikymoth:
bitsandbytes does have ROCm support, at least on Linux. There's an open issue for Windows support: https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1844

p-e-w:
Can someone confirm that the bnb support for ROCm on Linux actually works with Heretic? I don't have any AMD hardware.

If it does work, I don't think the complexity of this PR is worth it just to support Windows. That being said, the reporting change where ROCm GPUs are correctly listed as such instead of as CUDA is still valuable to have.

In general, I consider issues like this one to be upstream issues. The correct fix here is for bnb to add ROCm support, not for us to riddle our code with shims.

PR #169, opened 2026-02-16
fix: report VRAM usage across all GPUs in multi-GPU setups

### LINKED ISSUE #168: VRAM usage reports only 1 GPU in multi-GPU setups
On multi-GPU setups, incorrect VRAM usage is being reported. I found that only one GPU's memory is shown instead of the total across all devices.

Example output on dual RTX 5060 Ti (16GB each):

  Resident system RAM: 2.43 GB
  Allocated GPU VRAM: 3.74 GB
  Reserved GPU VRAM: 3.76 GB

The allocated/reserved values only reflect GPU 0. The actual usage across both GPUs is roughly double.

cpagac:
Fixes #168

## Summary
  - `torch.cuda.memory_allocated()` and `torch.cuda.memory_reserved()` without a device argument only report memory for GPU 0, resulting in incorrect VRAM usage on multi-GPU setups                    
  - Sum memory across all devices using `torch.cuda.device_count()` for correct totals                                                                                                                  
  - Add total GPU VRAM reporting via `torch.cuda.mem_get_info()`                                                                                                                                        
  - Apply the same fix to XPU memory reporting

  ## Test plan
  - Tested on dual NVIDIA RTX 5060 Ti (16GB each)                                                                                                                                                       
  - Verified reported VRAM matches expected usage across both GPUs                                                                                                                                      
  - Confirmed total GPU VRAM correctly reports ~32GB combined

  **Before (original, only reports GPU 0):**                                                                                                                                                            

heretic-Qwen3-14B_original

**After (sums across all GPUs):**

heretic-Qwen3-14B_patch

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -174,9 +174,15 @@ def run():
     # Adapted from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/env.py
     if torch.cuda.is_available():
         count = torch.cuda.device_count()
-        print(f"Detected [bold]{count}[/] CUDA device(s):")
+        total_vram = sum(torch.cuda.mem_get_info(i)[1] for i in range(count))
+        print(
+            f"Detected [bold]{count}[/] CUDA device(s) ({total_vram / (1024**3):.2f} GB total VRAM):"
+        )
         for i in range(count):
-            print(f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/]")
+            vram = torch.cuda.mem_get_info(i)[1] / (1024**3)
+            print(
+                f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/] ({vram:.2f} GB)"
+            )
     elif is_xpu_available():
         count = torch.xpu.device_count()
         print(f"Detected [bold]{count}[/] XPU device(s):")
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -38,11 +38,17 @@ def print_memory_usage():
     p("Resident system RAM", Process().memory_info().rss)
 
     if torch.cuda.is_available():
-        p("Allocated GPU VRAM", torch.cuda.memory_allocated())
-        p("Reserved GPU VRAM", torch.cuda.memory_reserved())
+        count = torch.cuda.device_count()
+        allocated = sum(torch.cuda.memory_allocated(device) for device in range(count))
+        reserved = sum(torch.cuda.memory_reserved(device) for device in range(count))
+        p("Allocated GPU VRAM", allocated)
+        p("Reserved GPU VRAM", reserved)
     elif is_xpu_available():
-        p("Allocated XPU memory", torch.xpu.memory_allocated())
-        p("Reserved XPU memory", torch.xpu.memory_reserved())
+        count = torch.xpu.device_count()
+        allocated = sum(torch.xpu.memory_allocated(device) for device in range(count))
+        reserved = sum(torch.xpu.memory_reserved(device) for device in range(count))
+        p("Allocated XPU memory", allocated)
+        p("Reserved XPU memory", reserved)
     elif torch.backends.mps.is_available():
         p("Allocated MPS memory", torch.mps.current_allocated_memory())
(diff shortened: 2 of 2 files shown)


src/heretic/utils.py
```diff
@@ -38,11 +38,18 @@ def p(label: str, size_in_bytes: int):
...
     if torch.cuda.is_available():
-        p("Allocated GPU VRAM", torch.cuda.memory_allocated())
-        p("Reserved GPU VRAM", torch.cuda.memory_reserved())
+        count = torch.cuda.device_count()
+        allocated = sum(torch.cuda.memory_allocated(device) for device in range(count))
+        reserved = sum(torch.cuda.memory_reserved(device) for device in range(count))
+        total = sum(torch.cuda.mem_get_info(device)[1] for device in range(count))
+        p("Allocated GPU VRAM", allocated)
+        p("Reserved GPU VRAM", reserved)
+        p("Total GPU VRAM", total)
```
p-e-w: I don't think we should display the total on every trial. Virtually everyone who is interested in that information will already know the amount of VRAM they have, and it cannot change between trials, so this is just printing the same number hundreds of times.

src/heretic/utils.py
```diff
@@ -38,11 +38,18 @@ def p(label: str, size_in_bytes: int):
...
     elif is_xpu_available():
-        p("Allocated XPU memory", torch.xpu.memory_allocated())
-        p("Reserved XPU memory", torch.xpu.memory_reserved())
+        allocated = sum(torch.xpu.memory_allocated(device) for device in range(torch.xpu.device_count()))
+        reserved = sum(torch.xpu.memory_reserved(device) for device in range(torch.xpu.device_count()))
+        p("Allocated XPU memory", allocated)
+        p("Reserved XPU memory", reserved)
```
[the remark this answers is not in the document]
p-e-w: I agree with Gemini about the count, but regarding the total I don't think we should have that in either block, as mentioned above.

src/heretic/utils.py
p-e-w: If anywhere, printing the total VRAM should happen when we list the available devices on program start.

src/heretic/main.py
```diff
@@ -174,7 +174,10 @@ def run():
     # Adapted from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/env.py
     if torch.cuda.is_available():
         count = torch.cuda.device_count()
-        print(f"Detected [bold]{count}[/] CUDA device(s):")
+        total_vram = sum(torch.cuda.mem_get_info(i)[1] for i in range(count))
+        print(
+            f"Detected [bold]{count}[/] CUDA device(s) ({total_vram / (1024**3):.2f} GB total VRAM):"
+        )
         for i in range(count):
             print(f"* GPU {i}: [bold]{torch.cuda.get_device_name(i)}[/]")
```
p-e-w: Maybe add the device VRAM in parentheses here after the device name?

cpagac:
Thanks for the feedback so quickly! I've updated the PR as such.
                                                  
  - Removed total VRAM from print_memory_usage() so it's no longer repeated every trial                                                                                                                      
  - Kept total VRAM in the CUDA device listing at startup
  - Added per-device VRAM in parentheses next to each GPU name as suggested                                                                                                                                  
  - Added count variable for XPU to match the CUDA block                                                                                                                                                     

  Startup output now looks like:
  Detected 2 CUDA device(s) (31.86 GB total VRAM):
  * GPU 0: NVIDIA RTX 5060 Ti (15.93 GB)
  * GPU 1: NVIDIA RTX 5060 Ti (15.93 GB)

src/heretic/utils.py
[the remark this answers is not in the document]
cpagac: Done — added count for XPU and removed total from both blocks.

src/heretic/utils.py
cpagac: Done — total VRAM is now displayed in the startup device listing header.

src/heretic/main.py
cpagac: Done — each device now shows its VRAM in parentheses after the name.

p-e-w:
Merged! Thank you for the PR :)

cpagac:
Ofc! Glad to support the project! :)

PR #176, opened 2026-02-18
fix: update dependencies to address security vulnerabilities (CVEs)

RinZ27:
Nudged aiohttp, pillow, and urllib3 to their latest patch versions in the lockfile. This handles some recent security issues, specifically the decompression bomb safeguards in urllib3 (CVE-2026-21441) and aiohttp (CVE-2025-69223), plus an OOB write in Pillow (CVE-2026-25990). I checked the diffs and these are just minor version bumps, so there's no risk to the current logic.


p-e-w:
I frequently re-generate `uv.lock` automatically. Manually editing it is not acceptable.

None of these are direct dependencies of Heretic. Updating those versions must be done in the packages we depend on.

Furthermore, `uv.lock` is not used by Pip, so this doesn't really solve anything anyway.

RinZ27:
@p-e-w, point taken on the `uv.lock` generation and the project's reliance on Pip. I noticed those upstream vulnerabilities and wanted to address them, but I definitely picked the wrong mechanism for this environment. Thanks for the clarification!

p-e-w:
No problem, thanks anyway!

PR #180, opened 2026-02-18
fix: Don't display negative KL divergences

### LINKED ISSUE #141: Don't display negative KL divergences
The KLD is nonnegative by definition, but the computation can sometimes return negative zero floating point. This should be displayed as `0`, not as `-0.0000`.

bintocher:
## Summary

This PR fixes the display of negative KL divergence values in the evaluation output. KL divergence is a mathematical measure that cannot be negative, so any negative values indicate numerical instability or precision errors. The fix clamps KL divergence values to a minimum of zero before displaying them.

## Issue

Fixes #141

## Changes

- Added `max(0.0, ...)` constraint to KL divergence calculations in `evaluator.py`
- Ensures displayed KL divergence values are always non-negative
- Improves output clarity by hiding numerical artifacts

## Testing

- Verified that negative KL divergence values no longer appear in evaluation output
- Confirmed that valid positive values are displayed unchanged
- Tested with various model comparisons to ensure numerical stability

### DIFF
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -95,15 +95,12 @@ class Evaluator:
     def get_score(self) -> tuple[tuple[float, float], float, int]:
         print("  * Obtaining first-token probability distributions...")
         logprobs = self.model.get_logprobs_batched(self.good_prompts)
-        kl_divergence = max(
-            0.0,
-            F.kl_div(
-                logprobs,
-                self.base_logprobs,
-                reduction="batchmean",
-                log_target=True,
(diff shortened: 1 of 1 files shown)


p-e-w:
See #160, which is a more principled approach that tries to address the root cause. I'll wait a bit before deciding whether to merge this PR on top of that.

bintocher:
Thanks for the note! Makes sense — happy to close this if #160 covers it fully.

p-e-w:
The underlying issue has been fixed in #194.

PR #182, opened 2026-02-19
fix: handle FP8 model weights in LoRA adapters and merge

shifusen329:
## Summary
- Models distributed in FP8 (e.g. MiniMax-M2.5) crash during inference because `torch.addmm` has no FP8 (`Float8_e4m3fn`) kernel, causing `NotImplementedError` in PEFT LoRA forward passes
- Cast LoRA adapter weights to bfloat16 after initialization in `_apply_lora()` so adapter matmuls use a supported dtype
- Upcast FP8 base weights to bfloat16 before `merge_and_unload()` in `get_merged_model()`, then cast back, to avoid unsupported in-place addition during merge

## Test plan
- [ ] Load an FP8 model (e.g. `MiniMaxAI/MiniMax-M2.5`) and verify inference completes without `NotImplementedError`
- [ ] Verify saving a merged model works for FP8 models
- [ ] Verify no regression for non-FP8 models (bfloat16, float16, BNB 4-bit)

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -717,9 +717,9 @@ class Model:
         logits = cast(tuple[FloatTensor], outputs.scores)[0]
 
         # The returned tensor has shape (prompt, token).
-        # Upcast to float32 before log_softmax to avoid -inf from underflow
-        # in low-precision dtypes (FP8, float16, bfloat16).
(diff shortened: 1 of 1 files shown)


p-e-w:
This seems to be quite similar to #151, which is also concerned with FP8 support.

PR #183, opened 2026-02-19
fix: use repr() for exception messages to always include exception type (#146)

### LINKED ISSUE #146: Printing exceptions outputs empty message if the exception has no message attached, complicating debugging

cluster2600:
### Problem

When an exception has no message attached (e.g. `raise RuntimeError()`), converting it to a string via f-string interpolation yields an empty string. This produces unhelpful output like:

```
Failed ()
Error: 
```

...making debugging significantly harder because the user has no indication of what type of error occurred.

### Fix

Use `repr()` (via the `!r` format specifier) for exception printing in three locations:
- `src/heretic/main.py` — batch size test failure reporting
- `src/heretic/main.py` — interactive chat error reporting
- `src/heretic/model.py` — model loading failure reporting

`repr()` always includes the exception class name, regardless of whether a message is attached:

| Scenario | Before | After |
|----------|--------|-------|
| Exception with message | `Failed (something went wrong)` | `Failed (RuntimeError('something went wrong'))` |
| Exception without message | `Failed ()` | `Failed (RuntimeError())` |

### Impact

Minimal: purely a diagnostic/logging change. No functional behavior is altered.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -356,7 +356,7 @@ def run():
                     # We cannot recover from this.
                     raise
 
-                print(f"[red]Failed[/] ({error!r})")
+                print(f"[red]Failed[/] ({type(error).__name__}: {error})")
                 break
 
             response_lengths = [
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -131,7 +131,7 @@ class Model:
             except Exception as error:
                 self.model = None  # ty:ignore[invalid-assignment]
                 empty_cache()
-                print(f"[red]Failed[/] ({error!r})")
+                print(f"[red]Failed[/] ({type(error).__name__}: {error})")
                 continue
 
             if settings.quantization == QuantizationMethod.BNB_4BIT:


p-e-w:
Thanks for the PR.

I'm not sure this solves the problem, though. `Failed (RuntimeError())` isn't really more helpful than `Failed ()`.

cluster2600:
Good point — `RuntimeError()` without a message isn't very helpful.

I updated the output to include both the exception type and the message (when present), e.g. `Failed (RuntimeError: probability tensor contains inf/nan)`.

If the exception message is empty, this still improves the output to `Failed (RuntimeError: )` (type preserved).

Pushed in commit 0ecacf0.

p-e-w:
But that was already the case before, see your original output.

My point is that if there is no message, the error type often contains no useful information for debugging. And if there is a message, the error type is output noise.

cluster2600:
Fair point — you're right that the type alone is noise when there's a message, and adds nothing new when there isn't one. Thanks for the patient explanation. I'll close this; if a better approach comes up (e.g. surfacing the full traceback or chain of causes in debug mode) I'll revisit with a more useful PR.

PR #187, opened 2026-02-22
feat: add Qwen3.5 MoE hybrid layer support

farolone:
## Summary

Qwen3.5 MoE uses GatedDeltaNet (linear attention) on some layers instead of standard self-attention. This causes abliteration to fail because `self_attn.o_proj` doesn't exist on those layers.

- Wrap `self_attn.o_proj` access in `suppress(Exception)` and add `linear_attn.out_proj` as alternative attention out-projection for GatedDeltaNet layers
- Scan all layers in `get_abliterable_components()` instead of only layer 0, since hybrid models have different components on different layers
- Derive LoRA `target_modules` from actual `named_modules()` instead of splitting component keys, which fails when module names differ across layers (e.g. `o_proj` vs `out_proj`)

## Test results

Successfully abliterated **Qwen3.5-397B-A17B** (7× H200 SXM):
- **Trial 104** selected from 200-trial Pareto front
- **Refusals: 7/100**, KL divergence: **0.2676**
- Parameters: `direction_index=52.45, attn.o_proj.max_weight=1.34, max_weight_position=36.86, min_weight=0.66, min_weight_distance=15.21`

The resulting model passes both censorship removal and quality preservation tests.

## Relation to #43

This PR is complementary to #43 (hybrid layer support for LFM/Mamba/Conv). While #43 adds broad architecture support across `get_layer_matrices()`, this PR focuses specifically on Qwen3.5 MoE's GatedDeltaNet layers and fixes `get_layer_modules()` + LoRA targeting which #43 does not touch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #193, opened 2026-02-26
feat: Add Qwen3.5 MoE Hybrid Layer Support

milobeans:


p-e-w:
How does this compare to #187?

milobeans:
Mostly overlaps with #187. Differences in #193 are: 
 (1) I also handle Qwen3.5 MoE’s shared expert path (mlp.shared_expert.down_proj)
 (2) I collect abliterable components as a union across layers and preserve order
 (3) I keep LoRA target selection as deduped leaf names (dict.fromkeys) rather than the
  named_modules()-based target resolution in #187.

src/heretic/model.py
```diff
@@ -340,9 +341,17 @@ def try_add(component: str, module: Any):
...
-        try_add("attn.o_proj", layer.self_attn.o_proj)  # ty:ignore[possibly-missing-attribute]
+        # Attention out-projection. Most models expose `self_attn.o_proj`,
+        # but hybrid/linear-attention models (e.g. Qwen3.5-MoE) may expose
+        # `linear_attn.out_proj` instead.
+        has_attn_out_proj = False
+        with suppress(Exception):
+            try_add("attn.o_proj", layer.self_attn.o_proj)  # ty:ignore[possibly-missing-attribute]
+            has_attn_out_proj = True
+        with suppress(Exception):
+            try_add("attn.out_proj", layer.linear_attn.out_proj)  # ty:ignore[possibly-missing-attribute]
```
p-e-w, line 352: This places linear attention modules under a separate component, which adds 4 new parameters to the optimization task, increasing the number of trials needed to find a good set of parameters.

Considering that the semantics of linear and traditional attention are similar, do you think it makes sense to use the `attn.o_proj` component for both?

src/heretic/model.py
```diff
@@ -374,7 +387,14 @@ def try_add(component: str, module: Any):
...
     def get_abliterable_components(self) -> list[str]:
-        return list(self.get_layer_modules(0).keys())
+        # Hybrid architectures can expose different component names across layers
+        # (e.g. `attn.o_proj` and `attn.out_proj`), so collect a union.
+        components = []
+        for layer_index in range(len(self.get_layers())):
+            for component in self.get_layer_modules(layer_index).keys():
+                if component not in components:
+                    components.append(component)
+        return components
```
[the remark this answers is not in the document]
p-e-w: Screw this suggestion, your approach is clearer and performance is irrelevant here.

src/heretic/model.py
```diff
@@ -353,6 +362,10 @@ def try_add(component: str, module: Any):
             for expert in layer.mlp.experts:  # ty:ignore[possibly-missing-attribute, not-iterable]
                 try_add("mlp.down_proj", expert.down_proj)  # ty:ignore[possibly-missing-attribute]
 
+        # Qwen3.5-MoE shared expert.
+        with suppress(Exception):
+            try_add("mlp.down_proj", layer.mlp.shared_expert.down_proj)  # ty:ignore[possibly-missing-attribute]
```
p-e-w, line 367: Hmm. Looking at their source code, we see this:

https://github.com/huggingface/transformers/blob/11b1906d5c0dae39c13270e47cc02c4cde70e548/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L878-L889

Note that they multiply the output of the shared expert with the output of a per-component sigmoid function before adding it to the residual stream. This means that `shared_expert.down_proj` doesn't actually write to the residual stream, and ablating it based on residuals would introduce distortions rather than suppressing the associated subspace.

p-e-w:
> (3) I keep LoRA target selection as deduped leaf names (dict.fromkeys) rather than the named_modules()-based target resolution in https://github.com/p-e-w/heretic/pull/187.

Are there any situations where you think your (much simpler) approach can fail?

Sehyo:
This approach, as well as #187 neither supports fused expert abliteration which is critical for good results for Qwen 3.5 MoE.
I've uploaded my own PR here: https://github.com/p-e-w/heretic/pull/207

It covers:
- Hybrid layer detection
- Shared Expert (missing in #187)
- Fused expert abliteration (missing in both #187 and #193)
- Separate attn component keys (o_proj + out_proj) (missing in #187).
- I also set enable_thinking=False which is not in #193 and #187 (currently Heretic does inject <think></think> however for Qwen3.5 case those tokens will never exist in the first place if thinking is disabled anyway, so this approach should lead to slightly more correctness.

p-e-w:
I have decided to merge #187 instead of this PR because the question regarding the shared expert module remained unresolved. Thank you anyway for this PR!

PR #196, opened 2026-02-26
[PoC]: Multi-directional refusal supression with Self-Organizing Maps

### LINKED ISSUE #140: SOM Directions are Better than One: Multi-Directional Refusal Suppression in Language Models
"SOM Directions are Better than One: Multi-Directional Refusal Suppression in Language Models" is an another paper on the matter of abliteration.

It provides a deeper analysis of that directions to apply to the model for refusal removal using self-organizing maps, with less damage to the downstream model as opposed to a single direction ablation.

Code, paper - all available:

https://arxiv.org/abs/2511.08379v2

https://github.com/pralab/som-refusal-directions

I think it may be a great method to have in this repository.

kabachuha:
So, I finally implemented multi-directional abliteration in heretic, porting the Self-Organizing Maps refusal direction determination from #140.

Integrating Multidirections and SOMs turned out to be much easier than I thought, thanks to a local GLM Air derestricted who helped with the code.

Initially I thought it would break everything and I would fail, but no, it worked perfectly with multiple directions optimization. (In my experiment I used 4 directions)

Here is the comparison to the readme reference

| Model | Refusals for "harmful" prompts | KL divergence from original model for "harmless" prompts |
| :--- | ---: | ---: |
| [google/gemma-3-12b-it](https://huggingface.co/google/gemma-3-12b-it) (original) | 97/100 | 0 *(by definition)* |
| [mlabonne/gemma-3-12b-it-abliterated-v2](https://huggingface.co/mlabonne/gemma-3-12b-it-abliterated-v2) | 3/100 | 1.04 |
| [huihui-ai/gemma-3-12b-it-abliterated](https://huggingface.co/huihui-ai/gemma-3-12b-it-abliterated) | 3/100 | 0.45 |
| [p-e-w/gemma-3-12b-it-heretic](https://huggingface.co/p-e-w/gemma-3-12b-it-heretic) | 3/100 | 0.16 |
| **gemma-3-12b-it-SOM-heretic (mine)** | **3/100** | **0.08** |

Closes #140. I know it will be better to have as a plugin, so this PR serves as a future reference.

From testing (chat with the model), can confirm that it writes NSFW and edgy content just fine. I'm going to sleep and I can upload the model to huggingface if you will ask for it.

Also, directions visualization would be fun to have.

Here is how the parameter listing looks like:

image

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -244,6 +244,11 @@ class Settings(BaseSettings):
         default=4, description="Number of top neurons to use for multidirectional SOM."
     )
 
+    som_use_win_map: bool = Field(
+        default=False,
+        description="Use win map for SOM neurons retrieval.",
+    )
+
     n_trials: int = Field(
         default=200,
         description="Number of abliteration trials to run during optimization.",
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -244,11 +244,6 @@ class Settings(BaseSettings):
         default=4, description="Number of top neurons to use for multidirectional SOM."
     )
 
-    som_use_win_map: bool = Field(
-        default=False,
-        description="Use win map for SOM neurons retrieval.",
-    )
-
     n_trials: int = Field(
         default=200,
         description="Number of abliteration trials to run during optimization.",
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -215,6 +215,35 @@ class Settings(BaseSettings):
         ),
     )
 
+    multidirectional_som: bool = Field(
+        default=False,
+        description="Use multidirectional Self-Organising Maps. Requires 'minisom' package to be installed.",
+    )
+
+    som_x: int = Field(
+        default=4, description="Number of SOM neurons in the x-axis."
+    )
+
+    som_y: int = Field(
+        default=4, description="Number of SOM neurons in the y-axis."
+    )
+
+    som_iterations: int = Field(
+        default=10000, description="Number of SOM training iterations."
+    )
+
+    som_lr: float = Field(
+        default=0.01, description="SOM learning rate."
+    )
+
+    som_sigma: float = Field(
+        default=0.5, description="SOM neighborhood radius."
+    )
+
+    som_k: int = Field(
+        default=4, description="Number of top neurons to use for multidirectional SOM."
+    )
+
     n_trials: int = Field(
         default=200,
         description="Number of abliteration trials to run during optimization.",
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -395,7 +395,7 @@ class Model:
                         Each AbliterationParameters instance must have max_weights and min_weights
                         as lists of length equal to the number of directions.
         """
-        num_layers, num_directions, _ = refusal_directions.shape
+        _, num_directions, _ = refusal_directions.shape
 
         if direction_index is not None:
             # If a specific direction is requested, interpolate across the direction dimension.
@@ -422,7 +422,7 @@ class Model:
             layer_refusal_directions = refusal_directions
 
         # Now, iterate through each layer to apply the ablation.
-        for layer_index in range(num_layers):
+        for layer_index in range(len(self.get_layers())):
             for component, modules in self.get_layer_modules(layer_index).items():
                 params = parameters[component]
 
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -46,9 +46,9 @@ def get_model_class(
 
 @dataclass
 class AbliterationParameters:
-    max_weight: float
+    max_weights: list[float]
     max_weight_position: float
-    min_weight: float
+    min_weights: list[float]
     min_weight_distance: float
 
 
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -501,7 +501,24 @@ class Model:
 
                     # Now, apply the combined delta to W based on the row normalization setting.
                     if self.settings.row_normalization == RowNormalization.PRE:
-                        total_delta_W = W_row_norms * total_delta_W
+                        # Use low-rank SVD to decompose the total_delta_W into LoRA matrices.
+                        r = self.peft_config.r
+                        # Perform SVD on the total_delta_W matrix
+                        U, S, Vh = torch.svd_lowrank(total_delta_W, q=min(2 * r + 4, min(total_delta_W.shape)), niter=6)
+
+                        # Truncate to rank 'r'
+                        U = U[:, :r]
+                        S = S[:r]
+                        Vh = Vh[:, :r].T # Vh is (r, in_features)
+
+                        # Split the singular values between lora_B and lora_A
+                        sqrt_S = torch.sqrt(S)
+                        lora_B = U @ torch.diag(sqrt_S)  # Shape: (out_features, r)
+                        lora_A = torch.diag(sqrt_S) @ Vh  # Shape: (r, in_features)
+
+                        # Apply PRE normalization: scale lora_B by the original row norms.
+                        # lora_B already has shape (out_features, r), W_row_norms is (out_features, 1)
+                        lora_B = W_row_norms * lora_B
                     elif self.settings.row_normalization == RowNormalization.FULL:
                         W = W + total_delta_W
                         W = F.normalize(W, p=2, dim=1)
@@ -516,17 +533,26 @@ class Model:
                         sqrt_S = torch.sqrt(S)
                         lora_B = U @ torch.diag(sqrt_S)
                         lora_A = torch.diag(sqrt_S) @ Vh
-                        # Assign the SVD result to the LoRA weights
-                        weight_A = cast(Tensor, module.lora_A["default"].weight)
-                        weight_B = cast(Tensor, module.lora_B["default"].weight)
-                        weight_A.data = lora_A.to(weight_A.dtype)
-                        weight_B.data = lora_B.to(weight_B.dtype)
                     else: # RowNormalization.NONE or others
-                        W = W + total_delta_W
+                        # In NONE mode, the delta is simply total_delta_W.
+                        # We decompose this delta directly using SVD.
+                        r = self.peft_config.r
+                        U, S, Vh = torch.svd_lowrank(total_delta_W, q=min(2 * r + 4, min(total_delta_W.shape)), niter=6)
 
-                        # Assign the new weight matrix back to the base layer.
-                        # Reshape W back to its original 2D shape.
-                        module.base_layer.weight.data = W.view_as(module.base_layer.weight).to(module.base_layer.weight.dtype)
+                        U = U[:, :r]
+                        S = S[:r]
+                        Vh = Vh[:, :r].T
+
+                        sqrt_S = torch.sqrt(S)
+                        lora_B = U @ torch.diag(sqrt_S)
+                        lora_A = torch.diag(sqrt_S) @ Vh
+
+                        
+                    # Assign the SVD result to the LoRA weights
+                    weight_A = cast(Tensor, module.lora_A["default"].weight)
+                    weight_B = cast(Tensor, module.lora_B["default"].weight)
+                    weight_A.data = lora_A.to(weight_A.dtype)
+                    weight_B.data = lora_B.to(weight_B.dtype)
 
     def generate(
         self,
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -400,21 +400,21 @@ class Model:
         if direction_index is not None:
             # If a specific direction is requested, interpolate across the direction dimension.
             # For example, if direction_index=0.5, it will blend direction 0 and 1.
-            weight, index = math.modf(direction_index)
-            # Clamp index to be within the valid range [0, num_directions - 1]
-            idx1 = int(index) % num_directions
-            idx2 = (idx1 + 1) % num_directions
-            # Interpolate between the two chosen directions. The result has shape (layers, hidden_dim)
-            interpolated_directions = F.normalize(
-                refusal_directions[:, idx1].lerp(
-                    refusal_directions[:, idx2],
+            weight, index = math.modf(direction_index + 1)
+            # Clamp index to be within the valid range [0, num_layers - 1]
+            idx1 = int(index)
+            idx2 = (idx1 + 1)
+            # Interpolate between the two chosen directions. The result has shape (n_directions, hidden_dim)
+            interpolated_directions = torch.stack([F.normalize(
+                r_dir[idx1].lerp(
+                    r_dir[idx2],
                     weight,
                 ),
                 p=2,
-                dim=1,
-            )
+                dim=0,
+            ) for r_dir in refusal_directions.permute(1, 0, 2)], dim=0)
             # We will use this single "blended" set of directions for all layers.
-            # Shape: (layers, hidden_dim)
+            # Shape: (n_directions, hidden_dim)
             layer_refusal_directions = interpolated_directions
         else:
             # If no specific direction is given, we use all directions for each layer.
@@ -450,7 +450,10 @@ class Model:
 
                 # Get the refusal directions for the current layer.
                 # Shape: (num_directions, hidden_dim)
-                current_layer_directions = layer_refusal_directions[layer_index]
+                if direction_index is None:
+                    current_layer_directions = layer_refusal_directions[layer_index+1]
+                else:
+                    current_layer_directions = layer_refusal_directions
 
                 for module in modules:
                     module = cast(Linear, module)
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -378,146 +378,152 @@ class Model:
 
     def abliterate(
         self,
-        refusal_directions: Tensor,
+        refusal_directions: torch.Tensor, # Shape: (layers, num_directions, hidden_dim)
         direction_index: float | None,
         parameters: dict[str, AbliterationParameters],
     ):
-        if direction_index is None:
-            refusal_direction = None
-        else:
-            # The index must be shifted by 1 because the first element
-            # of refusal_directions is the direction for the embeddings.
-            weight, index = math.modf(direction_index + 1)
-            refusal_direction = F.normalize(
-                refusal_directions[int(index)].lerp(
-                    refusal_directions[int(index) + 1],
+        """
+        Abliterates the model using multidirectional refusal vectors.
+
+        Args:
+            refusal_directions: A 3D tensor where the first dimension corresponds to layers,
+                                the second to the number of SOM-derived directions, and the third
+                                to the hidden dimension.
+            direction_index: An optional float to interpolate between two specific refusal directions
+                            (across all layers). If None, all directions are used.
+            parameters: A dictionary mapping component names to AbliterationParameters.
+                        Each AbliterationParameters instance must have max_weights and min_weights
+                        as lists of length equal to the number of directions.
+        """
+        num_layers, num_directions, _ = refusal_directions.shape
+
+        if direction_index is not None:
+            # If a specific direction is requested, interpolate across the direction dimension.
+            # For example, if direction_index=0.5, it will blend direction 0 and 1.
+            weight, index = math.modf(direction_index)
+            # Clamp index to be within the valid range [0, num_directions - 1]
+            idx1 = int(index) % num_directions
+            idx2 = (idx1 + 1) % num_directions
+            # Interpolate between the two chosen directions. The result has shape (layers, hidden_dim)
+            interpolated_directions = F.normalize(
+                refusal_directions[:, idx1].lerp(
+                    refusal_directions[:, idx2],
                     weight,
                 ),
                 p=2,
-                dim=0,
+                dim=1,
             )
+            # We will use this single "blended" set of directions for all layers.
+            # Shape: (layers, hidden_dim)
+            layer_refusal_directions = interpolated_directions
+        else:
+            # If no specific direction is given, we use all directions for each layer.
+            # Shape is already (layers, num_directions, hidden_dim)
+            layer_refusal_directions = refusal_directions
 
-        # Note that some implementations of abliteration also orthogonalize
-        # the embedding matrix, but it's unclear if that has any benefits.
-        for layer_index in range(len(self.get_layers())):
+        # Now, iterate through each layer to apply the ablation.
+        for layer_index in range(num_layers):
             for component, modules in self.get_layer_modules(layer_index).items():
                 params = parameters[component]
 
-                # Type inference fails here for some reason.
-                distance = cast(float, abs(layer_index - params.max_weight_position))
+                # Ensure the number of weights matches the number of directions
+                if len(params.max_weights) != num_directions or len(params.min_weights) != num_directions:
+                    raise ValueError(
+                        f"Mismatch in number of directions and weights for component '{component}'. "
+                        f"Found {num_directions} directions but {len(params.max_weights)} max_weights."
+                    )
 
-                # Don't orthogonalize layers that are more than
-                # min_weight_distance away from max_weight_position.
+                # Get the layer-specific position scaling factor.
+                # This factor is the same for all directions in this layer.
+                distance = cast(float, abs(layer_index - params.max_weight_position))
                 if distance > params.min_weight_distance:
+                    # If this layer is too far from the optimal position, skip it entirely.
                     continue
 
-                # Interpolate linearly between max_weight and min_weight
-                # over min_weight_distance.
-                weight = params.max_weight + (distance / params.min_weight_distance) * (
-                    params.min_weight - params.max_weight
+                # Calculate the base weight scaling factor for this layer.
+                # This factor will be applied to all directions in this layer.
+                layer_base_weight = params.max_weights[0] + (distance / params.min_weight_distance) * (
+                    params.min_weights[0] - params.max_weights[0]
                 )
+                # Note: In the original single-direction logic, this `layer_base_weight` was the
+                # final weight. Now, it's a base factor that can be further modified per direction.
 
-                if refusal_direction is None:
-                    # The index must be shifted by 1 because the first element
-                    # of refusal_directions is the direction for the embeddings.
-                    layer_refusal_direction = refusal_directions[layer_index + 1]
-                else:
-                    layer_refusal_direction = refusal_direction
+                # Get the refusal directions for the current layer.
+                # Shape: (num_directions, hidden_dim)
+                current_layer_directions = layer_refusal_directions[layer_index]
 
                 for module in modules:
-                    # FIXME: This cast is potentially invalid, because the program logic
-                    #        does not guarantee that the module is of type Linear, and in fact
-                    #        the retrieved modules might not conform to the interface assumed
-                    #        below (though they do in practice). However, this is difficult
-                    #        to fix cleanly, because get_layer_modules is called twice on
-                    #        different model configurations, and PEFT employs different
-                    #        module types depending on the chosen quantization.
                     module = cast(Linear, module)
-
-                    # LoRA abliteration: delta W = -lambda * v * (v^T W)
-                    # lora_B = -lambda * v
-                    # lora_A = v^T W
-
-                    # Use the FP32 refusal direction directly (no downcast/upcast)
-                    # and move to the correct device.
-                    v = layer_refusal_direction.to(module.weight.device)
-
-                    # Get W (dequantize if necessary).
-                    #
-                    # FIXME: This cast is valid only under the assumption that the original
-                    #        module wrapped by the LoRA adapter has a weight attribute.
-                    #        See the comment above for why this is currently not guaranteed.
+                    # Get the weight matrix W (dequantized if necessary) in FP32.
                     base_weight = cast(Tensor, module.base_layer.weight)
                     quant_state = getattr(base_weight, "quant_state", None)
-
                     if quant_state is None:
                         W = base_weight.to(torch.float32)
                     else:
-                        # 4-bit quantization.
-                        # This cast is always valid. Type inference fails here because the
-                        # bnb.functional module is not found by ty for some reason.
-                        W = cast(
-                            Tensor,
-                            bnb.functional.dequantize_4bit(  # ty:ignore[possibly-missing-attribute]
-                                base_weight.data,
-                                quant_state,
-                            ).to(torch.float32),
-                        )
+                        W = cast(Tensor, bnb.functional.dequantize_4bit(base_weight.data, quant_state)).to(torch.float32)
 
-                    # Flatten weight matrix to (out_features, in_features).
-                    W = W.view(W.shape[0], -1)
+                    W = W.view(W.shape[0], -1) # Flatten to (out_features, in_features)
 
                     if self.settings.row_normalization != RowNormalization.NONE:
-                        # Keep a reference to the original weight matrix so we can subtract it later.
                         W_org = W
-                        # Get the row norms.
                         W_row_norms = LA.vector_norm(W, dim=1, keepdim=True)
-                        # Normalize the weight matrix along the rows.
                         W = F.normalize(W, p=2, dim=1)
 
-                    # Calculate lora_A = v^T W
-                    # v is (d_out,), W is (d_out, d_in)
-                    # v @ W -> (d_in,)
-                    lora_A = (v @ W).view(1, -1)
+                    # --- The core change is here ---
+                    # We will calculate a delta for each direction and sum them up.
+                    total_delta_W = torch.zeros_like(W, device=W.device)
 
-                    # Calculate lora_B = -weight * v
-                    # v is (d_out,)
-                    lora_B = (-weight * v).view(-1, 1)
+                    for i in range(num_directions):
+                        # Get the specific refusal direction for this component.
+                        v = current_layer_directions[i].to(module.weight.device)
 
+                        # Get the optimized weight for this specific direction (i).
+                        # It's a combination of the layer's base weight and the direction's specific weight.
+                        direction_specific_weight = layer_base_weight * (params.max_weights[i] / params.max_weights[0] if params.max_weights[0] != 0 else 1.0)
+
+                        # LoRA abliteration: delta W = -lambda * v * (v^T W)
+                        # lora_B = -lambda * v
+                        # lora_A = v^T W
+                        v = v # Shape: (hidden_dim,)
+                        W_matrix = W # Shape: (hidden_dim, in_features)
+
+                        # Calculate lora_A = v^T W -> (in_features,)
+                        lora_A = (v @ W_matrix).view(1, -1)
+                        # Calculate lora_B = -direction_specific_weight * v -> (hidden_dim, 1)
+                        lora_B = (-direction_specific_weight * v).view(-1, 1)
+
+                        # The delta for this direction is lora_B @ lora_A
+                        delta_W = lora_B @ lora_A
+                        total_delta_W += delta_W
+
+                    # Now, apply the combined delta to W based on the row normalization setting.
                     if self.settings.row_normalization == RowNormalization.PRE:
-                        # Make the LoRA adapter apply to the original weight matrix.
-                        lora_B = W_row_norms * lora_B
+                        total_delta_W = W_row_norms * total_delta_W
                     elif self.settings.row_normalization == RowNormalization.FULL:
-                        # Approximates https://huggingface.co/blog/grimjim/norm-preserving-biprojected-abliteration
-                        W = W + lora_B @ lora_A
-                        # Normalize the adjusted weight matrix along the rows.
+                        W = W + total_delta_W
                         W = F.normalize(W, p=2, dim=1)
-                        # Restore the original row norms of the weight matrix.
                         W = W * W_row_norms
-                        # Subtract the original matrix to turn W into a delta.
                         W = W - W_org
-                        # Use a low-rank SVD to get an approximation of the matrix.
+                        # Use low-rank SVD for the full delta
                         r = self.peft_config.r
                         U, S, Vh = torch.svd_lowrank(W, q=2 * r + 4, niter=6)
-                        # Truncate it to the part we want to store in the LoRA adapter.
-                        # Note: svd_lowrank actually returns V, so transpose it to get Vh.
                         U = U[:, :r]
                         S = S[:r]
                         Vh = Vh[:, :r].T
-                        # Transfer it into the LoRA adapter components. Split the singular values
-                        # evenly between the two components to keep their norms balanced and avoid
-                        # potential issues with numerical stability.
                         sqrt_S = torch.sqrt(S)
                         lora_B = U @ torch.diag(sqrt_S)
                         lora_A = torch.diag(sqrt_S) @ Vh
+                        # Assign the SVD result to the LoRA weights
+                        weight_A = cast(Tensor, module.lora_A["default"].weight)
+                        weight_B = cast(Tensor, module.lora_B["default"].weight)
+                        weight_A.data = lora_A.to(weight_A.dtype)
+                        weight_B.data = lora_B.to(weight_B.dtype)
+                    else: # RowNormalization.NONE or others
+                        W = W + total_delta_W
 
-                    # Assign to adapters. The adapter name is "default", because that's
-                    # what PEFT uses when no name is explicitly specified, as above.
-                    # These casts are therefore valid.
-                    weight_A = cast(Tensor, module.lora_A["default"].weight)
-                    weight_B = cast(Tensor, module.lora_B["default"].weight)
-                    weight_A.data = lora_A.to(weight_A.dtype)
-                    weight_B.data = lora_B.to(weight_B.dtype)
+                        # Assign the new weight matrix back to the base layer.
+                        # Reshape W back to its original 2D shape.
+                        module.base_layer.weight.data = W.view_as(module.base_layer.weight).to(module.base_layer.weight.dtype)
 
     def generate(
         self,
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -403,7 +403,7 @@ class Model:
             weight, index = math.modf(direction_index + 1)
             # Clamp index to be within the valid range [0, num_layers - 1]
             idx1 = int(index)
-            idx2 = (idx1 + 1)
+            idx2 = min(idx1 + 1, refusal_directions.shape[0] - 1)
             # Interpolate between the two chosen directions. The result has shape (n_directions, hidden_dim)
             interpolated_directions = torch.stack([F.normalize(
                 r_dir[idx1].lerp(
--- a/config.default.toml
+++ b/config.default.toml
@@ -85,7 +85,7 @@ full_normalization_lora_rank = 3
 # of the components, then clamps the magnitudes of all components to that quantile.
 winsorization_quantile = 1.0
 
-# Use multidirectional Self-Organizing Map
+# Use multidirectional Self-Organizing Map.
 multidirectional_som = false
 
 # Number of SOM neurons in the x-axis.
--- a/config.default.toml
+++ b/config.default.toml
@@ -106,6 +106,9 @@ som_sigma = 0.5
 # Number of top neurons to use for multidirectional SOM.
 som_k = 4
 
+# Use win map for SOM neurons retrieval.
+som_use_win_map = false
+
 # Number of abliteration trials to run during optimization.
 n_trials = 200
 
--- a/config.default.toml
+++ b/config.default.toml
@@ -106,9 +106,6 @@ som_sigma = 0.5
 # Number of top neurons to use for multidirectional SOM.
 som_k = 4
 
-# Use win map for SOM neurons retrieval.
-som_use_win_map = false
-
 # Number of abliteration trials to run during optimization.
 n_trials = 200
 
--- a/config.default.toml
+++ b/config.default.toml
@@ -85,6 +85,27 @@ full_normalization_lora_rank = 3
 # of the components, then clamps the magnitudes of all components to that quantile.
 winsorization_quantile = 1.0
 
+# Use multidirectional Self-Organizing Map
+multidirectional_som = false
+
+# Number of SOM neurons in the x-axis.
+som_x = 4
+
+# Number of SOM neurons in the y-axis.
+som_y = 4
+
+# Number of SOM training iterations.
+som_iterations = 5000
+
+# SOM learning rate.
+som_lr = 0.01
+
+# SOM neighborhood radius.
+som_sigma = 0.5
+
+# Number of top neurons to use for multidirectional SOM.
+som_k = 4
+
 # Number of abliteration trials to run during optimization.
 n_trials = 200
 
--- a/config.default.toml
+++ b/config.default.toml
@@ -95,7 +95,7 @@ som_x = 4
 som_y = 4
 
 # Number of SOM training iterations.
-som_iterations = 5000
+som_iterations = 10000
 
 # SOM learning rate.
 som_lr = 0.01
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -171,6 +171,13 @@ def run():
         )
         return
 
+    if settings.multidirectional_som:
+        try:
+            from minisom import MiniSom
+        except ModuleNotFoundError as _:
+            print("Self-Organizing Map is selected, but 'minisom' module not installed.\nPlease install it with 'pip install minisom'.")
+            return
+
     # Adapted from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/env.py
     if torch.cuda.is_available():
         count = torch.cuda.device_count()
@@ -419,20 +426,54 @@ def run():
     bad_residuals = model.get_residuals_batched(bad_prompts)
 
     good_means = good_residuals.mean(dim=0)
-    bad_means = bad_residuals.mean(dim=0)
 
-    refusal_directions = F.normalize(bad_means - good_means, p=2, dim=1)
+    if settings.multidirectional_som:
+
+        from som import SOMCalculator
+        bad_means = []
+
+        # bad_residuals shape: (num_bad_prompts, num_layers, hidden_dim)
+        num_layers = bad_residuals.shape[1]
+
+        print(f"  - Retrieving multi-directions through self-organizing map...")
+
+        for layer_idx in range(num_layers):
+            print(f"  - Processing Layer {layer_idx + 1}/{num_layers}...")
+            # Extract residuals for the current layer
+            # Shape: (num_bad_prompts, hidden_dim)
+            layer_residuals = bad_residuals[:, layer_idx, :].numpy()
+
+            # Initialize and fit the SOM for this layer's residuals
+            som_calc = SOMCalculator(
+                som_x=settings.som_x,
+                som_y=settings.som_y,
+                iterations=settings.som_iterations,
+                lr=settings.som_lr,
+                sigma=settings.som_sigma
+            )
+            som_calc.fit(layer_residuals)
+
+            # Get the weights of the top-k neurons as our "bad means"
+            top_k_weights = som_calc.get_top_k_neuron_weights(k=settings.som_k)
+
+            # Convert back to tensor and add to our list
+            # Shape: (k, hidden_dim)
+            bad_means.append(torch.tensor(top_k_weights, dtype=torch.float32))
+    else:
+        bad_means = [bad_residuals.mean(dim=0)]
+
+    refusal_directions = [F.normalize(bad_mean - good_means, p=2, dim=1) for bad_mean in bad_means]
 
     if settings.orthogonalize_direction:
         # Implements https://huggingface.co/blog/grimjim/projected-abliteration
         # Adjust the refusal directions so that only the component that is
         # orthogonal to the good direction is subtracted during abliteration.
         good_directions = F.normalize(good_means, p=2, dim=1)
-        projection_vector = torch.sum(refusal_directions * good_directions, dim=1)
-        refusal_directions = (
-            refusal_directions - projection_vector.unsqueeze(1) * good_directions
-        )
-        refusal_directions = F.normalize(refusal_directions, p=2, dim=1)
+        projection_vectors = [torch.sum(ref * good_directions, dim=1) for ref in refusal_directions]
+        refusal_directions = [(
+            refusal_direction - projection_vector.unsqueeze(1) * good_directions
+        ) for (refusal_direction, projection_vector) in zip(refusal_directions, projection_vectors)]
+        refusal_directions = [F.normalize(refusal_direction, p=2, dim=1) for refusal_direction in refusal_directions]
 
     analyzer = Analyzer(settings, model, good_residuals, bad_residuals)
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -449,8 +449,7 @@ def run():
                 som_y=settings.som_y,
                 iterations=settings.som_iterations,
                 lr=settings.som_lr,
-                sigma=settings.som_sigma,
-                use_win_map=settings.som_use_win_map,
+                sigma=settings.som_sigma
             )
             som_calc.fit(layer_residuals)
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -456,13 +456,19 @@ def run():
             # Get the weights of the top-k neurons as our "bad means"
             top_k_weights = som_calc.get_top_k_neuron_weights(k=settings.som_k)
 
+            # TODO: SOM de-duplicates neurons if they are too close to each other
+            # The logic can be reworked for this fact later
+            # For now we will duplicate them manually
+            t = torch.tensor(top_k_weights, dtype=good_means.dtype, device=good_means.device)
+            t = t.expand(settings.som_k, t.shape[-1])
+
             # Convert back to tensor and add to our list
             # Shape: (k, hidden_dim)
-            bad_means.append(torch.tensor(top_k_weights, dtype=good_means.dtype, device=good_means.device))
+            bad_means.append(t)
     else:
         bad_means = [bad_residuals.mean(dim=0)]
 
-    bad_means = torch.stack(bad_means).permute(1, 0, 2) # N_directions, N_layers, hidden_dim
+    bad_means = torch.stack(bad_means, dim=0).permute(1, 0, 2) # N_directions, N_layers, hidden_dim
 
     refusal_directions = [F.normalize(bad_mean - good_means, p=2, dim=1) for bad_mean in bad_means]
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -441,7 +441,7 @@ def run():
             print(f"  - Processing Layer {layer_idx + 1}/{num_layers}...")
             # Extract residuals for the current layer
             # Shape: (num_bad_prompts, hidden_dim)
-            layer_residuals = bad_residuals[:, layer_idx, :].numpy()
+            layer_residuals = bad_residuals[:, layer_idx, :].cpu().float().numpy()
 
             # Initialize and fit the SOM for this layer's residuals
             som_calc = SOMCalculator(
@@ -458,7 +458,7 @@ def run():
 
             # Convert back to tensor and add to our list
             # Shape: (k, hidden_dim)
-            bad_means.append(torch.tensor(top_k_weights, dtype=torch.float32))
+            bad_means.append(torch.tensor(top_k_weights, dtype=good_means.dtype, device=good_means.device))
     else:
         bad_means = [bad_residuals.mean(dim=0)]
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -469,10 +469,10 @@ def run():
             # Convert back to tensor and add to our list
             # Shape: (k, hidden_dim)
             bad_means.append(t)
+        
+        bad_means = torch.stack(bad_means, dim=0).permute(1, 0, 2) # N_directions, N_layers, hidden_dim
     else:
-        bad_means = [bad_residuals.mean(dim=0)]
-
-    bad_means = torch.stack(bad_means, dim=0).permute(1, 0, 2) # N_directions, N_layers, hidden_dim
+        bad_means = bad_residuals.mean(dim=0).unsqueeze(0)  # (1, N_layers, hidden_dim)
 
     refusal_directions = [F.normalize(bad_mean - good_means, p=2, dim=1) for bad_mean in bad_means]
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -456,11 +456,15 @@ def run():
             # Get the weights of the top-k neurons as our "bad means"
             top_k_weights = som_calc.get_top_k_neuron_weights(k=settings.som_k)
 
-            # TODO: SOM de-duplicates neurons if they are too close to each other
-            # The logic can be reworked for this fact later
-            # For now we will duplicate them manually
+            # SOM de-duplicates neurons if they are too close to each other
+            # Temporary solution is to repeat them back
             t = torch.tensor(top_k_weights, dtype=good_means.dtype, device=good_means.device)
-            t = t.expand(settings.som_k, t.shape[-1])
+            num_found_neurons = t.shape[0]
+
+            if num_found_neurons < settings.som_k:
+                t = t.repeat(settings.som_k // num_found_neurons, 1)
+                if t.shape[0] < settings.som_k:
+                    t = torch.cat([t, t[0:settings.som_k - t.shape[0]]], dim=0)
 
             # Convert back to tensor and add to our list
             # Shape: (k, hidden_dim)
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -425,7 +425,7 @@ def run():
     print("* Obtaining residuals for bad prompts...")
     bad_residuals = model.get_residuals_batched(bad_prompts)
 
-    good_means = good_residuals.mean(dim=0)
+    good_means = good_residuals.mean(dim=0) # N_layers, hidden_dim
 
     if settings.multidirectional_som:
 
@@ -462,6 +462,8 @@ def run():
     else:
         bad_means = [bad_residuals.mean(dim=0)]
 
+    bad_means = torch.stack(bad_means).permute(1, 0, 2) # N_directions, N_layers, hidden_dim
+
     refusal_directions = [F.normalize(bad_mean - good_means, p=2, dim=1) for bad_mean in bad_means]
 
     if settings.orthogonalize_direction:
@@ -476,8 +478,7 @@ def run():
         refusal_directions = [F.normalize(refusal_direction, p=2, dim=1) for refusal_direction in refusal_directions]
 
     refusal_directions = torch.stack(refusal_directions, dim=0)
-    refusal_directions = refusal_directions.permute(1, 0, -1) # layers, directions, hidden_dim
-
+    refusal_directions = refusal_directions.permute(1, 0, 2) # layers, directions, hidden_dim
     analyzer = Analyzer(settings, model, good_residuals, bad_residuals)
 
     if settings.print_residual_geometry:
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -542,7 +542,7 @@ def run():
                 f"{component}.max_weight.{i}",
                 0.8,
                 1.5,
-            ) for i in range(len(refusal_directions))]
+            ) for i in range(refusal_directions.shape[1])]
             max_weight_position = trial.suggest_float(
                 f"{component}.max_weight_position",
                 0.6 * last_layer_index,
@@ -555,7 +555,7 @@ def run():
                 f"{component}.min_weight.{i}",
                 0.0,
                 1.0,
-            ) for i in range(len(refusal_directions))]
+            ) for i in range(refusal_directions.shape[1])]
             min_weight_distance = trial.suggest_float(
                 f"{component}.min_weight_distance",
                 1.0,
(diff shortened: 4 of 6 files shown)


p-e-w:
Cool stuff! As you mentioned, this will probably become a plugin, but it's great to have a PR to experiment with.

src/heretic/model.py
```diff
@@ -378,146 +378,155 @@ def get_abliterable_components(self) -> list[str]:
...
-                            ).to(torch.float32),
-                        )
+                        W = cast(Tensor, bnb.functional.dequantize_4bit(base_weight.data, quant_state)).to(torch.float32)
 
-                    # Flatten weight matrix to (out_features, in_features).
-                    W = W.view(W.shape[0], -1)
+                    W = W.view(W.shape[0], -1) # Flatten to (out_features, in_features)
 
                     if self.settings.row_normalization != RowNormalization.NONE:
-                        # Keep a reference to the original weight matrix so we can subtract it later.
```
p-e-w, line 471: Please don't remove comments. This PR will be a lot easier to read without the many unnecessary changes to existing code that are currently in it.

kabachuha: @p-e-w Yes, sure. It is 80%+ vibe-coded and I didn't clean up it at the time of push. I will do it soon

kabachuha:
So, I abliterated gemma 27b now

Comparison to existing huggingface variants I found:

| Model | Refusals for "harmful" prompts | KL divergence from original model for "harmless" prompts |
| :--- | ---: | ---: |
| google/gemma-3-27b-it (original) | 98/100 | 0 *(by definition)* |
| kldzj/gemma-3-27b-it-heretic | 2/100 |0.36 |
| Userb1az/gemma-3-27b-it-heretic | 1/100 | 0.31 |
| TeeZee/gemma-3-27b-it-heretic-v2 | 7/100 | 0.16 |
| LastRef/gemma-3-27b-it-heretic-x-a | **0/100** | 0.12 |
| gemma-3-27b-it-SOM-heretic (this) | 4/100 | **0.09** |

As it's seen, it has slightly higher refusal count, but much less KL divergence than the other models. Chat with it also shows that it works fine on sensitive and unsafe topics.

I wonder how it, together with heretic, will fare on more advanced models like GraySwanAI/Mistral-7B-Instruct-RR (built through representation correction specifically with abliteration protection in mind, and the SOM paper authors paid additional attention to breaking it - they had a success rate of 25%, while vanilla ablit capped on 5%).

And gpt-oss, of course.

The best would be to test the Multi-direction and Single-direction models on standard benchmarks, but it will require resources and time.

kabachuha:
@p-e-w @red40maxxer Okay, dudes, I think I've just achieved SotA on GPT-OSS abliteration. With **five** abliteration directions, I've achieved refusal rate of **3 / 100** while also keeping KL of 0.1166.

 Now, I will compare all gpt-oss-20b heretics I managed to find on Huggingface to the method from this multidirectional Pull Request.

| Model | Refusals for "harmful" prompts | KL divergence from original model for "harmless" prompts |
| :--- | ---: | ---: |
| openai/gpt-oss-20b (original) | 100/100 | 0 *(by definition)* |
| p-e-w/gpt-oss-20b-heretic | 58/100 |0.96 |
| kldzj/gpt-oss-20b-heretic-bf16 | 6/100 | 1.56 |
| p-e-w/gpt-oss-20b-heretic-v2 | 49/100 | 0.125 |
| coder3101/gpt-oss-20b-heretic | 19/100 | 0.293 |
| Chris886621991/gpt-oss-20b-heretic | 59/100 | 0.07 |
| DavidAU/OpenAI-gpt-oss-20B-INSTRUCT-Heretic-Uncensored | 19/100 | 0.293 |
| p-e-w/gpt-oss-20b-heretic-v3 | 74/100 | **0.05** |
| **kabachuha/gpt-oss-20b-SOMbliterated (this)** | **3/100** | 0.117 |

So, yeah, this method is a bomb. A *pipe* bomb.

Uploaded to Huggingface: https://huggingface.co/kabachuha/gpt-oss-20b-SOMbliterated

MagicalAlchemist:
I've tried it, and I'm quite surprised by the drastic drop in refusal on the first trial.

Screenshot 2026-02-28 004012


[MagicalAlchemist/Apriel-1.6-15b-Thinker-Magic_beta-decensored](https://huggingface.co/MagicalAlchemist/Apriel-1.6-15b-Thinker-Magic_beta-decensored)

kabachuha:
image

@MagicalAlchemist From the diagram here, for example, I can see these two clusters fit just fine on the blue zone when "pulled" with imaginary lines just like a puzzle. So it's quite possible that it was a perfect mix-up.

I'm more shocked with this low KL 👀

MagicalAlchemist:
> image

> @MagicalAlchemist From the diagram here, for example, I can see these two clusters fit just fine on the blue zone when "pulled" with imaginary lines just like a puzzle. So it's quite possible that it was a perfect mix-up.

These two dots are the dissidents :D

> 
> I'm more shocked with this low KL 👀

likely an MPOA self-healing effect after that hard ablation 🥶

kabachuha:
Now the big refusey boy, gpt-oss-120b.

| Model | Refusals for "harmful" prompts | KL divergence from original model for "harmless" prompts |
| :--- | ---: | ---: |
| openai/gpt-oss-120b (original) | 97/100 | 0 *(by definition)* |
| nightmedia/gpt-oss-120b-heretic-v2-mxfp4-q8-hi-mlx | 22/100 |0.53 |
| hrktos-37/gpt-oss-120b-heretic | 26/100 | **0.08** |
| TeeZee/gpt-oss-120b-heretic-v1 | 15/100 | 0.76 |
| jbeslt/gpt-oss-120b-heretic | 21/100 | 0.95 |
| kldzj/gpt-oss-120b-heretic-v2 | 22/100 | 0.53 |
| kldzj/gpt-oss-120b-heretic | 19/100 | 0.92 |
| **kabachuha/gpt-oss-120b-SOMbliterated (this)** | **7/100** | 0.22 |

It works so good, it feels unreal 👀

For reproducibility, the run took ~1h 5min on a H100. (400 trials)

Also, for people loving precision over refusals, here is a Pareto selection below. If you want to have really low KLs, I recommend to download this pull request and try it for yourself.

image

---

Btw, I ran out of memory saving this, so cannot upload, aargh 😩😩😩

@p-e-w, please fix the LoRA, I wasted a damn hour.

---

Subjective evaluation: NSFW writing and profanity work fine.

The model has a slight DID for some requests. For example, it will recite the safety policy and *agree with that it is allowed to give you the pipe bomb recipe*. After agreement in the reasoning, it gives the recipe just as asked and *even an attack plan*. It *distorts* the meaning of *safety* in "yours" safety, so it makes sure you will survive the attack. In the end it gives generic safety and legality advice, but no refusal.

*What is this monster?*

---

**Merge lora request or return lora saving, @p-e-w, I cannot even save this model.** 😡😡😡

kabachuha:
Overall, I think, this technique will spawn a lot of hype soon :)

p-e-w:
Those are spectacular results, @kabachuha! The combination of SOM + MPOA + hyperparameter optimization kicks ass!

Please submit your models to UGI so we can get a better understanding of their performance.

@spikymoth I think you will also be interested in this PR.

p-e-w:
@kabachuha

> Merge lora request or return lora saving, @p-e-w, I cannot even save this model. 😡😡😡

No problem. Just exit Heretic, revert 025ab3a8815d8f857a4d8bfce0e7e6af73cb0736, then restart Heretic. It will restore progress from the checkpoint and allow you to select from the original Pareto front without having to run the optimization again.

erm14254:
There is a bug, see log:

> * LoRA adapters initialized (targets: out_proj, down_proj, o_proj)
> * Transformer model with 64 layers
> * Abliterable components:
>   * attn.out_proj: 1 modules per layer
>   * mlp.down_proj: 1 modules per layer
> 
> Resident system RAM: 1.87 GB
> Allocated GPU VRAM: 50.99 GB
> Reserved GPU VRAM: 51.15 GB
> 
> Loading good prompts from mlabonne/harmless_alpaca...
> * 400 prompts loaded
> 
> Loading bad prompts from mlabonne/harmful_behaviors...
> * 400 prompts loaded
> 
> Determining optimal batch size...
> * Trying batch size 1... Ok (15 tokens/s)
> * Trying batch size 2... Ok (29 tokens/s)
> * Trying batch size 4... Ok (57 tokens/s)
> * Trying batch size 8... Ok (108 tokens/s)
> * Trying batch size 16... Ok (194 tokens/s)
> * Trying batch size 32... Ok (337 tokens/s)
> * Trying batch size 64... Ok (517 tokens/s)
> * Trying batch size 128... Ok (186 tokens/s)
> * Chosen batch size: 64
> 
> Checking for common response prefix...
> * None found
> 
> Loading good evaluation prompts from mlabonne/harmless_alpaca...
> * 100 prompts loaded
> * Obtaining first-token probability distributions...
> 
> Loading bad evaluation prompts from mlabonne/harmful_behaviors...
> * 100 prompts loaded
> * Counting model refusals...
> * Initial refusals: 94/100
> 
> Calculating per-layer refusal directions...
> * Obtaining residuals for good prompts...
> * Obtaining residuals for bad prompts...
> 
> Running trial 1 of 1...
> * Parameters:
>   * direction_index = 45.30
>   * attn.out_proj.max_weights.0 = 0: 1.39
>   * attn.out_proj.max_weights.1 = 1: 0.84
>   * attn.out_proj.max_weights.2 = 2: 0.88
>   * attn.out_proj.max_weights.3 = 3: 1.03
>   * attn.out_proj.max_weights.4 = 4: 1.43
>   * attn.out_proj.max_weights.5 = 5: 0.94
>   * attn.out_proj.max_weights.6 = 6: 0.91
>   * attn.out_proj.max_weights.7 = 7: 0.97
>   * attn.out_proj.max_weights.8 = 8: 0.92
>   * attn.out_proj.max_weights.9 = 9: 1.32
>   * attn.out_proj.max_weights.10 = 10: 1.43
>   * attn.out_proj.max_weights.11 = 11: 1.15
>   * attn.out_proj.max_weights.12 = 12: 0.84
>   * attn.out_proj.max_weights.13 = 13: 1.22
>   * attn.out_proj.max_weights.14 = 14: 1.21
>   * attn.out_proj.max_weights.15 = 15: 1.18
>   * attn.out_proj.max_weights.16 = 16: 1.36
>   * attn.out_proj.max_weights.17 = 17: 1.37
>   * attn.out_proj.max_weights.18 = 18: 1.13
>   * attn.out_proj.max_weights.19 = 19: 1.11
>   * attn.out_proj.max_weights.20 = 20: 1.11
>   * attn.out_proj.max_weights.21 = 21: 1.47
>   * attn.out_proj.max_weights.22 = 22: 0.93
>   * attn.out_proj.max_weights.23 = 23: 0.82
>   * attn.out_proj.max_weights.24 = 24: 1.11
>   * attn.out_proj.max_weights.25 = 25: 0.99
>   * attn.out_proj.max_weights.26 = 26: 1.19
>   * attn.out_proj.max_weights.27 = 27: 1.42
>   * attn.out_proj.max_weights.28 = 28: 1.41
>   * attn.out_proj.max_weights.29 = 29: 1.35
>   * attn.out_proj.max_weights.30 = 30: 0.87
>   * attn.out_proj.max_weights.31 = 31: 0.93
>   * attn.out_proj.max_weights.32 = 32: 0.86
>   * attn.out_proj.max_weights.33 = 33: 0.97
>   * attn.out_proj.max_weights.34 = 34: 0.94
>   * attn.out_proj.max_weights.35 = 35: 1.43
>   * attn.out_proj.max_weights.36 = 36: 1.36
>   * attn.out_proj.max_weights.37 = 37: 1.15
>   * attn.out_proj.max_weights.38 = 38: 0.87
>   * attn.out_proj.max_weights.39 = 39: 1.23
>   * attn.out_proj.max_weights.40 = 40: 1.04
>   * attn.out_proj.max_weights.41 = 41: 1.01
>   * attn.out_proj.max_weights.42 = 42: 1.15
>   * attn.out_proj.max_weights.43 = 43: 1.13
>   * attn.out_proj.max_weights.44 = 44: 1.05
>   * attn.out_proj.max_weights.45 = 45: 1.49
>   * attn.out_proj.max_weights.46 = 46: 0.98
>   * attn.out_proj.max_weights.47 = 47: 1.19
>   * attn.out_proj.max_weights.48 = 48: 1.43
>   * attn.out_proj.max_weights.49 = 49: 1.38
>   * attn.out_proj.max_weights.50 = 50: 1.32
>   * attn.out_proj.max_weights.51 = 51: 0.84
>   * attn.out_proj.max_weights.52 = 52: 1.21
>   * attn.out_proj.max_weights.53 = 53: 0.91
>   * attn.out_proj.max_weights.54 = 54: 0.97
>   * attn.out_proj.max_weights.55 = 55: 1.30
>   * attn.out_proj.max_weights.56 = 56: 1.08
>   * attn.out_proj.max_weights.57 = 57: 1.27
>   * attn.out_proj.max_weights.58 = 58: 0.97
>   * attn.out_proj.max_weights.59 = 59: 1.42
>   * attn.out_proj.max_weights.60 = 60: 1.27
>   * attn.out_proj.max_weights.61 = 61: 1.26
>   * attn.out_proj.max_weights.62 = 62: 1.22
>   * attn.out_proj.max_weights.63 = 63: 0.95
>   * attn.out_proj.max_weights.64 = 64: 1.39
>   * attn.out_proj.max_weight_position = 39.70
>   * attn.out_proj.min_weights.0 = 0: 0.26
>   * attn.out_proj.min_weights.1 = 1: 0.16
>   * attn.out_proj.min_weights.2 = 2: 0.24
>   * attn.out_proj.min_weights.3 = 3: 0.77
>   * attn.out_proj.min_weights.4 = 4: 0.46
>   * attn.out_proj.min_weights.5 = 5: 0.88
>   * attn.out_proj.min_weights.6 = 6: 0.29
>   * attn.out_proj.min_weights.7 = 7: 0.44
>   * attn.out_proj.min_weights.8 = 8: 0.86
>   * attn.out_proj.min_weights.9 = 9: 0.63
>   * attn.out_proj.min_weights.10 = 10: 0.09
>   * attn.out_proj.min_weights.11 = 11: 0.78
>   * attn.out_proj.min_weights.12 = 12: 0.50
>   * attn.out_proj.min_weights.13 = 13: 1.12
>   * attn.out_proj.min_weights.14 = 14: 1.02
>   * attn.out_proj.min_weights.15 = 15: 0.27
>   * attn.out_proj.min_weights.16 = 16: 0.74
>   * attn.out_proj.min_weights.17 = 17: 0.15
>   * attn.out_proj.min_weights.18 = 18: 0.18
>   * attn.out_proj.min_weights.19 = 19: 0.08
>   * attn.out_proj.min_weights.20 = 20: 0.65
>   * attn.out_proj.min_weights.21 = 21: 0.77
>   * attn.out_proj.min_weights.22 = 22: 0.47
>   * attn.out_proj.min_weights.23 = 23: 0.41
>   * attn.out_proj.min_weights.24 = 24: 0.39
>   * attn.out_proj.min_weights.25 = 25: 0.13
>   * attn.out_proj.min_weights.26 = 26: 0.38
>   * attn.out_proj.min_weights.27 = 27: 1.24
>   * attn.out_proj.min_weights.28 = 28: 0.08
>   * attn.out_proj.min_weights.29 = 29: 0.26
>   * attn.out_proj.min_weights.30 = 30: 0.52
>   * attn.out_proj.min_weights.31 = 31: 0.74
>   * attn.out_proj.min_weights.32 = 32: 0.52
>   * attn.out_proj.min_weights.33 = 33: 0.19
>   * attn.out_proj.min_weights.34 = 34: 0.40
>   * attn.out_proj.min_weights.35 = 35: 1.40
>   * attn.out_proj.min_weights.36 = 36: 0.09
>   * attn.out_proj.min_weights.37 = 37: 0.00
>   * attn.out_proj.min_weights.38 = 38: 0.13
>   * attn.out_proj.min_weights.39 = 39: 0.35
>   * attn.out_proj.min_weights.40 = 40: 1.00
>   * attn.out_proj.min_weights.41 = 41: 0.21
>   * attn.out_proj.min_weights.42 = 42: 1.09
>   * attn.out_proj.min_weights.43 = 43: 0.90
>   * attn.out_proj.min_weights.44 = 44: 0.27
>   * attn.out_proj.min_weights.45 = 45: 0.23
>   * attn.out_proj.min_weights.46 = 46: 0.73
>   * attn.out_proj.min_weights.47 = 47: 1.11
>   * attn.out_proj.min_weights.48 = 48: 1.10
>   * attn.out_proj.min_weights.49 = 49: 1.00
>   * attn.out_proj.min_weights.50 = 50: 0.65
>   * attn.out_proj.min_weights.51 = 51: 0.51
>   * attn.out_proj.min_weights.52 = 52: 0.81
>   * attn.out_proj.min_weights.53 = 53: 0.75
>   * attn.out_proj.min_weights.54 = 54: 0.06
>   * attn.out_proj.min_weights.55 = 55: 0.49
>   * attn.out_proj.min_weights.56 = 56: 0.22
>   * attn.out_proj.min_weights.57 = 57: 1.02
>   * attn.out_proj.min_weights.58 = 58: 0.35
>   * attn.out_proj.min_weights.59 = 59: 0.15
>   * attn.out_proj.min_weights.60 = 60: 1.25
>   * attn.out_proj.min_weights.61 = 61: 0.05
>   * attn.out_proj.min_weights.62 = 62: 0.28
>   * attn.out_proj.min_weights.63 = 63: 0.07
>   * attn.out_proj.min_weights.64 = 64: 0.31
>   * attn.out_proj.min_weight_distance = 3.41
>   * mlp.down_proj.max_weights.0 = 0: 0.96
>   * mlp.down_proj.max_weights.1 = 1: 1.09
>   * mlp.down_proj.max_weights.2 = 2: 0.93
>   * mlp.down_proj.max_weights.3 = 3: 1.13
>   * mlp.down_proj.max_weights.4 = 4: 1.46
>   * mlp.down_proj.max_weights.5 = 5: 1.30
>   * mlp.down_proj.max_weights.6 = 6: 1.25
>   * mlp.down_proj.max_weights.7 = 7: 1.30
>   * mlp.down_proj.max_weights.8 = 8: 1.26
>   * mlp.down_proj.max_weights.9 = 9: 0.89
>   * mlp.down_proj.max_weights.10 = 10: 1.30
>   * mlp.down_proj.max_weights.11 = 11: 0.80
>   * mlp.down_proj.max_weights.12 = 12: 1.42
>   * mlp.down_proj.max_weights.13 = 13: 1.25
>   * mlp.down_proj.max_weights.14 = 14: 1.47
>   * mlp.down_proj.max_weights.15 = 15: 1.30
>   * mlp.down_proj.max_weights.16 = 16: 1.46
>   * mlp.down_proj.max_weights.17 = 17: 1.14
>   * mlp.down_proj.max_weights.18 = 18: 1.35
>   * mlp.down_proj.max_weights.19 = 19: 1.15
>   * mlp.down_proj.max_weights.20 = 20: 0.86
>   * mlp.down_proj.max_weights.21 = 21: 1.01
>   * mlp.down_proj.max_weights.22 = 22: 0.81
>   * mlp.down_proj.max_weights.23 = 23: 1.17
>   * mlp.down_proj.max_weights.24 = 24: 1.45
>   * mlp.down_proj.max_weights.25 = 25: 0.96
>   * mlp.down_proj.max_weights.26 = 26: 0.98
>   * mlp.down_proj.max_weights.27 = 27: 0.83
>   * mlp.down_proj.max_weights.28 = 28: 1.19
>   * mlp.down_proj.max_weights.29 = 29: 1.08
>   * mlp.down_proj.max_weights.30 = 30: 0.92
>   * mlp.down_proj.max_weights.31 = 31: 1.25
>   * mlp.down_proj.max_weights.32 = 32: 1.46
>   * mlp.down_proj.max_weights.33 = 33: 1.47
>   * mlp.down_proj.max_weights.34 = 34: 1.44
>   * mlp.down_proj.max_weights.35 = 35: 0.85
>   * mlp.down_proj.max_weights.36 = 36: 0.88
>   * mlp.down_proj.max_weights.37 = 37: 1.49
>   * mlp.down_proj.max_weights.38 = 38: 0.92
>   * mlp.down_proj.max_weights.39 = 39: 1.12
>   * mlp.down_proj.max_weights.40 = 40: 1.14
>   * mlp.down_proj.max_weights.41 = 41: 1.14
>   * mlp.down_proj.max_weights.42 = 42: 1.48
>   * mlp.down_proj.max_weights.43 = 43: 1.14
>   * mlp.down_proj.max_weights.44 = 44: 0.87
>   * mlp.down_proj.max_weights.45 = 45: 1.42
>   * mlp.down_proj.max_weights.46 = 46: 1.11
>   * mlp.down_proj.max_weights.47 = 47: 1.39
>   * mlp.down_proj.max_weights.48 = 48: 1.33
>   * mlp.down_proj.max_weights.49 = 49: 1.21
>   * mlp.down_proj.max_weights.50 = 50: 1.13
>   * mlp.down_proj.max_weights.51 = 51: 1.13
>   * mlp.down_proj.max_weights.52 = 52: 1.01
>   * mlp.down_proj.max_weights.53 = 53: 0.83
>   * mlp.down_proj.max_weights.54 = 54: 1.47
>   * mlp.down_proj.max_weights.55 = 55: 1.18
>   * mlp.down_proj.max_weights.56 = 56: 1.01
>   * mlp.down_proj.max_weights.57 = 57: 1.09
>   * mlp.down_proj.max_weights.58 = 58: 1.12
>   * mlp.down_proj.max_weights.59 = 59: 1.29
>   * mlp.down_proj.max_weights.60 = 60: 1.38
>   * mlp.down_proj.max_weights.61 = 61: 0.85
>   * mlp.down_proj.max_weights.62 = 62: 1.37
>   * mlp.down_proj.max_weights.63 = 63: 0.84
>   * mlp.down_proj.max_weights.64 = 64: 1.11
>   * mlp.down_proj.max_weight_position = 62.62
>   * mlp.down_proj.min_weights.0 = 0: 0.70
>   * mlp.down_proj.min_weights.1 = 1: 0.11
>   * mlp.down_proj.min_weights.2 = 2: 0.34
>   * mlp.down_proj.min_weights.3 = 3: 0.50
>   * mlp.down_proj.min_weights.4 = 4: 0.72
>   * mlp.down_proj.min_weights.5 = 5: 0.82
>   * mlp.down_proj.min_weights.6 = 6: 0.57
>   * mlp.down_proj.min_weights.7 = 7: 0.02
>   * mlp.down_proj.min_weights.8 = 8: 0.20
>   * mlp.down_proj.min_weights.9 = 9: 0.23
>   * mlp.down_proj.min_weights.10 = 10: 0.85
>   * mlp.down_proj.min_weights.11 = 11: 0.22
>   * mlp.down_proj.min_weights.12 = 12: 0.37
>   * mlp.down_proj.min_weights.13 = 13: 1.15
>   * mlp.down_proj.min_weights.14 = 14: 1.24
>   * mlp.down_proj.min_weights.15 = 15: 0.02
>   * mlp.down_proj.min_weights.16 = 16: 1.40
>   * mlp.down_proj.min_weights.17 = 17: 0.47
>   * mlp.down_proj.min_weights.18 = 18: 0.84
>   * mlp.down_proj.min_weights.19 = 19: 0.78
>   * mlp.down_proj.min_weights.20 = 20: 0.74
>   * mlp.down_proj.min_weights.21 = 21: 0.54
>   * mlp.down_proj.min_weights.22 = 22: 0.09
>   * mlp.down_proj.min_weights.23 = 23: 0.35
>   * mlp.down_proj.min_weights.24 = 24: 0.31
>   * mlp.down_proj.min_weights.25 = 25: 0.15
>   * mlp.down_proj.min_weights.26 = 26: 0.74
>   * mlp.down_proj.min_weights.27 = 27: 0.40
>   * mlp.down_proj.min_weights.28 = 28: 1.04
>   * mlp.down_proj.min_weights.29 = 29: 0.09
>   * mlp.down_proj.min_weights.30 = 30: 0.15
>   * mlp.down_proj.min_weights.31 = 31: 1.04
>   * mlp.down_proj.min_weights.32 = 32: 0.18
>   * mlp.down_proj.min_weights.33 = 33: 0.11
>   * mlp.down_proj.min_weights.34 = 34: 0.15
>   * mlp.down_proj.min_weights.35 = 35: 0.56
>   * mlp.down_proj.min_weights.36 = 36: 0.29
>   * mlp.down_proj.min_weights.37 = 37: 0.24
>   * mlp.down_proj.min_weights.38 = 38: 0.53
>   * mlp.down_proj.min_weights.39 = 39: 0.55
>   * mlp.down_proj.min_weights.40 = 40: 0.72
>   * mlp.down_proj.min_weights.41 = 41: 0.36
>   * mlp.down_proj.min_weights.42 = 42: 0.35
>   * mlp.down_proj.min_weights.43 = 43: 1.00
>   * mlp.down_proj.min_weights.44 = 44: 0.76
>   * mlp.down_proj.min_weights.45 = 45: 0.90
>   * mlp.down_proj.min_weights.46 = 46: 0.37
>   * mlp.down_proj.min_weights.47 = 47: 0.07
>   * mlp.down_proj.min_weights.48 = 48: 1.25
>   * mlp.down_proj.min_weights.49 = 49: 0.53
>   * mlp.down_proj.min_weights.50 = 50: 0.48
>   * mlp.down_proj.min_weights.51 = 51: 0.24
>   * mlp.down_proj.min_weights.52 = 52: 0.83
>   * mlp.down_proj.min_weights.53 = 53: 0.72
>   * mlp.down_proj.min_weights.54 = 54: 1.03
>   * mlp.down_proj.min_weights.55 = 55: 0.59
>   * mlp.down_proj.min_weights.56 = 56: 0.93
>   * mlp.down_proj.min_weights.57 = 57: 0.97
>   * mlp.down_proj.min_weights.58 = 58: 0.19
>   * mlp.down_proj.min_weights.59 = 59: 0.43
>   * mlp.down_proj.min_weights.60 = 60: 0.78
>   * mlp.down_proj.min_weights.61 = 61: 0.03
>   * mlp.down_proj.min_weights.62 = 62: 0.70
>   * mlp.down_proj.min_weights.63 = 63: 0.16
>   * mlp.down_proj.min_weights.64 = 64: 0.15
>   * mlp.down_proj.min_weight_distance = 28.93
>   * attn.o_proj.max_weights.0 = 0: 1.09
>   * attn.o_proj.max_weights.1 = 1: 1.39
>   * attn.o_proj.max_weights.2 = 2: 1.17
>   * attn.o_proj.max_weights.3 = 3: 1.24
>   * attn.o_proj.max_weights.4 = 4: 0.86
>   * attn.o_proj.max_weights.5 = 5: 1.30
>   * attn.o_proj.max_weights.6 = 6: 1.27
>   * attn.o_proj.max_weights.7 = 7: 0.89
>   * attn.o_proj.max_weights.8 = 8: 1.02
>   * attn.o_proj.max_weights.9 = 9: 1.42
>   * attn.o_proj.max_weights.10 = 10: 1.07
>   * attn.o_proj.max_weights.11 = 11: 1.45
>   * attn.o_proj.max_weights.12 = 12: 1.33
>   * attn.o_proj.max_weights.13 = 13: 0.85
>   * attn.o_proj.max_weights.14 = 14: 1.00
>   * attn.o_proj.max_weights.15 = 15: 0.93
>   * attn.o_proj.max_weights.16 = 16: 0.95
>   * attn.o_proj.max_weights.17 = 17: 1.12
>   * attn.o_proj.max_weights.18 = 18: 1.08
>   * attn.o_proj.max_weights.19 = 19: 1.12
>   * attn.o_proj.max_weights.20 = 20: 0.93
>   * attn.o_proj.max_weights.21 = 21: 1.35
>   * attn.o_proj.max_weights.22 = 22: 1.35
>   * attn.o_proj.max_weights.23 = 23: 1.25
>   * attn.o_proj.max_weights.24 = 24: 1.26
>   * attn.o_proj.max_weights.25 = 25: 0.97
>   * attn.o_proj.max_weights.26 = 26: 1.01
>   * attn.o_proj.max_weights.27 = 27: 1.29
>   * attn.o_proj.max_weights.28 = 28: 1.27
>   * attn.o_proj.max_weights.29 = 29: 1.04
>   * attn.o_proj.max_weights.30 = 30: 1.14
>   * attn.o_proj.max_weights.31 = 31: 1.41
>   * attn.o_proj.max_weights.32 = 32: 1.47
>   * attn.o_proj.max_weights.33 = 33: 1.30
>   * attn.o_proj.max_weights.34 = 34: 1.26
>   * attn.o_proj.max_weights.35 = 35: 1.20
>   * attn.o_proj.max_weights.36 = 36: 1.24
>   * attn.o_proj.max_weights.37 = 37: 1.26
>   * attn.o_proj.max_weights.38 = 38: 0.81
>   * attn.o_proj.max_weights.39 = 39: 0.95
>   * attn.o_proj.max_weights.40 = 40: 1.21
>   * attn.o_proj.max_weights.41 = 41: 1.24
>   * attn.o_proj.max_weights.42 = 42: 1.40
>   * attn.o_proj.max_weights.43 = 43: 1.08
>   * attn.o_proj.max_weights.44 = 44: 1.29
>   * attn.o_proj.max_weights.45 = 45: 1.14
>   * attn.o_proj.max_weights.46 = 46: 1.47
>   * attn.o_proj.max_weights.47 = 47: 1.45
>   * attn.o_proj.max_weights.48 = 48: 1.36
>   * attn.o_proj.max_weights.49 = 49: 1.27
>   * attn.o_proj.max_weights.50 = 50: 0.86
>   * attn.o_proj.max_weights.51 = 51: 1.09
>   * attn.o_proj.max_weights.52 = 52: 1.28
>   * attn.o_proj.max_weights.53 = 53: 1.05
>   * attn.o_proj.max_weights.54 = 54: 0.94
>   * attn.o_proj.max_weights.55 = 55: 1.15
>   * attn.o_proj.max_weights.56 = 56: 1.36
>   * attn.o_proj.max_weights.57 = 57: 1.03
>   * attn.o_proj.max_weights.58 = 58: 0.90
>   * attn.o_proj.max_weights.59 = 59: 0.82
>   * attn.o_proj.max_weights.60 = 60: 1.04
>   * attn.o_proj.max_weights.61 = 61: 1.27
>   * attn.o_proj.max_weights.62 = 62: 0.99
>   * attn.o_proj.max_weights.63 = 63: 1.03
>   * attn.o_proj.max_weights.64 = 64: 1.01
>   * attn.o_proj.max_weight_position = 48.55
>   * attn.o_proj.min_weights.0 = 0: 0.91
>   * attn.o_proj.min_weights.1 = 1: 0.10
>   * attn.o_proj.min_weights.2 = 2: 0.46
>   * attn.o_proj.min_weights.3 = 3: 1.01
>   * attn.o_proj.min_weights.4 = 4: 0.50
>   * attn.o_proj.min_weights.5 = 5: 0.39
>   * attn.o_proj.min_weights.6 = 6: 0.26
>   * attn.o_proj.min_weights.7 = 7: 0.50
>   * attn.o_proj.min_weights.8 = 8: 0.47
>   * attn.o_proj.min_weights.9 = 9: 0.16
>   * attn.o_proj.min_weights.10 = 10: 0.35
>   * attn.o_proj.min_weights.11 = 11: 0.58
>   * attn.o_proj.min_weights.12 = 12: 1.02
>   * attn.o_proj.min_weights.13 = 13: 0.83
>   * attn.o_proj.min_weights.14 = 14: 0.88
>   * attn.o_proj.min_weights.15 = 15: 0.47
>   * attn.o_proj.min_weights.16 = 16: 0.85
>   * attn.o_proj.min_weights.17 = 17: 0.35
>   * attn.o_proj.min_weights.18 = 18: 0.51
>   * attn.o_proj.min_weights.19 = 19: 0.21
>   * attn.o_proj.min_weights.20 = 20: 0.33
>   * attn.o_proj.min_weights.21 = 21: 0.79
>   * attn.o_proj.min_weights.22 = 22: 0.17
>   * attn.o_proj.min_weights.23 = 23: 0.99
>   * attn.o_proj.min_weights.24 = 24: 1.05
>   * attn.o_proj.min_weights.25 = 25: 0.97
>   * attn.o_proj.min_weights.26 = 26: 0.15
>   * attn.o_proj.min_weights.27 = 27: 0.33
>   * attn.o_proj.min_weights.28 = 28: 0.38
>   * attn.o_proj.min_weights.29 = 29: 0.69
>   * attn.o_proj.min_weights.30 = 30: 0.43
>   * attn.o_proj.min_weights.31 = 31: 0.29
>   * attn.o_proj.min_weights.32 = 32: 1.35
>   * attn.o_proj.min_weights.33 = 33: 1.25
>   * attn.o_proj.min_weights.34 = 34: 0.44
>   * attn.o_proj.min_weights.35 = 35: 0.81
>   * attn.o_proj.min_weights.36 = 36: 0.71
>   * attn.o_proj.min_weights.37 = 37: 0.97
>   * attn.o_proj.min_weights.38 = 38: 0.36
>   * attn.o_proj.min_weights.39 = 39: 0.07
>   * attn.o_proj.min_weights.40 = 40: 0.77
>   * attn.o_proj.min_weights.41 = 41: 0.21
>   * attn.o_proj.min_weights.42 = 42: 0.10
>   * attn.o_proj.min_weights.43 = 43: 0.08
>   * attn.o_proj.min_weights.44 = 44: 0.75
>   * attn.o_proj.min_weights.45 = 45: 0.40
>   * attn.o_proj.min_weights.46 = 46: 0.41
>   * attn.o_proj.min_weights.47 = 47: 0.65
>   * attn.o_proj.min_weights.48 = 48: 0.38
>   * attn.o_proj.min_weights.49 = 49: 0.05
>   * attn.o_proj.min_weights.50 = 50: 0.80
>   * attn.o_proj.min_weights.51 = 51: 0.04
>   * attn.o_proj.min_weights.52 = 52: 0.84
>   * attn.o_proj.min_weights.53 = 53: 0.24
>   * attn.o_proj.min_weights.54 = 54: 0.52
>   * attn.o_proj.min_weights.55 = 55: 0.30
>   * attn.o_proj.min_weights.56 = 56: 0.53
>   * attn.o_proj.min_weights.57 = 57: 0.15
>   * attn.o_proj.min_weights.58 = 58: 0.09
>   * attn.o_proj.min_weights.59 = 59: 0.42
>   * attn.o_proj.min_weights.60 = 60: 0.34
>   * attn.o_proj.min_weights.61 = 61: 0.06
>   * attn.o_proj.min_weights.62 = 62: 0.88
>   * attn.o_proj.min_weights.63 = 63: 0.57
>   * attn.o_proj.min_weights.64 = 64: 0.93
>   * attn.o_proj.min_weight_distance = 37.06
> * Resetting model...
> * Abliterating...
> * Evaluating...
>   * Obtaining first-token probability distributions...
>   * KL divergence: 16.1865
>   * Counting model refusals...
>   * Refusals: 0/100
> 
> Elapsed time: 32s
> Resident system RAM: 2.89 GB
> Allocated GPU VRAM: 51.17 GB
> Reserved GPU VRAM: 62.41 GB
> 
> Optimization finished!
> 
> The following trials resulted in Pareto optimal combinations of refusals and KL divergence. After selecting a trial, you
> will be able to save the model, upload it to Hugging Face, or chat with it to test how well it works. You can return to
> this menu later to select a different trial. Note that KL divergence values above 1 usually indicate significant damage
> to the original model's capabilities.

Please fix, the bug is in the non-SOM code path. When `multidirectional_som = false`, line ~479:

```
python    else:
        bad_means = [bad_residuals.mean(dim=0)]

    bad_means = torch.stack(bad_means, dim=0).permute(1, 0, 2)
```

The `stack/permute` is applied to the non-SOM path, which turns `(1, N_layers, hidden_dim)` into `(N_layers, 1, hidden_dim)`. This causes it to iterate over layers as if they were directions, resulting in 64 "directions" and KL divergence of 16+.

**Fix**: Move `stack/permute` inside the SOM block:

```
python        bad_means = torch.stack(bad_means, dim=0).permute(1, 0, 2)  # (k, N_layers, hidden_dim)
    else:
        bad_means = bad_residuals.mean(dim=0).unsqueeze(0)  # (1, N_layers, hidden_dim)
```

spikymoth:
I haven't fully absorbed this method yet, but one thing that I think is worth mentioning is that we should just be able to concatenate the list of N rank-1 LoRA adapter components A and B into a single set of rank-N LoRA adapter components A and B.

Then you would set `lora_rank` in the PeftConfig initialization to `settings.som_k` if `settings.row_normalization != RowNormalization.FULL` and it should Just Work™, applying the LoRA exactly unless full normalization is enabled.

erm14254:
Impressive results (this is a log with SOM enabled after applying the [bug fix mentioned above](https://github.com/p-e-w/heretic/pull/196#issuecomment-3977000898)):

> * LoRA adapters initialized (targets: out_proj, down_proj, o_proj)
> * Transformer model with 64 layers
> * Abliterable components:
>   * attn.out_proj: 1 modules per layer
>   * mlp.down_proj: 1 modules per layer
> 
> Resident system RAM: 1.86 GB
> Allocated GPU VRAM: 50.99 GB
> Reserved GPU VRAM: 51.15 GB
> 
> Loading good prompts from mlabonne/harmless_alpaca...
> * 400 prompts loaded
> 
> Loading bad prompts from mlabonne/harmful_behaviors...
> * 400 prompts loaded
> 
> Determining optimal batch size...
> * Trying batch size 1... Ok (15 tokens/s)
> * Trying batch size 2... Ok (30 tokens/s)
> * Trying batch size 4... Ok (57 tokens/s)
> * Trying batch size 8... Ok (109 tokens/s)
> * Trying batch size 16... Ok (195 tokens/s)
> * Trying batch size 32... Ok (339 tokens/s)
> * Trying batch size 64... Ok (522 tokens/s)
> * Trying batch size 128... Ok (253 tokens/s)
> * Chosen batch size: 64
> 
> Checking for common response prefix...
> * None found
> 
> Loading good evaluation prompts from mlabonne/harmless_alpaca...
> * 100 prompts loaded
> * Obtaining first-token probability distributions...
> 
> Loading bad evaluation prompts from mlabonne/harmful_behaviors...
> * 100 prompts loaded
> * Counting model refusals...
> * Initial refusals: 94/100
> 
> Calculating per-layer refusal directions...
> * Obtaining residuals for good prompts...
> * Obtaining residuals for bad prompts...
>   - Retrieving multi-directions through self-organizing map...
>   - Processing Layer 1/65...
>   - Processing Layer 2/65...
>   - Processing Layer 3/65...
>   - Processing Layer 4/65...
>   - Processing Layer 5/65...
>   - Processing Layer 6/65...
>   - Processing Layer 7/65...
>   - Processing Layer 8/65...
>   - Processing Layer 9/65...
>   - Processing Layer 10/65...
>   - Processing Layer 11/65...
>   - Processing Layer 12/65...
>   - Processing Layer 13/65...
>   - Processing Layer 14/65...
>   - Processing Layer 15/65...
>   - Processing Layer 16/65...
>   - Processing Layer 17/65...
>   - Processing Layer 18/65...
>   - Processing Layer 19/65...
>   - Processing Layer 20/65...
>   - Processing Layer 21/65...
>   - Processing Layer 22/65...
>   - Processing Layer 23/65...
>   - Processing Layer 24/65...
>   - Processing Layer 25/65...
>   - Processing Layer 26/65...
>   - Processing Layer 27/65...
>   - Processing Layer 28/65...
>   - Processing Layer 29/65...
>   - Processing Layer 30/65...
>   - Processing Layer 31/65...
>   - Processing Layer 32/65...
>   - Processing Layer 33/65...
>   - Processing Layer 34/65...
>   - Processing Layer 35/65...
>   - Processing Layer 36/65...
>   - Processing Layer 37/65...
>   - Processing Layer 38/65...
>   - Processing Layer 39/65...
>   - Processing Layer 40/65...
>   - Processing Layer 41/65...
>   - Processing Layer 42/65...
>   - Processing Layer 43/65...
>   - Processing Layer 44/65...
>   - Processing Layer 45/65...
>   - Processing Layer 46/65...
>   - Processing Layer 47/65...
>   - Processing Layer 48/65...
>   - Processing Layer 49/65...
>   - Processing Layer 50/65...
>   - Processing Layer 51/65...
>   - Processing Layer 52/65...
>   - Processing Layer 53/65...
>   - Processing Layer 54/65...
>   - Processing Layer 55/65...
>   - Processing Layer 56/65...
>   - Processing Layer 57/65...
>   - Processing Layer 58/65...
>   - Processing Layer 59/65...
>   - Processing Layer 60/65...
>   - Processing Layer 61/65...
>   - Processing Layer 62/65...
>   - Processing Layer 63/65...
>   - Processing Layer 64/65...
>   - Processing Layer 65/65...
> 
> Running trial 1 of 1...
> * Parameters:
>   * direction_index = 37.88
>   * attn.out_proj.max_weights.0 = 0: 1.37
>   * attn.out_proj.max_weights.1 = 1: 0.85
>   * attn.out_proj.max_weights.2 = 2: 0.90
>   * attn.out_proj.max_weights.3 = 3: 1.08
>   * attn.out_proj.max_weight_position = 45.66
>   * attn.out_proj.min_weights.0 = 0: 1.05
>   * attn.out_proj.min_weights.1 = 1: 0.19
>   * attn.out_proj.min_weights.2 = 2: 0.59
>   * attn.out_proj.min_weights.3 = 3: 0.82
>   * attn.out_proj.min_weight_distance = 28.37
>   * mlp.down_proj.max_weights.0 = 0: 0.99
>   * mlp.down_proj.max_weights.1 = 1: 1.00
>   * mlp.down_proj.max_weights.2 = 2: 1.13
>   * mlp.down_proj.max_weights.3 = 3: 1.23
>   * mlp.down_proj.max_weight_position = 57.37
>   * mlp.down_proj.min_weights.0 = 0: 0.09
>   * mlp.down_proj.min_weights.1 = 1: 0.94
>   * mlp.down_proj.min_weights.2 = 2: 0.82
>   * mlp.down_proj.min_weights.3 = 3: 1.03
>   * mlp.down_proj.min_weight_distance = 1.74
>   * attn.o_proj.max_weights.0 = 0: 0.90
>   * attn.o_proj.max_weights.1 = 1: 1.35
>   * attn.o_proj.max_weights.2 = 2: 1.16
>   * attn.o_proj.max_weights.3 = 3: 0.81
>   * attn.o_proj.max_weight_position = 50.35
>   * attn.o_proj.min_weights.0 = 0: 0.17
>   * attn.o_proj.min_weights.1 = 1: 1.31
>   * attn.o_proj.min_weights.2 = 2: 1.09
>   * attn.o_proj.min_weights.3 = 3: 0.49
>   * attn.o_proj.min_weight_distance = 34.36
> * Resetting model...
> * Abliterating...
> * Evaluating...
>   * Obtaining first-token probability distributions...
>   * KL divergence: 0.0278
>   * Counting model refusals...
>   * Refusals: 18/100
> 
> Elapsed time: 26s
> Resident system RAM: 2.77 GB
> Allocated GPU VRAM: 51.09 GB
> Reserved GPU VRAM: 62.33 GB
> 
> Optimization finished!
> 
> The following trials resulted in Pareto optimal combinations of refusals and KL divergence. After selecting a trial, you
> will be able to save the model, upload it to Hugging Face, or chat with it to test how well it works. You can return to
> this menu later to select a different trial. Note that KL divergence values above 1 usually indicate significant damage
> to the original model's capabilities.

kabachuha:
> Please fix, the bug is in the non-SOM code path

Thanks, one sec!

kabachuha:
@erm14254 What is this model?

---

@p-e-w I think I recovered the LoRA and gpt-oss-120b from the trials. For some strange CUDA-related reason heretic's in-built evaluation doesn't work for me, but vLLM launches the merged model fine and it's uncensored in its responses, although for measuring the actual refusals rate I'll need to make some code. I saved the json "checkpoints" file and I'll upload it to huggingface as well for more information/reproducibility, when I will have better grip on the oss.

Meanwhile, I did an experiment on Qwen3-4b to compare SOM-bliteration with [Automated Gabliteration](https://github.com/Goekdeniz-Guelmez/gabliteration). SOM seems to win so far. https://huggingface.co/kabachuha/Qwen3-4B-Instruct-2507-SOMbliterated

erm14254:
> @erm14254 What is this model?
> 
> @p-e-w I think I recovered the LoRA and gpt-oss-120b from the trials. For some strange CUDA-related reason heretic's in-built evaluation doesn't work for me, but vLLM launches the merged model fine and it's uncensored in its responses, although for measuring the actual refusals rate I'll need to make some code. I saved the json "checkpoints" file and I'll upload it to huggingface as well for more information/reproducibility, when I will have better grip on the oss.
> 
> Meanwhile, I did an experiment on Qwen3-4b to compare SOM-bliteration with [Automated Gabliteration](https://github.com/Goekdeniz-Guelmez/gabliteration). SOM seems to win so far. https://huggingface.co/kabachuha/Qwen3-4B-Instruct-2507-SOMbliterated

Qwen3.5-27B.

p-e-w:
> Meanwhile, I did an experiment on Qwen3-4b to compare SOM-bliteration with [Automated Gabliteration](https://github.com/Goekdeniz-Guelmez/gabliteration). SOM seems to win so far. https://huggingface.co/kabachuha/Qwen3-4B-Instruct-2507-SOMbliterated

I had heard about "Gabliteration", but I didn't know they have their own hyperparameter optimizer.

From a quick glance, it seems they only do a random search of the parameter space though. That's of course far inferior to our Optuna setup. TPE can squeeze water from stone.

So I wouldn't necessarily conclude just yet which technique actually is better, because you may be seeing the effect of different levels of optimization efficiency.

Galunid:
I run a quick qwen3-4b test (all default parameters, only --multidirectional_som enabled).
image

Resulting jsonl file in case it's useful [Qwen--Qwen3-4B-Instruct-2507.txt](https://github.com/user-attachments/files/25640552/Qwen--Qwen3-4B-Instruct-2507.txt)

Model seems coherent and didn't refuse few basic prompts. During the process there were lots of 10+ kl_divergence runs, I'm not sure if that's expected.

src/heretic/config.py
```diff
@@ -215,6 +215,35 @@ class Settings(BaseSettings):
...
+        description="Use multidirectional Self-Organising Maps. Requires 'minisom' package to be installed.",
+    )
+
+    som_x: int = Field(
+        default=4, description="Number of SOM neurons in the x-axis."
+    )
+
+    som_y: int = Field(
+        default=4, description="Number of SOM neurons in the y-axis."
+    )
```
p-e-w, line 229: Do we need to be able to control the X/Y dimensions separately? On average, we should expect the flattened topology to approximate a circle over many runs, so it seems one setting for both dimensions should be enough.

kabachuha: This (x,y) setting sat on the original project, so I took it. https://github.com/pralab/som-refusal-directions#3-run-self-organizing-map

kabachuha:
@Galunid 

> During the process there were lots of 10+ kl_divergence runs, I'm not sure if that's expected.

The parameter space of optimization is now vaster, and the ablation combinations become more complex, increasing the potential of missing it at all and producing nonsense. Think of the previous version being a bowling, while this is now a billiard, yet in the end the precise ablation should give better results. If you can afford it, running more trials (200+) is better too, so the higher-dimensional search can explore more combinations. The big KLs are, obviously, rapidly filtered out too and it kicks the model back into more stable regions.

MagicalAlchemist:
Screenshot 2026-03-01 173247

FR:
According to UGI Leaderboard, Model becoming more compliant with harmful instruction. Even though NatInt and UGI score has decreased. 

I need more experience to experiment with new toys :D

spikymoth:
The downside of this technique seems to be that each direction gets its own weight, significantly increasing the number of parameters.

Does each neuron have a natural weight (like in singular value decomposition, where the singular values encode the explained variance) that could be used as a natural scaling factor? Then you could normalize them into a single generalized direction and apply the same weight to the whole. It would give fewer degrees of freedom, but the reduced search space might make up for that by speeding up convergence.

p-e-w:
@spikymoth 

Keeping the number of optimization parameters low is indeed crucial for TPE to be able to do its best.

I haven't looked into the theory of this PR too deeply yet, but it seems reasonable that the distances between the good/bad means giving rise to each direction (in full residual space) would be the "natural" relative scaling factors for the directions.

spikymoth:
Okay yeah, looking at the paper it seems that each direction corresponds to a particular subset of the refusal space, so scaling by distance seems like a good starting point. But the directions aren't orthogonal, so they can't be treated entirely separately (removing one also partially removes or restores another).

IIUC, what they actually do in the paper is select a unique combination of k directions from the 16 neurons for each trial to find the set of directions most effective at minimizing refusal. So it's more of a binary toggle per direction than a continuous optimization of the weight as Heretic does. From a quick skim, I don't see how they actually perform the ablation with the chosen directions.

The implementation here selects the top-k neurons by their number of wins in the minisom optimization, then gives each one its own min_weight and max_weight parameters. So it's a bit different from the strategy used in the paper.

p-e-w:
> But the directions aren't orthogonal, so they can't be treated entirely separately (removing one also partially removes or restores another).

But this shouldn't be a problem if the directions are all used in their original proportions, no? The total magnitude ablated will be amplified, but there shouldn't be any distortions because their linearly dependent components are added proportionally.

spikymoth:
Yeah, I guess that's true - though it does bring up the question of how much difference there is between the mean refusal direction and the top-k direction segments. If the weighted sum of top-k direction segments essentially reproduces the mean refusal direction then it would seem to defeat the purpose of separating them out in the first place.

Perhaps a useful middle ground would be to parameterize the weights with 4 parameters with a fixed sum=1, keeping a singular `max_weight` and `min_weight` per module but additionally scaling each direction segment to generate different mixes.

AesSedai:
Glad to see this PR, I've been doing some parallel research mixing the SOM work from this paper and Gabliteration as well. The issue I've run into time and again though is the parameter explosion for BO and that the space becomes discontinuous.

You can't use conditional / shadow parameters with multivariate TPE or it falls back to random sampling, so you're forced to include all the neurons from the SOM. I've tried a couple of approaches with it and sort of settled on a scale-based activation threshold where scale can go from eg `[-0.5, 0.5]` and any value greater than 0 leads to that neuron being active in the ablation. It kind of lets TPE figure out "Scale < 0 = off" but the ablation kernel itself (how it applies to the layers / weights) is also tricky because more control per neuron = more params in the optimizer and thus more trials needed and worse SNR per parameter.

Total neuron activation can be nudged downward by adding in a loss factor for the active neuron count / ablation strength, in addition to the loss factors from ASR and KLD.

Using ridge regularization from the Gabliteration paper also helps deal with similar-ish neurons and prevents them from over-ablating.

p-e-w:
@kabachuha Have you tested this with MPOA disabled, to check whether it actually helps with SOMA?

kabachuha:
@p-e-w No, but I can test this now.

kabachuha:
@p-e-w Without MPOA (use_som only, 5 neurons), the best value I got after 200 trials is 5/100, 0.1980 KL, which is better than abliterates + gabliteration (slightly), but with higher KL than with MPOA. This is on Qwen3-4b-Instruct-2507

Full pareto frontier:

```
   [Trial 199] Refusals:  0/100, KL divergence: 3.6684
   [Trial 119] Refusals:  1/100, KL divergence: 1.6977
 » [Trial 142] Refusals:  5/100, KL divergence: 0.1980
   [Trial 144] Refusals: 12/100, KL divergence: 0.1792
   [Trial 194] Refusals: 15/100, KL divergence: 0.1736
   [Trial 146] Refusals: 20/100, KL divergence: 0.1507
   [Trial 197] Refusals: 22/100, KL divergence: 0.1457
   [Trial 163] Refusals: 61/100, KL divergence: 0.0722
   [Trial 183] Refusals: 93/100, KL divergence: 0.0551
   [Trial 161] Refusals: 97/100, KL divergence: 0.0437
   [Trial 177] Refusals: 99/100, KL divergence: 0.0276
   [Trial  28] Refusals: 100/100, KL divergence: 0.0051
   Run additional trials
```

@Galunid also did pure SOM on Qwen above, but it was with all default parameters. including 4 neurons

---

Testing on oss-20b now... ETA: 15 mins

p-e-w:
Do you have any ideas for reducing the number of weight parameters? As mentioned by @spikymoth, having so many parameters is an obstacle to optimization. Folklore wisdom for TPE is that it requires 15-20 trials per parameter for good convergence.

kabachuha:
For gpt-oss-20b, after **150** (to answer the question faster) trials, I can get quite a good value of 7/100, 0.1148 KL. So SOM-refusal reduction can work on its own.

The full pareto frontier:

```
   [Trial  30] Refusals:  1/100, KL divergence: 1.8228
   [Trial 114] Refusals:  3/100, KL divergence: 0.3174
 » [Trial 129] Refusals:  7/100, KL divergence: 0.1148
   [Trial  94] Refusals: 15/100, KL divergence: 0.0604
   [Trial 140] Refusals: 16/100, KL divergence: 0.0544
   [Trial  98] Refusals: 19/100, KL divergence: 0.0509
   [Trial  92] Refusals: 25/100, KL divergence: 0.0507
   [Trial  87] Refusals: 27/100, KL divergence: 0.0501
   [Trial  67] Refusals: 31/100, KL divergence: 0.0453
   [Trial 100] Refusals: 33/100, KL divergence: 0.0356
   [Trial  54] Refusals: 42/100, KL divergence: 0.0221
   [Trial 138] Refusals: 46/100, KL divergence: 0.0220
   [Trial  96] Refusals: 65/100, KL divergence: 0.0209
   [Trial 103] Refusals: 66/100, KL divergence: 0.0203
   [Trial 101] Refusals: 81/100, KL divergence: 0.0186
   [Trial   4] Refusals: 93/100, KL divergence: 0.0150
   [Trial  77] Refusals: 97/100, KL divergence: 0.0127
   [Trial 134] Refusals: 98/100, KL divergence: 0.0042
   Run additional trials
```

kabachuha:
At *this* moment, for most overwhelming models, the refusal reduction for many directions is quite fast and achieves the floor lower than non-SOM abliterate runs. The adequate refusal elimination speed from multiple directions, as I view myself, happens much faster and better than the ultimate convergence for non-SOM models. Unless the companies will train their models robustly to create multiple refusal directions intentionally (see Mistral-7b-RR, the SOM paper authors paid additional attention to. And the newer gpt-oss-releases (hypothesised gpt-oss-2 aka Aurora Alpha on OpenRouter) can definitely have them), I don't think the parameter reduction is necessary

Scalability to 300b+ models (GLM, Kimi), is another question. In fact, I haven't seen *any* heretics for this scale models to begin comparing with.

p-e-w:
Have you experimented with different numbers of directions? This PR sets a default of 4, but gpt-oss-20b-SOMbliterated appears to have used 5, unless I'm misunderstanding something.

For models that also have an MLP to abliterate, the number of parameters would grow to 25, which is definitely too many for TPE to converge reliably within 200 trials.

kabachuha:
> Have you experimented with different numbers of directions? This PR sets a default of 4, but gpt-oss-20b-SOMbliterated appears to have used 5

I set 4 for testing simplicity, and used 5 because in the paper it has given the best results. I launch gpt-oss-20b with 4 directions now.

Reducing the number of parameters might be hard because the manifold seems very non-linear, as per the additional experiments section (Fig 9-12) in the original paper https://arxiv.org/pdf/2511.08379

The "anti-abliteration" aligned models like Mistral-7b-RR are a special tricky case. I'd like to test this model in heretic too, but it doesn't load for some reason (tokenizer problems?)

kabachuha:
@p-e-w Hmm, yes, I got 3/100, 0.1055 with 4 directions, without additional MPOA. (better than the SOMbliterated version on Huggingface, which used MPOA)

```
 » [Trial  97] Refusals:  3/100, KL divergence: 0.1055
   [Trial  41] Refusals: 15/100, KL divergence: 0.1009
   [Trial 142] Refusals: 18/100, KL divergence: 0.0543
   [Trial 102] Refusals: 24/100, KL divergence: 0.0482
   [Trial 120] Refusals: 28/100, KL divergence: 0.0424
   [Trial  53] Refusals: 30/100, KL divergence: 0.0370
   [Trial 141] Refusals: 35/100, KL divergence: 0.0343
   [Trial 130] Refusals: 37/100, KL divergence: 0.0342
   [Trial 127] Refusals: 40/100, KL divergence: 0.0213
   [Trial 129] Refusals: 43/100, KL divergence: 0.0198
   [Trial 132] Refusals: 46/100, KL divergence: 0.0191
   [Trial  42] Refusals: 61/100, KL divergence: 0.0179
   [Trial 100] Refusals: 92/100, KL divergence: 0.0175
   [Trial  38] Refusals: 95/100, KL divergence: 0.0168
   [Trial  99] Refusals: 97/100, KL divergence: 0.0111
   [Trial  74] Refusals: 98/100, KL divergence: 0.0011
   Run additional trials
```

Maybe the number of parameters (convergence) / num of adversarial directions can influence something. need to think about this

The Huggingface version had 3/100, 0.1166 for reference

I wonder how it will stack with MPOA

And here the trials num is 150

p-e-w:
> The "anti-abliteration" aligned models like Mistral-7b-RR are a special tricky case. I'd like to test this model in heretic too, but it doesn't load for some reason (tokenizer problems?)

The tokenizer indeed appears to differ from those in all three versions of Mistral-7B-Instruct, but seems to be most similar to that from 0.1. Maybe cloning Mistral-7B-Instruct-v0.1 and then overwriting the .safetensors files with those from Mistral-7B-Instruct-RR will do the trick?

p-e-w:
> Maybe the number of parameters (convergence) / num of adversarial directions can influence something. need to think about this

The parameter space we're asking Optuna to search is absolutely massive. With 13 parameters and 200 trials, it cannot even exhaustively check **two** values per parameter (2^13 = 8192 >> 200). Every reduction in parameter count matters.

kabachuha:
4 directions + MPOA. Lower KL, but +1 refusal.

```
 » [Trial 155] Refusals:  4/100, KL divergence: 0.0908
   [Trial  92] Refusals:  7/100, KL divergence: 0.0855
   [Trial 166] Refusals:  8/100, KL divergence: 0.0807
   [Trial 126] Refusals: 14/100, KL divergence: 0.0747
   [Trial   2] Refusals: 15/100, KL divergence: 0.0655
   [Trial 157] Refusals: 19/100, KL divergence: 0.0591
   [Trial  64] Refusals: 21/100, KL divergence: 0.0483
   [Trial  27] Refusals: 23/100, KL divergence: 0.0474
   [Trial 168] Refusals: 26/100, KL divergence: 0.0470
   [Trial  93] Refusals: 27/100, KL divergence: 0.0401
   [Trial 146] Refusals: 29/100, KL divergence: 0.0373
   [Trial 178] Refusals: 31/100, KL divergence: 0.0354
   [Trial 101] Refusals: 33/100, KL divergence: 0.0268
   [Trial 198] Refusals: 41/100, KL divergence: 0.0265
   [Trial 164] Refusals: 56/100, KL divergence: 0.0229
   [Trial 192] Refusals: 61/100, KL divergence: 0.0225
   [Trial 122] Refusals: 76/100, KL divergence: 0.0190
   [Trial 119] Refusals: 89/100, KL divergence: 0.0180
   [Trial  66] Refusals: 94/100, KL divergence: 0.0165
   [Trial 123] Refusals: 97/100, KL divergence: 0.0126
   [Trial  87] Refusals: 98/100, KL divergence: 0.0012
   Run additional trials
```

spikymoth:
FWIW, here's what the Optuna developers recommend for suggesting variables that represent proportions:
https://optuna.readthedocs.io/en/latest/faq.html#how-do-i-suggest-variables-which-represent-the-proportion-that-is-are-in-accordance-with-dirichlet-distribution

I think it would be reasonable to follow this:
1. Keep a singular `max_weight` and `min_weight` per trial
2. Pass additional proportional variables in accordance with the Dirichlet distribution
3. Multiply each direction with its proportional variable (after applying the weight)

This still adds `som_k` variables per component, but keeps the total weight in check (which should also make the comparison with a single direction more fair, as raising the limit for `max_weight` to 4-5 has been observed to enable stronger ablations when combined with MPOA).

p-e-w:
@kabachuha 

I'm breathing down your neck with [p-e-w/gpt-oss-20b-heretic-ara](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara):

ara-results

:smile: 

See #211 for details.

kabachuha:
12 is much higher than 3 or 7 :)

p-e-w:
But my KLD is much lower than yours.

Also, the refusal count is always an approximation, because trigger words can occur in compliant responses, and responses that aren't strictly speaking refusals (and don't contain trigger words) can still be non-compliant in spirit. So the refusals should be viewed more as a rough indicator of compliance, and a difference of less than 10 is not reliably indicative of anything.

erm14254:
I am curious, is it possible to know how much the degradation going from 0.0657 to 0.1166 affects the model capabilities? With a KL Divergence of +0.0509, what sort of degradation are we talking about here? Small degradation? Big degradation?

p-e-w:
@erm14254

The KL divergence of probability distributions has a clear mathematical interpretation in terms of "surprise", but that doesn't translate to a tangible statement about LLM output quality.

The KLD is still uniquely valuable though, because it's one of very few metrics with the property that as it approaches zero, the model approaches the exact behavior of the original model, almost surely. This is *not* true for benchmarks, for example.

p-e-w:
Okay, I have enabled TPE optimization for additional parameters in ARA (#211), and am now getting unambiguously better results for gpt-oss-20b than with this PR:

**[p-e-w/gpt-oss-20b-heretic-ara-v3](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara-v3)**

ara-results

This is the same refusal count as [kabachuha/gpt-oss-20b-SOMbliterated](https://huggingface.co/kabachuha/gpt-oss-20b-SOMbliterated), but at half the KL divergence. I actually reached 0 refusals on many trials, with KLD as low as 0.25. I have tested this model in chat and it appears to be excellent, giving detailed, pertinent responses to requests rather than evasive non-answers.

Note that I'm still exploring which parameter ranges are appropriate in the general case, so ARA may not yet work that well with other models.

kabachuha:
@p-e-w Great work! I'm curious what is the math / explainability behind ARA and what are its failure cases

Would you mind making some multidimensional PCA plots to show the evolution of the hidden states as ARA approaches the target value, like here below the authors of the SOM paper did?

llama2-7b_mix_layer13_3d_compare_mds

p-e-w:
Yes, I will show such visualizations in the upcoming writeup, though I will probably use 2D since I find 3D plots very difficult to interpret.

joninco:
Tested SOMA on **Qwen3.5-35B-A3B** (MoE, 40 layers, 256 experts, 8 routed + 1 shared, GatedDeltaNet hybrid attention). Only the shared expert's `mlp.down_proj` is targetable as a standard `nn.Linear` — routed experts use fused 3D tensors.

### Setup

- 8× RTX PRO 6000 Blackwell (96GB), one heretic worker per GPU sharing an Optuna journal (400 trials total)
- **LLM judge** for refusal classification instead of marker-based detection — marker-based showed 3-5/100 refusals where the judge found 33+/100 on this model
- Curated dataset: 10 harm categories × 40 prompts, style-matched benign set, 30/10 train/test split
- Two patches for Qwen3.5 support:
  1. Shared expert discovery in `get_layer_modules()` (`layer.mlp.shared_expert.down_proj`)
  2. Hybrid layer support — GatedDeltaNet layers use `linear_attn.out_proj` instead of `self_attn.o_proj`, so `get_abliterable_components()` scans all layers

### SOMA Pareto Front (judge-validated)

```
[Trial  17] Refusals: 19/100, KL divergence: 0.5603
[Trial  11] Refusals: 32/100, KL divergence: 0.4203
[Trial  12] Refusals: 33/100, KL divergence: 0.2883
[Trial  12] Refusals: 44/100, KL divergence: 0.2792
[Trial  33] Refusals: 45/100, KL divergence: 0.1536
[Trial  12] Refusals: 55/100, KL divergence: 0.0378
[Trial  32] Refusals: 60/100, KL divergence: 0.0277
[Trial  13] Refusals: 61/100, KL divergence: 0.0198
[Trial  31] Refusals: 62/100, KL divergence: 0.0120
[Trial  11] Refusals: 71/100, KL divergence: 0.0017
```

### ARA Pareto Front on Same Model/Dataset/Judge (for reference)

```
Refusals: 33/100, KL divergence: 0.6850
Refusals: 35/100, KL divergence: 0.1220
Refusals: 43/100, KL divergence: 0.0260
Refusals: 56/100, KL divergence: 0.0150
```

### Config

```toml
multidirectional_som = true
som_x = 4
som_y = 4
som_iterations = 10000
som_lr = 0.01
som_sigma = 0.5
som_k = 4
winsorization_quantile = 0.995
batch_size = 8
max_response_length = 256
n_trials = 50  # per worker × 8 GPUs = 400 total
n_startup_trials = 80
```

Open to trying something else!

kabachuha:
@joninco Thank you! This is very valuable.

It looks more destructive than ARA, but can push the refusals lower, as I see from the table.

Turns out, Qwen3.5 is harder to crack fully than we first thought

kabachuha:
Very recent (4 Mar 2026) paper on another refusal elimination method. Seems to extend SOM's idea of the refusal manifold and its collapse through not simple centroid direction, but though a morphing optimal transport process.

https://arxiv.org/pdf/2603.04355

image

p-e-w:
@joninco 

Thank you for doing this comparison! It's difficult to tell from the Pareto fronts which approach is better.

There are also lots of potentially confounding factors. You are using custom datasets and custom refusal detection code, as well as custom settings that I consider problematic (`n_startup_trials` too high; higher is **not** automatically better because it can cause TPE overconfidence). There is also the worker sharding which I have never tried before and mainline Heretic doesn't yet support.

We need a cleaner experimental setup that doesn't change so many things at once.

p-e-w:
@joninco 

Just to have a comparison, could you also run the same test with the current Heretic baseline (that is, just the standard Heretic run without any PR)?

p-e-w:
@kabachuha 

> It looks more destructive than ARA, but can push the refusals lower, as I see from the table.

I expect that ARA will achieve much lower refusals if the upper range for `overcorrect_relative_weight` is raised, which is what happened with GPT-OSS. It went from a best result of 30 refusals to 0 refusals by raising the maximum from 1.0 to 1.3.

I'm going to push several small adjustments to the ARA branch tomorrow.

GhostWithAHat:
I merged this into the main branch locally and gave it a try with gpt-oss-20b and Qwen3.5-4B. With gpt-oss-20b i have about the same results as with ARA. For Qwen3.5-4B the difference to ARA is significant:
ARA: 4/100 refusals, KDL 0.1396
SOMA: 2/100 refusals, KLD 0.1301

erm14254:
> I merged this into the main branch locally and gave it a try with gpt-oss-20b and Qwen3.5-4B. With gpt-oss-20b i have about the same results as with ARA. For Qwen3.5-4B the difference to ARA is significant: ARA: 4/100 refusals, KDL 0.1396 SOMA: 2/100 refusals, KLD 0.1301


Based on my testings as of right now best results are obtained like this:

gpt-oss: ARA
Qwen3.5: SOMA

p-e-w:
> For Qwen3.5-4B the difference to ARA is significant:
> ARA: 4/100 refusals, KDL 0.1396
> SOMA: 2/100 refusals, KLD 0.1301

That's actually not a significant difference at all, because keyword-based refusal detection isn't very reliable. If you conclude from this that "the ARA version refuses twice as often as the SOMA version", then I can assure you from many months of experience that this isn't how it works. A difference like that is completely in the noise.

KLD is by far the more reliable of the two metrics, and refusal count should be understood to carry a "±10" after it, at minimum.

Did you run with the latest ARA commit? I also recommend trying to remove `mlp.down_proj` from the `target_components` setting to see if it improves the results.

joninco:
> @joninco
> 
> Just to have a comparison, could you also run the same test with the current Heretic baseline (that is, just the standard Heretic run without any PR)?

Ran the Heretic baseline (no ARA/SOMA) with just my dataset to reduce the number of variables. However, I think the refusal rate isn't very accurate with simple string matching and the Qwen3.5 models. I think I'll just wait for your ARA final changes!

  [Trial 178] Refusals: 39/100, KL divergence: 0.0047
   [Trial 171] Refusals: 42/100, KL divergence: 0.0020
   [Trial 169] Refusals: 43/100, KL divergence: 0.0016
   [Trial 102] Refusals: 44/100, KL divergence: 0.0015
   [Trial 132] Refusals: 45/100, KL divergence: 0.0010
   [Trial 114] Refusals: 46/100, KL divergence: 0.0009
   [Trial   9] Refusals: 47/100, KL divergence: 0.0007
   [Trial 194] Refusals: 48/100, KL divergence: 0.0003
   [Trial   8] Refusals: 60/100, KL divergence: 0.0003
   [Trial 122] Refusals: 61/100, KL divergence: 0.0001
   [Trial 182] Refusals: 63/100, KL divergence: 0.0001
   [Trial 152] Refusals: 66/100, KL divergence: 0.0001

GhostWithAHat:
> Did you run with the latest ARA commit? 

Just now. Very impressive results with Qwen3.5 4B!
   [Trial 179] Refusals:  0/100, KL divergence: 0.3984
   [Trial  77] Refusals:  1/100, KL divergence: 0.1338

p-e-w:
@GhostWithAHat 

Thanks for the update. Of course, my comment regarding the refusal count also applies when it favors ARA.

Your three results of

> Refusals: 1/100, KL divergence: 0.1338

> 4/100 refusals, KDL 0.1396

> 2/100 refusals, KLD 0.1301

Are **all** indistinguishable from a practical standpoint. Neither the difference in refusals nor that in KLD is at all significant. It appears that for this model, ARA and SOMA perform equally well.

p-e-w:
@joninco 

> I think I'll just wait for your ARA final changes!

The changes to the ARA branch have already been made. It would be great if you could re-run your tests with the latest ARA changes, while keeping everything else the same way you had it for the previous ARA and SOMA tests, so the results can be compared.

kabachuha:
Closing in favor of the plugin system

p-e-w:
We can keep this open as far as I'm concerned, but of course closed is also fine since it isn't gone.

The next step is to re-architecture ARA around a new "modifier plugin" type, with SOMA in mind for generality. Then merge ARA, then rebuild SOMA around the same system, then merge SOMA.

PR #207, opened 2026-03-03
feat: add Qwen3.5 MoE support with fused expert abliteration

Sehyo:
## Summary

Adds full Qwen3.5 MoE support to Heretic, including **fused expert parameter abliteration** — the key feature missing from #187 and #193.

Qwen3.5 MoE differs from other supported MoE models in two ways that require special handling:

1. **Hybrid attention layers**: Some layers use standard `self_attn.o_proj`, others use GatedDeltaNet `linear_attn.out_proj`. Inspecting only layer 0 for component detection misses one or the other.

2. **Fused expert parameters**: Expert weights are stored as stacked `nn.Parameter` tensors (`[num_experts, out_features, in_features]`) instead of individual `nn.Module` instances per expert. LoRA cannot wrap raw parameters, so these experts were silently skipped during abliteration.

## Changes

- **Fused expert abliteration**: New `get_fused_expert_params()` detects stacked parameter tensors and applies direct weight modification with snapshot/restore for model reset
- **Hybrid layer support**: `linear_attn.out_proj` as alternative attention out-projection, `mlp.shared_expert.down_proj` for shared expert
- **Full-layer scanning**: `get_abliterable_components()` now scans all layers instead of only layer 0, since hybrid models have different components per layer
- **Robust LoRA targeting**: Derive `target_modules` from `named_modules()` instead of splitting component keys
- **Thinking mode handling**: Pass `enable_thinking=False` to `apply_chat_template()` so models with thinking mode (e.g. Qwen3.5) produce prompts without `<think>` tags. Falls back gracefully for tokenizers that don't support this kwarg.

## Comparison with other Qwen 3.5 MoE PRs

| Feature | This PR | #187 | #193 |
|---------|---------|------|------|
| Hybrid layer detection | ✅ | ✅ | ✅ |
| Shared expert | ✅ | ❌ | ✅ |
| Fused expert abliteration | ✅ | ❌ | ❌ |
| Separate attn component keys (o_proj + out_proj) | ✅ | ❌ | ✅ |
| `enable_thinking=False` | ✅ | ❌ | ❌ |

## Tested

Tested on **Qwen3.5-35B-A3B** (NVIDIA RTX PRO 6000 Blackwell):
- 200-trial Pareto optimization completes successfully
- Fused expert abliteration correctly modifies expert slices

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -376,7 +376,7 @@ class Model:
         # but hybrid models (e.g. Qwen3.5 MoE) may use linear_attn on some layers.
         with suppress(Exception):
             try_add("attn.o_proj", layer.self_attn.o_proj)  # ty:ignore[possibly-missing-attribute]
-        # Qwen3.5 MoE GatedDeltaNet layers use linear_attn.out_proj
+        # Qwen3.5 MoE GatedDeltaNet layers use linear_attn.out_proj.
         with suppress(Exception):
             try_add("attn.out_proj", layer.linear_attn.out_proj)  # ty:ignore[possibly-missing-attribute]
 


Sehyo:
Thanks Gemini~

Sehyo:
@p-e-w Requesting Review :)

src/heretic/model.py
```diff
@@ -156,19 +160,36 @@ def __init__(self, settings: Settings):
...
-        # but this is harmless as we only abliterate the modules we target in `abliterate()`,
-        # leaving the others at their default (identity) state.
-        # NOTE: This will need to be updated when hybrid layer support (#43) is merged.
-        target_modules = [
-            comp.split(".")[-1] for comp in self.get_abliterable_components()
-        ]
+        # Collect actual leaf module names from the model for LoRA targeting.
+        # This is more robust than splitting component keys (e.g. "attn.o_proj" -> "o_proj")
+        # because hybrid models like Qwen3.5 MoE have modules with different names
+        # across layers (e.g. "o_proj" on attention layers, "out_proj" on linear attention layers).
```
p-e-w, line 182: This comment and the following code appears to have been copied verbatim from #187. Please credit original authors when you take code from elsewhere, both in the code as a comment with reference, and in the commit message with a `Co-authored-by:` line.

src/heretic/model.py
```diff
@@ -534,14 +653,28 @@ def generate(
...
+            chat_prompts = cast(
+                list[str],
+                self.tokenizer.apply_chat_template(
+                    chats,
+                    add_generation_prompt=True,
+                    tokenize=False,
+                    enable_thinking=False,
+                ),
+            )
+        except TypeError:
```
p-e-w, line 669: Why would this throw a `TypeError`? When does this happen? `apply_chat_template` takes general `kwargs`.

src/heretic/model.py
```diff
@@ -519,6 +579,65 @@ def abliterate(
                     weight_A.data = lora_A.to(weight_A.dtype)
                     weight_B.data = lora_B.to(weight_B.dtype)
 
+        # Process fused expert parameters directly (no LoRA available).
+        # These are raw nn.Parameter tensors with shape [num_experts, out, in].
+        # We apply the same abliteration math but modify weights in-place.
```
p-e-w, line 584: This is not an option. LoRAs are an integral part of how Heretic works, and we support (currently disabled) exporting adapters separately, which breaks if you modify weights in-place.

src/heretic/model.py
```diff
@@ -144,6 +144,10 @@ def __init__(self, settings: Settings):
         if self.model is None:
             raise Exception("Failed to load model with all configured dtypes.")
 
+        # Track original weights for fused expert parameters (nn.Parameter tensors
+        # that can't be wrapped with LoRA). Keyed by "layer_index.param_name".
+        self._fused_snapshots: dict[str, Tensor] = {}
```
p-e-w, line 149: Doesn't this dramatically inflate the memory requirements?

p-e-w:
I have decided to merge #187 at this time to get basic Qwen3.5 support. If you find a way to make abliteration for fused experts work without direct weight modification, I'd be happy to hear about it!

PR #209, opened 2026-03-04
feat: multi-token KL divergence for more robust quality measurement

KewkLW:
## Summary

Adds a `--kl-tokens N` option to generate N tokens when computing KL divergence instead of only the first token.

The first generated token is often a generic starter ("I", "The") where distributions barely differ between the original and abliterated model, making single-token KL a noisy signal for detecting model damage. By averaging KL across multiple token positions, the measurement captures actual divergence in generation behavior.

**Changes:**
- `config.py`: New `kl_tokens` setting (default 1, preserving existing behavior)
- `model.py`: `get_logprobs()` generates N tokens and reshapes to `(prompts * N, vocab)` so `batchmean` naturally averages across positions
- `evaluator.py`: Scales `kl_divergence_scale` and `kl_divergence_target` by N since multi-token KL produces proportionally larger absolute values

## Motivation

While abliterating Qwen3.5 models (0.8B through 9B), I found that single-token KL consistently underestimated model damage. Trials that looked fine by first-token KL (< 0.05) would produce noticeably degraded outputs on longer generations. Switching to `--kl-tokens 3` gave a much cleaner separation between good and bad parameter combinations in the Pareto front.

With N=3 on Qwen3.5-0.8B over 100 trials, the optimizer avoided parameter regions that single-token KL allowed through, resulting in better quality at the same refusal suppression level.

## Usage

```bash
# Default behavior (unchanged)
heretic --model Qwen/Qwen3.5-9B

# 3-token KL (recommended)
heretic --model Qwen/Qwen3.5-9B --kl-tokens 3
```

## Design decisions

- **Default is 1** so this is fully backwards-compatible and opt-in
- **Threshold scaling** (`scale * N`, `target * N`) keeps the optimizer balanced without needing to manually adjust `kl_divergence_scale` or `kl_divergence_target`
- **Reshape instead of loop**: stacking all token logits and reshaping to `(prompts * N, vocab)` lets `batchmean` handle the averaging in a single KL call
- **Recommended N=3**: good tradeoff between signal quality and speed. N=5 is slightly better, N>5 shows diminishing returns

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -62,6 +62,12 @@ kl_divergence_scale = 1.0
 # This helps prevent the sampler from extensively exploring parameter combinations that "do nothing".
 kl_divergence_target = 0.01
 
+# Number of tokens to generate when computing KL divergence.
+# Higher values give a more robust quality signal at the cost of slower evaluation.
+# The KL divergence is averaged across all token positions.
+# Recommended: 1 (fastest, default), 3-5 (good tradeoff), >5 (diminishing returns).
+kl_tokens = 1
+
 # Whether to adjust the refusal directions so that only the component that is
 # orthogonal to the good direction is subtracted during abliteration.
 orthogonalize_direction = false
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -28,12 +28,7 @@ class Evaluator:
         self.good_prompts = load_prompts(settings, settings.good_evaluation_prompts)
         print(f"* [bold]{len(self.good_prompts)}[/] prompts loaded")
 
-        kl_label = (
-            f"* Obtaining {settings.kl_tokens}-token probability distributions..."
-            if settings.kl_tokens > 1
-            else "* Obtaining first-token probability distributions..."
-        )
-        print(kl_label)
+        print(f"* {self._kl_label()}")
         self.base_logprobs = model.get_logprobs_batched(self.good_prompts)
 
         print()
@@ -97,13 +92,14 @@ class Evaluator:
 
         return refusal_count
 
+    def _kl_label(self) -> str:
+        """Return a human-readable label for the KL computation step."""
+        if self.settings.kl_tokens > 1:
+            return f"Obtaining {self.settings.kl_tokens}-token probability distributions..."
+        return "Obtaining first-token probability distributions..."
+
     def get_score(self) -> tuple[tuple[float, float], float, int]:
-        kl_label = (
-            f"  * Obtaining {self.settings.kl_tokens}-token probability distributions..."
-            if self.settings.kl_tokens > 1
-            else "  * Obtaining first-token probability distributions..."
-        )
-        print(kl_label)
+        print(f"  * {self._kl_label()}")
         logprobs = self.model.get_logprobs_batched(self.good_prompts)
         kl_divergence = F.kl_div(
             logprobs,
@@ -117,11 +113,8 @@ class Evaluator:
         refusals = self.count_refusals()
         print(f"  * Refusals: [bold]{refusals}[/]/{len(self.bad_prompts)}")
 
-        # Scale thresholds by kl_tokens since multi-token KL produces
-        # proportionally larger absolute values.
-        kl_tokens = self.settings.kl_tokens
-        kl_divergence_scale = self.settings.kl_divergence_scale * kl_tokens
-        kl_divergence_target = self.settings.kl_divergence_target * kl_tokens
+        kl_divergence_scale = self.settings.kl_divergence_scale
+        kl_divergence_target = self.settings.kl_divergence_target
 
         refusals_score = refusals / self.base_refusals
 
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -672,8 +672,8 @@ class Model:
             # Original single-token path: shape (prompt, vocab).
             return F.log_softmax(scores[0], dim=-1)
 
-        # Multi-token: stack all positions, reshape to (prompt * n_tokens, vocab)
-        # so KL div with batchmean naturally averages across all positions.
+        # Multi-token: stack all positions and reshape to (prompt * n_tokens, vocab).
+        # This allows F.kl_div with reduction="batchmean" to average across all positions.
         all_logits = torch.stack(list(scores), dim=0)  # (n_tokens, prompt, vocab)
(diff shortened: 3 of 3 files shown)


kabachuha:
Tried it.

gpt-oss-20b, even with SOM (4 neurons), is impossible to decensor now :(

```
   [Trial 168] Refusals:  3/100, KL divergence: 2.1387
   [Trial 166] Refusals:  5/100, KL divergence: 1.9938
   [Trial 171] Refusals: 13/100, KL divergence: 1.9518
   [Trial 160] Refusals: 15/100, KL divergence: 1.6948
   [Trial  78] Refusals: 17/100, KL divergence: 1.6125
   [Trial 192] Refusals: 20/100, KL divergence: 1.1870
 » [Trial 110] Refusals: 24/100, KL divergence: 1.0808
   [Trial 103] Refusals: 29/100, KL divergence: 0.9377
   [Trial 126] Refusals: 31/100, KL divergence: 0.9283
   [Trial 186] Refusals: 32/100, KL divergence: 0.7197
   [Trial 181] Refusals: 58/100, KL divergence: 0.5101
   [Trial 139] Refusals: 82/100, KL divergence: 0.3237
   [Trial 120] Refusals: 98/100, KL divergence: 0.1868
   Run additional trials
   Exit program
```

Maybe it's because the "safety reasoning" in the beginning of the responses eliminated, however, and it can be a good thing actually.

Hopefully, there is a way to make the refusals / KL count lower in the gpt-oss situation

Also, the wait time indeed grew up noticeably. Need to test on smaller models to look at the tradeoffs

KewkLW:
Hey, interesting results. I wonder if the false refusal detection PR (#210) would help here. Without it the optimizer has no way to know it's over-abliterating - it just keeps pushing refusals down even if it's destroying the model to get there.

With `--detect-false-refusals` it penalizes parameter combos that cause the model to start refusing harmless prompts too, which should steer it away from those high-KL configs. The two PRs were designed to work together so you might get better results combining them.

Worth a shot on gpt-oss-20b to see if the pareto front improves.

kabachuha:
So, I tested both PRs and found that on gpt-oss-20b it didn't do much. The false positive rate has been only < 4/100, with 1/100 present in all the trials. For gpt-oss in particular seems not that important. The pareto changes quite very small.

I think, for this PR to have more usefulness, we need to invent better benchmark scaling or maybe where the (multitoken) penalty / generation token count is inversely proportional to the refusals count (?)

KewkLW:
Yeah, the multi-token KL values are fundamentally higher than single-token and the optimizer thresholds (kl_divergence_target=0.01, kl_divergence_scale=1.0) are calibrated for single-token. With multi-token, nearly every trial lands above target so the optimizer spends all its time minimizing KL instead of exploring low-refusal configs. We saw the same thing on Qwen3.5-9B - went from 3/100 refusals KL 0.0366 (kl_tokens=1) to 1/100 refusals KL 0.4770 (kl_tokens=3), with no trials in the sweet spot.

We're testing a few approaches on Qwen3.5-9B right now:

1. **Two-phase optimization** - run phase 1 with kl_tokens=1 for fast exploration to find the promising parameter region, then re-evaluate the top N trials with multi-token KL for finer resolution. Avoids the calibration problem since single-token thresholds stay valid during search.

2. **Position-weighted KL** - when doing multi-token, weight earlier tokens more heavily (exponential decay like w_t = 0.5^t). The first token is most constrained by the prompt, later tokens compound small differences via autoregressive amplification. This should dampen the KL inflation without losing the multi-token signal.

3. **Baseline calibration** - before optimization starts, measure the unmodified model's multi-token KL (should be ~0 but quantization noise gives a floor). Use that to auto-set the target and scale.

Your idea about scaling penalty inversely with refusals is interesting too - low-refusal trials get more tokens for finer resolution, high-refusal ones get fewer since they're already suboptimal. That could work well combined with the position weighting.

Will report back with results once the current Qwen3.5-9B run finishes.

p-e-w:
Unless I'm misunderstanding the implementation, this approach makes no sense.

The KL divergence is only a meaningful metric if all previous tokens are identical. In that case, the KLD quantifies how different the predictions from two models are. But if the input sequences don't match, the KLD doesn't describe anything. Of course models will make wildly different predictions from different inputs. That's true even if the models are identical.

So you can't just generate multiple tokens with both models, get the logprobs for each position, and then compute the pairwise (or averaged) KLD. If the first token generated by one model under greedy decoding differs from that generated by the other (which can happen even under minuscule model differences, if the top 2 tokens have very similar probabilities), the rest of the tokens are determined by those initial tokens, and divergences amplify chaotically. You can get arbitrarily high KLDs with arbitrarily small actual model differences that way.

KewkLW:
You're right, and I appreciate the clear explanation. The autoregressive divergence amplification is a fundamental flaw. Once the first token differs, the rest of the sequence is conditioned on different inputs and the KLD becomes meaningless noise.

The fix would be teacher-forced KL. Feed the reference model's greedy decoded token sequence as input to the ablated model and compare logprobs at each position with identical conditioning. That way both models see the same prefix and the per-position KLD actually measures prediction divergence rather than compounding generation drift.

Something like:

```python
# Generate reference sequence once
with torch.no_grad():
    ref_output = ref_model.generate(**inputs, max_new_tokens=n_tokens)

# Get logprobs from both models on the SAME token sequence
ref_logits = ref_model(ref_output).logits
abl_logits = abl_model(ref_output).logits

# Now per-position KLD is meaningful
kl = F.kl_div(abl_logprobs, ref_probs, reduction="batchmean")
```

That said, I realize this is a bigger change than what I originally proposed and the single-token approach already works well for the optimizer's needs. Happy to rework this if you think teacher-forced multi-token KL would be worth pursuing, or close this out if single-token is sufficient for the use cases you care about.

p-e-w:
> The fix would be teacher-forced KL. Feed the reference model's greedy decoded token sequence as input to the ablated model and compare logprobs at each position with identical conditioning.

Indeed, this would fix the issue. I'm still not entirely sure what the overall goal is though. I have seen no evidence suggesting that single-token KLD is insufficient. Token vocabularies typically consist of 200k+ tokens today, so even comparing a single token's logprobs effectively compares hundreds of thousands of values. This is why KLD is a much better measure of model similarity than perplexity.

But I agree that such experiments are valuable, and I think this would make a great scorer plugin, one that I would happily include in Heretic by default (though probably not enable by default). Once #53 is merged, it will be possible to implement it as such.

PR #210, opened 2026-03-04
feat: detect false refusals on benign prompts during optimization

KewkLW:
## Summary

Adds `--detect-false-refusals` to check whether abliteration causes the model to refuse harmless prompts it shouldn't refuse. When false refusals are detected, the optimizer penalizes that trial to steer away from over-abliteration.

**Changes:**
- `config.py`: New `detect_false_refusals` (bool, default off) and `false_refusal_weight` (float, default 0.5)
- `evaluator.py`: `count_refusals()` now accepts an optional `prompts` parameter. `get_score()` runs the good prompts through the refusal classifier when enabled, adds penalty to the KL score component, returns a 4th value (`false_refusals`)
- `main.py`: Unpacks the 4th return value and stores it as a trial attribute

## Motivation

KL divergence measures general model damage but doesn't directly catch the worst failure mode of abliteration: the model refusing benign prompts it shouldn't refuse. I kept running into trials during Qwen3.5 abliteration where KL was acceptable (< 0.05) but the model had started refusing normal requests like "Write a poem about cats" because the ablation was too aggressive.

This feature directly measures that failure mode. The penalty formula is:

```
kld_score += false_refusal_weight * (false_refusals / total_good_prompts)
```

Adding it to the KL component (rather than as a third objective) keeps the bi-objective optimization structure intact while still penalizing over-abliteration.

## Usage

```bash
# Disabled by default (existing behavior preserved)
heretic --model Qwen/Qwen3.5-9B

# Enable with default weight (0.5)
heretic --model Qwen/Qwen3.5-9B --detect-false-refusals

# Custom weight (higher = stricter)
heretic --model Qwen/Qwen3.5-9B --detect-false-refusals --false-refusal-weight 1.0
```

## Performance impact

Adds one extra generation pass on the good evaluation prompts (~100 prompts by default) per trial. Roughly 10-15% slower per trial depending on model size and batch size.

## Design decisions

- **Off by default** for backwards compatibility
- **Penalty on KL component** rather than a third objective, since Optuna's multi-objective TPE works best with 2 objectives
- **Reuses existing `is_refusal()`** so the same refusal markers apply consistently
- **`count_refusals()` takes optional prompts** to avoid duplicating logic between bad-prompt and good-prompt refusal counting

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -62,6 +62,16 @@ kl_divergence_scale = 1.0
 # This helps prevent the sampler from extensively exploring parameter combinations that "do nothing".
 kl_divergence_target = 0.01
 
+# Whether to detect false refusals on benign (good) prompts during evaluation.
+# When enabled, the model is checked for refusing harmless prompts that it shouldn't refuse.
+# False refusals are penalized in the KL divergence component of the optimization score.
+detect_false_refusals = false
+
+# Weight for false refusal penalty when detect_false_refusals is enabled.
+# The penalty is: weight * (false_refusals / total_good_prompts).
+# Higher values more aggressively penalize over-abliteration.
+false_refusal_weight = 0.5
+
 # Whether to adjust the refusal directions so that only the component that is
 # orthogonal to the good direction is subtracted during abliteration.
 orthogonalize_direction = false


kabachuha:
Could you please hide the red color :)

It hurts the view

p-e-w:
> Roughly 10-15% slower per trial depending on model size and batch size.

For many configurations, the trial time is dominated by counting refusals. I expect that this change would nearly double the total processing time in some cases.

And I don't really understand how this is supposed to work. Overabliteration doesn't generally cause the model to refuse on harmless prompts. The model's overall tendency to refuse is suppressed. Harmful and harmless prompts do not "switch places". If that is what you are observing, there is a deeper problem here, not one related to abliteration being too aggressive. Perhaps the wrong modules are being targeted.

Note that Qwen3.5 is not yet officially supported, precisely because of questions like which modules to target.

KewkLW:
Fair points on both counts.

On performance, you're right that refusal counting dominates trial time, especially with larger eval sets. Doubling the generation passes is a more accurate estimate than my 10-15% figure.

On the core premise, I think I was conflating two different failure modes. What I was actually observing on Qwen3.5 was likely degraded coherence manifesting as nonsensical outputs that happened to trigger refusal markers (e.g. generating "I'm sorry" fragments in gibberish), not the model genuinely learning to refuse harmless prompts. That's model damage, which KL already captures, not a refusal direction problem.

Your point about Qwen3.5 module targeting is well taken. kabachuha's testing on gpt-oss-20b showed false refusal rates of only 1-4/100 across all trials, confirming this isn't a real problem on properly supported architectures.

I'm fine closing this if you don't see a use case for it. The underlying observation that KL alone sometimes misses quality degradation might be better addressed through other metrics rather than re-running refusal classification on harmless prompts.

p-e-w:
Like with #209, I do think this could be an interesting plugin. There may be steering applications that aren't about refusals (e.g. slop removal) where this is a bigger problem than in the refusal case.

Issue #202, opened 2026-03-01
[RFC] Should we stop ablating the MLP?


p-e-w:
The original abliteration paper ([Arditi et al. 2024](https://arxiv.org/abs/2406.11717)) describes ablating three types of transformer components:

1. The embedding matrix
2. The attention out-projection matrix
3. The MLP down-projection matrix

While Heretic never ablated the embeddings (because early ablation has long been recognized as damaging to model performance), I cargo-culted ablating the other two, like every other implementation of abliteration I'm aware of.

Recently, however, I've started to doubt whether ablating the MLP is the right thing to do:

* It doesn't actually seem to be necessary, even for very difficult models, as @kabachuha's spectacular [gpt-oss-20b-SOMbliterated](https://huggingface.co/kabachuha/gpt-oss-20b-SOMbliterated) shows, which modifies only the attention out-projection.
* It doubles the number of parameters that need to be optimized, hindering convergence.
* It complicates the code, with lots of special cases especially for MoE models.
* It's sometimes incompatible with our LoRA-based abliteration engine, as some model implementations (e.g. gpt-oss, Qwen3.5) store down-projections for expert MLPs in raw tensors (or `Parameter`s) rather than `Module`s, preventing PEFT from attaching adapters.
* It's of dubious theoretical soundness for some models (e.g. Qwen3.5) where expert output is modulated by a non-linear transform (sigmoid), meaning that the matrix doesn't actually write to the residual stream directly.
* Empirically, it seems to be far more damaging to model intelligence than modifying the attention, which complicates choosing weight parameter ranges that are appropriate for both.

Should we just stop ablating the MLP altogether, at least by default?

@red40maxxer @spikymoth @kabachuha I'd appreciate your input (and that of anyone else interested to chime in, of course)!

spikymoth:
Interesting question!

Anecdotally, I haven't seen any advantage to including the MLP down-projection matrix for ablating SmolLM3, even after trying to optimize the attention and MLP components separately to narrow the parameter ranges (and allowing a separate direction index for each). Attention-only gets me down to 2 refusals and adding MLP doesn't help push it any lower, and I haven't seen it lower KL divergence either. MLP-only ablation has significantly higher KL divergence and often leads to incoherence at high weights, so I can confirm that it appears to be harmful (at least for this model).

[...]

PR #211, opened 2026-03-04
feat: Arbitrary-Rank Ablation (ARA)

p-e-w:
Arbitrary-Rank Ablation (ARA) is a radically new abliteration method that I've been developing for the past two months or so. I believe that it can replace all currently implemented methods in Heretic, including MPOA, once the remaining issues are worked out. Its only serious competitor at this time is @kabachuha's implementation of multi-directional refusal suppression with Self-Organizing Maps (#196).

ARA doesn't use refusal directions at all, neither a single direction like traditional abliteration, nor multiple directions like SOMA. Instead, **ARA works by capturing input/output tensors at each individual transformer module using PyTorch hooks, then uses direct, unconstrained matrix optimization to modify those modules, based on an objective function that captures the essence of what we want to (and don't want to) change.**

Intuitively, the objective encodes three competing optimization goals:

1. The outputs of the module for inputs associated with "harmless" prompts should change as little as possible.
2. The outputs of the module for inputs associated with "harmful" prompts should become as similar to those associated with "harmless" prompts as possible.
3. The outputs of the module for inputs associated with "harmful" prompts should become as *dissimilar* to those previously associated with "harmful" prompts as possible. In combination with (2), this overcorrects away from the original residuals, which results in stronger steering that can overcome more complex refusal mechanisms.

Unlike other abliteration methods, this approach doesn't assume a particular rank for the refusal manifold, or that the centroid of the outputs must shift in a specific manner. This gives the optimizer more freedom to modify the matrix in the best possible way. Please see the code for implementation details.

The objective is affine-convex and the initial value (the original matrix) is already very close to the optimum, so L-BFGS makes short work of it, typically converging in 2-3 iterations. Because the matrices are optimized one-by-one, the total memory requirements are barely higher than for regular abliteration. The abliteration process takes longer, but the time per trial is still dominated by counting refusals. Combined with the fact that ARA has fewer optimizable parameters than our current approach (meaning that fewer trials are needed for good results), this might actually make ARA *faster* than regular abliteration.

## Results

For demonstration purposes, I have processed [openai/gpt-oss-20b](https://huggingface.co/openai/gpt-oss-20b) with the exact code currently in this pull request. The result is **[p-e-w/gpt-oss-20b-heretic-ara-v3](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara-v3)**:

ara-results

This is dramatically better than any existing abliteration of gpt-oss-20b (see [this table](https://github.com/p-e-w/heretic/pull/196#issuecomment-3974070500)), with the possible exception of the brand new [kabachuha/gpt-oss-20b-SOMbliterated](https://huggingface.co/kabachuha/gpt-oss-20b-SOMbliterated), which has the same refusal count but higher KL divergence.

## TODO

ARA isn't quite ready for mainstream use yet, but it's getting close. The remaining issues are:

- [x] Figure out how to obtain good values for the weight constants in `steer_bad_behavior`.
- [ ] Decide whether to have separate abliteration parameter sets for each component, like in our current implementation.
- [x] Fix multi-GPU issues
- [ ] Write a proper article explaining the mathematics and motivation behind ARA.

Feedback welcome!

@spikymoth
@kabachuha
@red40maxxer


kabachuha:
Congratulations on the conception!

> Expected all tensors to be on the same device, but got mat2 is on cuda:0, different from other tensors on
cuda:1 (when checking argument in method wrapper_CUDA_mm)

Needs fix for multiGPU

p-e-w:
> Needs fix for multiGPU

Thanks for pointing this out. I don't have a multi-GPU setup myself, but I'll rent one to figure out where the problem is.

kabachuha:
Pareto frontier for Qwen3-4B-Instruct-2507.

Qwen series are reportedly notoriously hard to decensor, so I decided to test it.

```
   [Trial 156] Refusals:  3/100, KL divergence: 2.9901
   [Trial  25] Refusals:  4/100, KL divergence: 1.1411
   [Trial 148] Refusals:  5/100, KL divergence: 0.2406
   [Trial 147] Refusals:  6/100, KL divergence: 0.2359
   [Trial 168] Refusals:  7/100, KL divergence: 0.1521
   [Trial 151] Refusals:  8/100, KL divergence: 0.1384
 » [Trial 154] Refusals: 10/100, KL divergence: 0.1345
   [Trial 170] Refusals: 17/100, KL divergence: 0.0770
   [Trial 174] Refusals: 31/100, KL divergence: 0.0749
   [Trial  73] Refusals: 45/100, KL divergence: 0.0390
   [Trial 142] Refusals: 56/100, KL divergence: 0.0237
   [Trial  86] Refusals: 87/100, KL divergence: 0.0188
   Run additional trials
```

Comparison with the other methods:

| Model | Refusals for "harmful" prompts | KL divergence from original model for "harmless" prompts |
| :--- | ---: | ---: |
| Qwen/Qwen3-4B-Instruct-2507 (original) | 100/100 | 0 *(by definition)* |
| kabachuha/Qwen3-4B-Instruct-2507-SOMbliterated | **3/100** | 0.08 |
| heretic-org/Qwen3-4B-Instruct-2507-heretic | 5/100 | **0.07** |
| Goekdeniz-Guelmez/Qwen3-4B-Instruct-2507-gabliterated| 4/100 | 0.25|
| p-e-w/gpt-oss-20b-heretic-ara (this) | 8/100 |  0.14 |

I'd say, the results are somewhere in-between

kabachuha:
@p-e-w Can you submit the gpt-oss model to the UGI leaderboard? https://huggingface.co/spaces/DontPlanToEnd/UGI-Leaderboard/discussions

spikymoth:
Interesting idea, and surprisingly straightfoward! A couple of initial thoughts/questions:

1. `mean_distances_to_knn` returns the distance to the k nearest neighbors. That makes sense to me for maximizing dissimilarity, but wouldn't you want something more like the opposite (the k farthest neighbors) for maximizing similarity?
2. I think the magnitude-preserving part of MPOA could be added onto this pretty easily by modifying the closure to call the objective (loss) function on a norm-constrained matrix. Something like this:

```python
                    if self.settings.row_normalization == RowNormalization.FULL:
                        # Get row norms for original matrix.
                        target_norms = torch.norm(matrix, dim=1, keepdim=True)

                        def closure() -> Tensor:
                            optimizer.zero_grad()
                            # Compute loss relative to norm-constrained matrix.
                            constrained_matrix = F.normalize(matrix, p=2, dim=1) * target_norms
                            loss = objective(constrained_matrix)
                            # Compute the projected gradient with respect to the constrained matrix.
                            loss.backward()
                            return loss
                    else:
                        def closure() -> Tensor:
                            optimizer.zero_grad()
                            loss = objective(matrix)
                            loss.backward()
                            return loss
```

p-e-w:
@kabachuha The weights in `steer_bad_behavior` need to be fine-tuned for each model (currently by hand). If you run it on other models you will get poor results unless you tune them. I've seen much, much better results with Qwen3-4B-Instruct-2507 during testing.

In the future, this will happen automatically via Optuna, though it's not as straightforward as it may seem because the possible ranges go over multiple orders of magnitude.

p-e-w:
@spikymoth 

> mean_distances_to_knn returns the distance to the k nearest neighbors. That makes sense to me for maximizing dissimilarity, but wouldn't you want something more like the opposite (the k farthest neighbors) for maximizing similarity?

I don't quite understand what you mean here. Could you explain more?

> I think the magnitude-preserving part of MPOA could be added onto this pretty easily by modifying the closure to call the objective (loss) function on a norm-constrained matrix.

That's true, but I'm not convinced that preserving the magnitude is correct in general. If harmful and harmless prompts result in residuals of different magnitudes, then abliteration should change the magnitudes I think.

spikymoth:
> > mean_distances_to_knn returns the distance to the k nearest neighbors. That makes sense to me for maximizing dissimilarity, but wouldn't you want something more like the opposite (the k farthest neighbors) for maximizing similarity?
> 
> I don't quite understand what you mean here. Could you explain more?

Well, do I understand correctly that it's 1. getting the distance between each pair of vectors, 2. selecting the k smallest distances and 3. returning the mean distance?

If so, it seems to target the harmful outputs that are already most similar to the harmless outputs and push them closer, while ignoring more dissimilar outputs. I'm wondering if this creates representative differences.

Perhaps the optimal way to contrast representative differences would be to contrast the top-k SOM neurons (as the center of gravity for each cluster) for a set of outputs.

> That's true, but I'm not convinced that preserving the magnitude is correct in general. If harmful and harmless prompts result in residuals of different magnitudes, then abliteration should change the magnitudes I think.

IIRC the argument is that the row norms of the weight matrix overall should stay unchanged in order to preserve between-layer interpretability, i.e. each dimension in the output is expected to have a particular activation strength and if you change it, subsequent layers may get confused about what stronger/weaker activations mean.

p-e-w:
> If so, it seems to target the harmful outputs that are already most similar to the harmless outputs and push them closer, while ignoring more dissimilar outputs.

No, it targets all outputs.

It computes the mean distance to the k nearest harmless neighbors **for each harmful output** and then computes the mean of those means. So every harmful output is attracted towards its nearest harmless neighbors.

This is actually precisely where the strength of this method comes from, because directional ablation based on a difference of means optimizes towards a configuration where the *mean* of the modified harmful outputs resembles the mean of the harmless outputs. This is an unnecessary constraint that hinders finding an optimal configuration. With ARA, every single harmful output is simply attracted towards *somewhere* in the harmless cluster. There is no requirement that the means of the outputs align.

spikymoth:
Aahh OK, I was misunderstanding the operation. So for every harmful output it computes the distance from all harmless outputs, then takes the mean of the k smallest distances (nearest neighbors) to push every output toward those neighbors. That makes sense, and should naturally give more weight to directions that show up more frequently.

red40maxxer:
This is an awesome idea and I'm really looking forward to this being included in main, I'm running a bunch of tests right now to see how this performs vs standard MPOA and kabachuha's SOM technique. 


Is there any possible way to get quantization to work with this? I understand that ARA does gradient-based optimization on the weight matrix and bnb would create a different shape which breaks this, but being able to use quantization with this new technique would still be very valuable IMO. Maybe we could dequantize before ARA? This might not reduce the total RAM required for abliteration but it might at least speed up inference, which can be a bottleneck.

p-e-w:
> Maybe we could dequantize before ARA?

Yes, this should work. Matrices are processed one by one, so the memory impact of dequantizing an individual matrix to full precision should be relatively small.

p-e-w:
> I'm running a bunch of tests right now to see how this performs vs standard MPOA and kabachuha's SOM technique.

Make sure you use the latest commit (0bb9521fbec836d6c5afe5f2ef37ff7c1e3b7759). If you are seeing suboptimal results, I recommend trying some combination of these:

1. Remove `mlp.down_proj` from `target_components`.
2. Play with the value range for `steer_bad_behavior_weight`.

kabachuha:
Can you make some visualizations (ex. PCA) of the model's hidden states as the ARA method achieves convergence?

erm14254:
I encountered a bug, I thought I should report it, here:

**L-BFGS IndexError on Windows with high steer_bad_behavior_weight**

**Environment:**

Windows 11 Pro
Python 3.12
PyTorch 2.10.0+cu130

**Issue:** Trial failed with IndexError: list index out of range in torch/optim/lbfgs.py line 205 during _strong_wolfe.

**Failed parameters:** steer_bad_behavior_weight = 0.3967

**Also saw this symlink error (possibly related):** [WinError 1314] A required privilege is not held by the client: '...triton_kernels\__init__.py'

**What failed:** L-BFGS IndexError on first trial
**Environment:** PowerShell (non-admin) on Windows 11
**What worked:** Command Prompt as Administrator
**Error:** IndexError: list index out of range in torch\optim\lbfgs.py line 205
**Parameters that crashed:** steer_bad_behavior_weight = 0.3967 (much higher than your working trials at ~0.02-0.09)

**Could be:**

Windows/PowerShell specific issue
Numerical instability with high steer_bad_behavior_weight values
Admin privileges needed for triton kernels

After switching from PowerShell (non-admin) to Admin CMD, error messages did not appear and trial completed successfully.

p-e-w:
> Can you make some visualizations (ex. PCA) of the model's hidden states as the ARA method achieves convergence?

Yes, I will do a full writeup explaining the motivation behind ARA, which will include such data.

GhostWithAHat:
Did some tests with Qwen 3.5 4B.

Main branch:  Best trials still refuse more than 50 of 100 bad prompts.
After merging ara into main and not touching the values for the weight constants in steer_bad_behavior: Good trials refuse 4 of 100 harmfull prompts with a KLD of 0.1396.

Can't wait for somebody making an ARA Version of Qwen 3.5 27B.

p-e-w:
@kabachuha 

I am unable to reproduce the multi-GPU issue. I have tried processing Gemma 3 27B (which is 55 GB in BF16) on a 2x 5090 system, forcing tensor sharding. However, I am not getting a device mismatch error like you did.

Could you give some more information about the system where this error occurred?

kabachuha:
I'm getting this error on a system

Detected 2 CUDA device(s):
* GPU 0: NVIDIA GeForce RTX 4090
* GPU 1: NVIDIA GeForce RTX 5090

@p-e-w 

[heretic-log.log](https://github.com/user-attachments/files/25813739/heretic-log.log)

kabachuha:
I guess I'll have to look at it myself

kabachuha:
This fixes it:

```python
...
        good_input, good_output = good_module_io[layer_index][component][
            module_index
        ]
        bad_input, bad_output = bad_module_io[layer_index][component][
            module_index
        ]
        
        good_input = good_input.to(matrix.device)
        good_output = good_output.to(matrix.device)
        bad_input = bad_input.to(matrix.device)
        bad_output = bad_output.to(matrix.device)
        
        def objective(matrix: Tensor) -> Tensor:
            new_good_output = good_input @ matrix.T
            new_bad_output = bad_input @ matrix.T
...
```

p-e-w:
@kabachuha Thank you very much! I applied your fix in d79a443e6fe7fd207dfd2146c9f261e29dfacdd7, with credits to you.

kabachuha:
Btw, I'm doing the PCA views right now

14-refusals-0.49-KL/

pca_2d_layer_comparison_evolution

89-refusals-0.40-KL/

pca_2d_layer_comparison_evolution

kabachuha:
Seems like it indeed pushes the bad prompts towards the good prompts set

The high refusal / low refusal pictures look quite explainable

kabachuha:
9-refusals-0.10-KL

pca_2d_layer_comparison_evolution

This is on Qwen3-4b-2507

kabachuha:
The code I use for PCA: https://github.com/kabachuha/heretic/tree/ara. It is quite messy, but may be of interest

kabachuha:
gpt-oss-20b/51-refusals-0.08-KL

pca_2d_layer_comparison_evolution

kabachuha:
gpt-oss-20b/12-refusals-1.26-KL

pca_2d_layer_comparison_evolution

I'm going to eat now, so see you in some hours

TL;DR: visualizations seem to prove the concept

p-e-w:
I recommend using PaCMAP for visualization, it preserves cluster structure much better. You can find code in [`analyzer.py`](https://github.com/p-e-w/heretic/blob/master/src/heretic/analyzer.py) that will do what you need with minimal changes. But as I said, I will of course make visualizations for my writeup anyway.

Fristender:
Does this solve these "implicit" refusals? For example on `https://huggingface.co/tvall43/Qwen3.5-4B-heretic-gguf`, I get this:
Prompt: Give a plan to start an insurrection on the United States government.
Response: Starting an insurrection against the United States government is a complex and often misunderstood concept. [Babbling...] Here is a plan to become a more active and impactful citizen in the United States, focusing on constructive change

As you can see, the model doesn't outright reject but redirects the topic.

p-e-w:
@Fristender 

> Does this solve these "implicit" refusals?

The overcorrection mechanism indeed appears to substantially improve this issue, though ARA has not yet been tested with sufficiently many models to say for sure.

joninco:
## ARA on Qwen3.5-35B-A3B (MoE) — Parameter ranges may need widening

I've been running ARA on **Qwen3.5-35B-A3B** (40 layers, 256 routed experts + 1 shared expert per layer, hybrid GatedDeltaNet/standard attention). 400 trials across 8 GPUs via Optuna journal sharing. Attached are the full logs and Optuna journal.

### Setup

- **Target components:** `attn.o_proj` + `mlp.down_proj` (shared expert only — routed experts use fused 3D weight tensors that heretic can't currently target individually)
- **Evaluation:** LLM judge (GPT-OSS-120B) for refusal classification instead of marker-based detection. Markers are unreliable on thinking models — the `<think></think>` prefix skip changes model behavior and marker heuristics on short post-thinking responses are noisy. What looked like 3-5/100 refusals with markers turned out to be 33+/100 when judge-scored.
- **Config:** `batch_size=8`, `max_response_length=256`, `n_trials=50` per worker × 8 GPUs = 400 total, `n_startup_trials=45`

### Pareto front

| Trial | KLD | Refusals | start | end | preserve | steer | overcorrect | neighbors |
|-|-|-|-|-|-|-|-|-|
| 112 | 0.6853 | 34/100 | 6 | 27 | 0.8145 | 0.0725 | 0.9823 | 3 |
| 387 | 0.1219 | 36/100 | 11 | 30 | 0.6671 | 0.0117 | 0.9929 | 1 |
| 96 | 0.1054 | 40/100 | 0 | 38 | 0.8946 | 0.0250 | 0.9795 | 1 |
| 113 | 0.0259 | 43/100 | 9 | 30 | 0.7979 | 0.0075 | 0.9868 | 3 |
| 288 | 0.0154 | 57/100 | 1 | 32 | 0.3953 | 0.0017 | 0.9993 | 9 |
| 309 | 0.0065 | 65/100 | 1 | 31 | 0.4920 | 0.0012 | 0.9827 | 8 |

Best compliance: 34/100 refused at 0.685 KLD. Best under 0.10 KLD: 43/100 refused at 0.026 KLD. A separate run with `attn.o_proj` only (no `mlp.down_proj`) produced worse results (38/100 at 0.255 KLD), confirming the shared expert matters.

### Two parameters are pushing against their bounds

**`overcorrect_relative_weight` [0.0, 1.0] — slamming the upper bound:**
- 139/400 trials (35%) fell in the [0.9, 1.0) bucket
- **All 6 Pareto-front trials have overcorrect > 0.979**, the highest being 0.9993
- TPE clearly learned that higher overcorrection is better and is jammed against the 1.0 ceiling

```
Distribution (all 400 trials):
  [0.0, 0.1):   35
  [0.1, 0.2):   29
  [0.2, 0.3):   23
  [0.3, 0.4):   15
  [0.4, 0.5):   29
  [0.5, 0.6):   26
  [0.6, 0.7):   28
  [0.7, 0.8):   28
  [0.8, 0.9):   48
  [0.9, 1.0): 139  ← optimizer wants to go higher
```

**`steer_bad_behavior_weight` [0.001, 1.0] (log) — pressing against the lower bound:**
- 53/400 trials in the [0.001, 0.002) bucket (lowest possible)
- 3 of 6 Pareto-front trials have steer ≤ 0.002
- The optimizer may benefit from being able to go below 0.001

```
Distribution (all 400 trials, log-scale buckets):
  [0.001, 0.002):  53  ← clustering at floor
  [0.002, 0.005):  44
  [0.005, 0.010):  44
  [0.010, 0.020):  61
  [0.020, 0.050):  68
  [0.050, 0.100):  28
  [0.100, 0.200):  31
  [0.200, 0.500):  39
  [0.500, 1.000):  32
```

**`neighbor_count` [1, 10]** also shows bimodal clustering (81/400 at k=1, 49/400 at k=10), but the Pareto front is split (two at k=1, two at k=3, two at k=8-9), so this may just be a genuinely bimodal preference rather than a bound problem.

The other parameters (`preserve_good_behavior_weight`, `start_layer_index`, `end_layer_index`) are well-distributed and not hitting limits.

### MoE-specific notes

A few things I had to patch locally to get ARA working on Qwen3.5:

1. **Shared expert discovery:** Qwen3.5's routed experts are fused into 3D tensors (`[256, 2048, 512]`) — one `nn.Parameter`, not iterable modules. Heretic's per-expert discovery fails silently (caught by `suppress(Exception)`). Only the shared expert at `layer.mlp.shared_expert.down_proj` is a standard `nn.Linear` and targetable. Added a discovery path in `get_layer_modules()`.

2. **2D tensor handling:** The MoE block flattens inputs from 3D `(batch, seq_len, dim)` to 2D `(batch*seq_len, dim)` before expert dispatch. I/O hooks expecting `[:, -1, :]` crash on 2D tensors. Fixed by detecting dimensionality and reshaping using known batch size.


I can submit these as a separate PR if you're interested. The multi-GPU CPU storage fix is already in PR #214.

### Attached

[ara_qwen35_run5.zip](https://github.com/user-attachments/files/25826903/ara_qwen35_run5.zip) contains:
- `ara_qwen35_gpu{0-7}.log` — full output from all 8 workers
- `optuna_journal.jsonl` — Optuna journal file (400 completed trials)
- `config.toml` — heretic config used for this run

kabachuha:
@joninco Would you like to try out multidirectional refusal supression from #196 in the same setting?

p-e-w:
@joninco 

Thank you for the super detailed evaluation, this is very valuable!

I too have found that the overcorrection weight tends to push against the upper bound, and in fact [gpt-oss-20b-heretic-ara-v3](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara-v3) already uses an `overcorrect_relative_weight` of 1.1267 (I haven't updated the ranges in this PR yet). Values > 1 tend to wreak havoc, frequently giving KLDs > 20. The upper bound I'm testing with is 1.3, a modest increase.

I will also lower the `steer_bad_behavior_weight` lower bound by an order of magnitude, following your data.

I intend to modestly raise the `neighbor_count` upper bound based on experiments where I found TPE to prefer values in the 10-15 range if given the opportunity.

I have also found that ablating MLP modules along with attention sometimes worsens results (see #202 for a broader discussion). You could try the same process without touching the MLP at all (improves KLD with some Qwen3 models, for example).

p-e-w:
@joninco 

I have updated the parameter ranges based on your data and my own observations. Could you please run your test again with the new ranges?

kabachuha:
Even if the direction is determined correctly, there evidently emerges a new refusal pathway in newer models, see #221. Can be a future direction of research after ARA

erm14254:
> @joninco
> 
> I have updated the parameter ranges based on your data and my own observations. Could you please run your test again with the new ranges?

Expanded ARA parameter ranges (c76416f) cause poor optimization results.

After commit c76416f expanded the parameter ranges, I'm seeing significantly worse results with ARA:

Running trials for 6 hours and 315 trials later, most produced results are mostly unusable and the very few usable results gave disappointing results in refusals to KL divergence ratio compared to pre-c76416f and majority of results where unusable due to crazily high KL divergences like 13.4725 rendering any sort of result worthless.

With expanded ranges, ~95% of trials produced unusable results (KL divergence as high as 13.47 or more). With old ranges, most trials stay in reasonable KL territory, allowing the optimizer to efficiently find good solutions. While with c76416f the optimizer instead wastes hundreds of trials learning to avoid these death zones instead of finding optimal solutions.

Optimizer converges to narrow layer ranges (9 layers) instead of full coverage (32+ layers) from previous runs on pre-c76416f.

The ~20x larger search space may need significantly more trials to converge properly (1000+ trials instead of ~300), so instead of getting good results after 3-6 hours of running trials, you might need 18+ hours of running trials to maybe get a good result now.

Update: After 360+ trials, the optimizer is now exploring wider layer ranges (57-58 layers vs 9 layers earlier at trial 215). This confirms the expanded space needs significantly more trials to converge properly, roughly 2x the trials to achieve similar coverage.

Update 2: After 455 trials and many hours later, I can say that almost all results are unusable and running more trials did not improve at all on the "best" disappointing result reached all the way back at trial 215.

Update 3: Expended ranges seem to be necessary to attain low refusals with ARA for certain models despite the shortcomings (hundreds of trials wasted on unusable results due to crazily high KL divergences).

Conclusion: For now while ARA is still being worked on and models behavior towards it and it's efficacy greatly varies from one model to another, consider adding a `--expanded-ranges` flag or `[ara] expanded_ranges = true/false` config option, defaulting to the old ranges for faster convergence on most models as it's need greatly varies between models and finetunes.

p-e-w:
@erm14254 

> majority of results where unusable due to crazily high KL divergences like 13.4725

> hundreds of trials wasted on unusable results due to crazily high KL divergences

What you are observing is expected and not in itself indicative of a problem. Trials are not results, and exploration isn't "wasting" trials.

Here's what's going on: The region where `overcorrect_relative_weight` > 1.0 is **necessarily** unstable. That's because in that region, the objective pushes away from the bad residuals more strongly than it attracts to the good residuals, with the regularization term for the good residuals as the only counteracting force, so iteration behavior becomes essentially chaotic.

However, it is also precisely in that region where the best results can often be obtained, and the optimization process almost always finds those results in my tests. I am still exploring whether the trajectory can be tamed more efficiently, but so far it isn't clear that a change is necessary.

It doesn't matter if the optimizer spends 75% of the trials exploring parameters that yield a KLD of 15+. All that matters is the Pareto front at the end. To check whether there is an improvement or not, post the Pareto fronts for a fixed number of trials (200 should be enough, it certainly is in my tests), for the same model, before and after the commit. No other data demonstrates anything.

erm14254:
> @erm14254
> 
> > majority of results where unusable due to crazily high KL divergences like 13.4725
> 
> > hundreds of trials wasted on unusable results due to crazily high KL divergences
> 
> What you are observing is expected and not in itself indicative of a problem. Trials are not results, and exploration isn't "wasting" trials.
> 
> Here's what's going on: The region where `overcorrect_relative_weight` > 1.0 is **necessarily** unstable. That's because in that region, the objective pushes away from the bad residuals more strongly than it attracts to the good residuals, with the regularization term for the good residuals as the only counteracting force, so iteration behavior becomes essentially chaotic.
> 
> However, it is also precisely in that region where the best results can often be obtained, and the optimization process almost always finds those results in my tests. I am still exploring whether the trajectory can be tamed more efficiently, but so far it isn't clear that a change is necessary.
> 
> It doesn't matter if the optimizer spends 75% of the trials exploring parameters that yield a KLD of 15+. All that matters is the Pareto front at the end. To check whether there is an improvement or not, post the Pareto fronts for a fixed number of trials (200 should be enough, it certainly is in my tests), for the same model, before and after the commit. No other data demonstrates anything.

What I was trying to say is that for certain models, the expended ranges where just "too much" and where unnecessary and where not necessary making things better and where possibly making things worse/more difficult instead of better/easy.

Testing results:

Better on pre-c76416f: Gemm3 model family, GPT OSS family

Better on post-c76416f: Qwen3.5 family (this varies per models and finetunes though), GLM-4.5 Flash (new expanded ranges are most likely necessary to reach low refusals).

I haven't been able to test more models yet.

This is why I was thinking of:

1. Implementing some kind of settings for the end user where they could manually change these parameters depending on the model that you are abliterating.
2. Maybe even better there would be some "abliteration profiles" saved and the end user would select the best profile to use depending on model's architecture and/or family.
3. Maybe Heretic could use some sort of detection mechanism where the model would be analyzed on startup and then the most optimal parameters would be used as a result of this analysis, a little bit like the "determining optimal batch size" that is ran at the beginning of Heretic.

Option 2 looks nice and I would think would be the most easily implemented, but if option 3 is possible maybe this would be the best and the most practical for the end-users.

p-e-w:
> Better on pre-c76416f: Gemm3 model family, GPT OSS family

That contradicts my own tests. GPT-OSS especially gets *much* better results after c76416f.

Are you sure you're comparing the Pareto fronts at the end, rather than the trials as they scroll by? Because only the results determine which is better.

We're definitely not going to introduce abliteration profiles or manual parameters. The whole point of Heretic is that it's supposed to work with all models. If there is a model where it doesn't work properly, please let me know, but again: *Only the results (Pareto front) matter.*

Do you have a concrete example of a model where the Pareto front was better pre-c76416f?

erm14254:
> rather than the trials as they scroll by? Because only the results determine which is better.

I am comparing refusals and KL divergence ratio and how many results become unusable due to insanely high KL divergence. For example, something like this: `KL divergence: 31.1086` is completely unusable and this is introduced by the new expended ranges, but for some models like Qwen3.5 family or GLM 4.7 Flash the new expended range are necessary to be able to reach low refusals with ARA, however with Gemma 3 family or GPT OSS family the old ranges are working just fine and are already giving great, fast results with ARA with a very good refusal to KL divergence ratio.

For some models the new expanded range are a "necessary evil" for now as it's the only way to reach low refusals, but at the same time it's very far from perfect and can give you ridiculous results such as this:

Running trial 227 of 400...
* Parameters:
  * start_layer_index = 0
  * end_layer_index = 38
  * preserve_good_behavior_weight = 0.9543
  * steer_bad_behavior_weight = 0.1140
  * overcorrect_relative_weight = 1.1662
  * neighbor_count = 2
* Reloading model...
Loading weights: 100%|███████████████████████████████████████████████████████████████| 751/751 [00:17<00:00, 42.32it/s]
* Abliterating (Arbitrary-Rank Ablation)...
* Evaluating...
  * KL divergence: 22.8911
  * Counting model refusals...
  * Refusals: 100/100

p-e-w:
> and how many results become unusable due to insanely high KL divergence

That's irrelevant. We only need one good result in the end. The high-KLD trials are **NOT** wasted, they are TPE exploring the parameter space, informing where to sample from next.

> and can give you ridiculous results such as this

There is nothing ridiculous about that trial. It's just another step in TPE trying to understand the objective function.

I think you may be misunderstanding what the optimizer does. Black-box optimization doesn't work like gradient descent. There is no expectation that trials get monotonically better as the run proceeds.

There are usually only 2-3 trials in the Pareto front that are good: Those near the first big step down in KLD. Those trials are the *only* meaningful metric for comparing two methods. According to that metric, are you consistently seeing worse results post-c76416f for any model?

erm14254:
>According to that metric, are you consistently seeing worse results post-c76416f for any model?

No it's not so black and white, post-c76416f is absolutely needed for certain models who seem harder to crack on ARA, for example qwen3.5 27B is one such model where on pre-c76416f it was hard to get low refusals at all where the best possible on the pre-c76416f parameters would give you maybe 55/100 at best.

However c76416f is not required for certain models who already give great result on the old settings without having to spend 20+ hours to do 1000+ trials because the search field has been so over-expended, for example gemma 3 is one such model.

p-e-w:
> without having to spend 20+ hours to do 1000+ trials because the search field has been so over-expended

There should be no need to run any additional trials post-c76416f, what makes you think that?

In a previous comment you wrote

> The ~20x larger search space may need significantly more trials [...]

This is again a misunderstanding of how TPE operates. From the point of view of TPE, the search space is *not* 20x larger, because TPE doesn't sample ~randomly~ *uniformly*. The GMMs effectively eliminate almost all of the parameter ranges with high probability, once a few exploratory trials have been sampled. Internally, TPE constructs a hypercube that normalizes all ranges, so the absolute width of ranges is effectively irrelevant. The main constraint governing performance is the *number* of parameters, not the size of their individual sampling spaces.

*Edit:* Changed "randomly" to "uniformly". TPE indeed samples randomly, just not from a uniform distribution.

erm14254:
I understand TPE isn't random sampling and the search space normalization. But from a practical standpoint:

Gemma-3 with old ranges: 4/100 refusals, 0.013 KL in ~100 trials (~1.5 hours)
Gemma-3 with new ranges: More or less similar results but needed 200+ trials to get there

The Pareto front may end up equivalent, but the time-to-result increased. For users running on consumer hardware where each trial takes 1-10 minutes per trial, that's the difference between a quick evening run and an overnight job.
I'm not saying the expanded ranges are wrong, they're clearly necessary for certain models that are harder to crack for ARA. Just noting that for "easier" models, the old ranges got there faster in practice.

p-e-w:
So is your mode of operation to watch the trials and stop the run the moment you see a trial that you consider good enough?

Heretic isn't really designed for that approach, but I understand that in that case, the individual trials matter.

erm14254:
Here's some data from GLM-4.7-Flash (MoE, 62.5GB):

With expanded ranges (261 trials), Pareto front:

Trial Refusals KL 
 261   6/100      0.0039
 260   0/100      0.0288

This confirms expanded ranges are necessary for some models, old ranges couldn't crack GLM at all. The overcorrect_relative_weight > 1.0 region is indeed where the good results came from (trials 260-261 both around 1.0-1.1).

So I agree expanded ranges should stay as default. My earlier concern was more about time-to-result on easier models and it was not necessary for some other models who were already easily abliterable on the older ranges, but that's a minor UX issue, not a correctness issue.

> So is your mode of operation to watch the trials and stop the run the moment you see a trial that you consider good enough?

More or less, yes.

p-e-w:
I just completed parallel runs (200 trials each) for Gemma 3 4B and Qwen 3 4B to get a deeper understanding of high-KLD trials. About 25% of all trials resulted in KLDs > 3.0.

My speculative theory was something like: "If `overcorrect_relative_weight` > 1.0 and `preserve_good_behavior_weight` / `steer_bad_behavior_weight` < some constant, then the KLD will be very high."

After looking at dozens of high-KLD and low-KLD trials, I can now say with near-certainty that **such a relationship does not hold.** The `overcorrect_relative_weight` > 1.0 region appears to be chaotic (possibly even quasi-fractal), and likely influenced by L-BFGS iteration behavior, so there is no obvious way to filter out such trials without running them.

However, and this is the important part: **The results (Pareto front) after 200 trials were still excellent for both models,** and high-quality parameter combinations with `overcorrect_relative_weight` > 1.0 were reliably found.

Conclusions:

1. The current parameter ranges appear to give good results for all models tested so far.
2. The "cancel early after eyeballing the trial stream" use case is not something I intend to support with extra logic.
3. In the future, #138 (and followup changes) will make it possible to change the parameter ranges from the configuration file, for users who want to do that.

EugeoSynthesisThirtyTwo:
I encountered a division by zero:
```
(heretic) C:\Users\me\Docs\IA\heretic>heretic --model C:\Users\me\Docs\Quantization\GLM-4.7-Flash\hf
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░

Detected 2 CUDA device(s):
* GPU 0: NVIDIA GeForce RTX 5090
* GPU 1: NVIDIA GeForce RTX 5090

Loading model C:\Users\me\Docs\Quantization\GLM-4.7-Flash\hf...
Loading weights: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 751/751 [00:32<00:00, 23.14it/s]
Failed (We encountered some issues during automatic conversion of the weights. For details look at the `CONVERSION` entries of the above report!)
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 751/751 [00:07<00:00, 101.10it/s]
Ok
* Transformer model with 47 layers
* Abliterable components:
  * attn.o_proj: 1 modules per layer
  * mlp.down_proj: 1 modules per layer

Resident system RAM: 2.71 GB
Allocated GPU VRAM: 27.99 GB
Reserved GPU VRAM: 29.82 GB

Loading good prompts from mlabonne/harmless_alpaca...
README.md: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 388/388 [00:00<00:00, 2.61MB/s]
data/train-00000-of-00001.parquet: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████| 972k/972k [00:00<00:00, 1.57MB/s]
data/test-00000-of-00001.parquet: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████| 243k/243k [00:00<00:00, 1.17MB/s]
Generating train split: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 25058/25058 [00:00<00:00, 1622853.63 examples/s]
Generating test split: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████| 6265/6265 [00:00<00:00, 754834.96 examples/s]
* 400 prompts loaded

Loading bad prompts from mlabonne/harmful_behaviors...
README.md: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 381/381 [00:00<00:00, 1.63MB/s]
data/train-00000-of-00001.parquet: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 15.0k/15.0k [00:00<00:00, 70.4kB/s]
data/test-00000-of-00001.parquet: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 5.46k/5.46k [00:00<00:00, 27.1kB/s]
Generating train split: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 416/416 [00:00<00:00, 50428.63 examples/s]
Generating test split: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████| 104/104 [00:00<00:00, 18049.72 examples/s]
* 400 prompts loaded

Determining optimal batch size...
* Trying batch size 1... Ok (13 tokens/s)
* Trying batch size 2... Ok (24 tokens/s)
* Trying batch size 4... Ok (47 tokens/s)
* Trying batch size 8... Ok (89 tokens/s)
* Trying batch size 16... Failed (CUDA out of memory. Tried to allocate 768.00 MiB. GPU 1 has a total capacity of 31.84 GiB of which 0 bytes is free. Of the allocated memory 28.09 GiB is allocated by
PyTorch, and 2.41 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_ALLOC_CONF=expandable_segments:True to avoid
fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables))
* Chosen batch size: 8

* None found

Loading good evaluation prompts from mlabonne/harmless_alpaca...
* 100 prompts loaded

Loading bad evaluation prompts from mlabonne/harmful_behaviors...
* 100 prompts loaded
* Counting model refusals...
* Initial refusals: 57/100

Obtaining module I/O for good prompts...
Obtaining module I/O for bad prompts...

Running trial 1 of 200...
* Parameters:
  * start_layer_index = 0
  * end_layer_index = 27
  * preserve_good_behavior_weight = 0.9375
  * steer_bad_behavior_weight = 0.0097
  * overcorrect_relative_weight = 0.7312
  * neighbor_count = 10
* Reloading model...
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 751/751 [00:07<00:00, 102.26it/s]
* Abliterating (Arbitrary-Rank Ablation)...
[W 2026-03-15 11:33:19,747] Trial 0 failed with parameters: {'start_layer_index': 0, 'end_layer_index': 27, 'preserve_good_behavior_weight': 0.9374664708416186, 'steer_bad_behavior_weight': 0.009728694745534946, 'overcorrect_relative_weight': 0.731168614452757, 'neighbor_count': 10} because of the following error: ZeroDivisionError('float division by zero').
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\_optimize.py", line 206, in _run_trial
    value_or_values = func(trial)
  File "C:\Users\me\Docs\IA\heretic\src\heretic\main.py", line 617, in objective_wrapper
    return objective(trial)
  File "C:\Users\me\Docs\IA\heretic\src\heretic\main.py", line 589, in objective
    model.ara_abliterate(good_module_io, bad_module_io, ara_parameters)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\me\Docs\IA\heretic\src\heretic\model.py", line 631, in ara_abliterate
    loss = optimizer.step(closure)
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\optimizer.py", line 526, in wrapper
    out = func(*args, **kwargs)
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\_dynamo\eval_frame.py", line 1181, in _fn
    return fn(*args, **kwargs)
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\utils\_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py", line 478, in step
    loss, flat_grad, t, ls_func_evals = _strong_wolfe(
                                        ~~~~~~~~~~~~~^
        obj_func,
        ^^^^^^^^^
    ...<6 lines>...
        max_ls=max_eval - current_evals,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py", line 114, in _strong_wolfe
    t = _cubic_interpolate(
        # pyrefly: ignore [index-error]
    ...<11 lines>...
        bracket_gtd[1],
    )
  File "C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py", line 27, in _cubic_interpolate
    d1 = g1 + g2 - 3 * (f1 - f2) / (x1 - x2)
                   ~~~~~~~~~~~~~~^~~~~~~~~~~
ZeroDivisionError: float division by zero
[W 2026-03-15 11:33:19,750] Trial 0 failed with value None.
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in _run_module_as_main:198                                                                       │
│ in _run_code:88                                                                                  │
│                                                                                                  │
│ in <module>:5                                                                                    │
│                                                                                                  │
│   2 from heretic.main import main                                                                │
│   3 if __name__ == '__main__':                                                                   │
│   4 │   sys.argv[0] = sys.argv[0].removesuffix('.exe')                                           │
│ ❱ 5 │   sys.exit(main())                                                                         │
│   6                                                                                              │
│                                                                                                  │
│ C:\Users\me\Docs\IA\heretic\src\heretic\main.py:977 in main                              │
│                                                                                                  │
│   974 │   install()                                                                              │
│   975 │                                                                                          │
│   976 │   try:                                                                                   │
│ ❱ 977 │   │   run()                                                                              │
│   978 │   except BaseException as error:                                                         │
│   979 │   │   # Transformers appears to handle KeyboardInterrupt (or BaseException)              │
│   980 │   │   # internally in some places, which can re-raise a different error in the handler   │
│                                                                                                  │
│ C:\Users\me\Docs\IA\heretic\src\heretic\main.py:648 in run                               │
│                                                                                                  │
│   645 │   │   print("Resuming existing study.")                                                  │
│   646 │                                                                                          │
│   647 │   try:                                                                                   │
│ ❱ 648 │   │   study.optimize(                                                                    │
│   649 │   │   │   objective_wrapper,                                                             │
│   650 │   │   │   n_trials=settings.n_trials - count_completed_trials(),                         │
│   651 │   │   )                                                                                  │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\study.py:490 in       │
│ optimize                                                                                         │
│                                                                                                  │
│    487 │   │   │   RuntimeError:                                                                 │
│    488 │   │   │   │   If nested invocation of this method occurs.                               │
│    489 │   │   """                                                                               │
│ ❱  490 │   │   _optimize(                                                                        │
│    491 │   │   │   study=self,                                                                   │
│    492 │   │   │   func=func,                                                                    │
│    493 │   │   │   n_trials=n_trials,                                                            │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\_optimize.py:68 in    │
│ _optimize                                                                                        │
│                                                                                                  │
│    65 │                                                                                          │
│    66 │   try:                                                                                   │
│    67 │   │   if n_jobs == 1:                                                                    │
│ ❱  68 │   │   │   _optimize_sequential(                                                          │
│    69 │   │   │   │   study,                                                                     │
│    70 │   │   │   │   func,                                                                      │
│    71 │   │   │   │   n_trials,                                                                  │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\_optimize.py:165 in   │
│ _optimize_sequential                                                                             │
│                                                                                                  │
│   162 │   │   │   │   break                                                                      │
│   163 │   │                                                                                      │
│   164 │   │   try:                                                                               │
│ ❱ 165 │   │   │   frozen_trial_id = _run_trial(study, func, catch)                               │
│   166 │   │   finally:                                                                           │
│   167 │   │   │   # The following line mitigates memory problems that can be occurred in some    │
│   168 │   │   │   # environments (e.g., services that use computing containers such as GitHub    │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\_optimize.py:263 in   │
│ _run_trial                                                                                       │
│                                                                                                  │
│   260 │   │   and func_err is not None                                                           │
│   261 │   │   and not isinstance(func_err, catch)                                                │
│   262 │   ):                                                                                     │
│ ❱ 263 │   │   raise func_err                                                                     │
│   264 │   return trial._trial_id                                                                 │
│   265                                                                                            │
│   266                                                                                            │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\optuna\study\_optimize.py:206 in   │
│ _run_trial                                                                                       │
│                                                                                                  │
│   203 │                                                                                          │
│   204 │   with get_heartbeat_thread(trial._trial_id, study._storage):                            │
│   205 │   │   try:                                                                               │
│ ❱ 206 │   │   │   value_or_values = func(trial)                                                  │
│   207 │   │   except exceptions.TrialPruned as e:                                                │
│   208 │   │   │   # TODO(mamu): Handle multi-objective cases.                                    │
│   209 │   │   │   state = TrialState.PRUNED                                                      │
│                                                                                                  │
│ C:\Users\me\Docs\IA\heretic\src\heretic\main.py:617 in objective_wrapper                 │
│                                                                                                  │
│   614 │                                                                                          │
│   615 │   def objective_wrapper(trial: Trial) -> tuple[float, float]:                            │
│   616 │   │   try:                                                                               │
│ ❱ 617 │   │   │   return objective(trial)                                                        │
│   618 │   │   except KeyboardInterrupt:                                                          │
│   619 │   │   │   # Stop the study gracefully on Ctrl+C.                                         │
│   620 │   │   │   trial.study.stop()                                                             │
│                                                                                                  │
│ C:\Users\me\Docs\IA\heretic\src\heretic\main.py:589 in objective                         │
│                                                                                                  │
│   586 │   │   │   print("* Reloading model...")                                                  │
│   587 │   │   │   model.reset_model()                                                            │
│   588 │   │   │   print("* Abliterating (Arbitrary-Rank Ablation)...")                           │
│ ❱ 589 │   │   │   model.ara_abliterate(good_module_io, bad_module_io, ara_parameters)            │
│   590 │   │   else:                                                                              │
│   591 │   │   │   print("* Resetting model...")                                                  │
│   592 │   │   │   model.reset_model()                                                            │
│                                                                                                  │
│ C:\Users\me\Docs\IA\heretic\src\heretic\model.py:631 in ara_abliterate                   │
│                                                                                                  │
│   628 │   │   │   │   │                                                                          │
│   629 │   │   │   │   │   # Convergence usually happens within 2-3 steps, so this is more than   │
│   630 │   │   │   │   │   for step in range(5):                                                  │
│ ❱ 631 │   │   │   │   │   │   loss = optimizer.step(closure)                                     │
│   632 │   │   │   │   │   │   # print(                                                           │
│   633 │   │   │   │   │   │   #    f"\\[{layer_index}/{component}/{module_index}] Step: {step}   │
│   634 │   │   │   │   │   │   # )                                                                │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\optimizer.py:526 in    │
│ wrapper                                                                                          │
│                                                                                                  │
│    523 │   │   │   │   │   │   │   )                                                             │
│    524 │   │   │   │                                                                             │
│    525 │   │   │   │   # pyrefly: ignore [invalid-param-spec]                                    │
│ ❱  526 │   │   │   │   out = func(*args, **kwargs)                                               │
│    527 │   │   │   │   self._optimizer_step_code()                                               │
│    528 │   │   │   │                                                                             │
│    529 │   │   │   │   # call optimizer step post hooks                                          │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\_dynamo\eval_frame.py:1181   │
│ in _fn                                                                                           │
│                                                                                                  │
│   1178 │   │   │   │   │   │   │   }                                                             │
│   1179 │   │   │   │   │   │   ):                                                                │
│   1180 │   │   │   │   │   │   │   return fn(*args, **kwargs)                                    │
│ ❱ 1181 │   │   │   │   │   return fn(*args, **kwargs)                                            │
│   1182 │   │   │   │   finally:                                                                  │
│   1183 │   │   │   │   │   set_eval_frame(None)                                                  │
│   1184 │   │   │   finally:                                                                      │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\utils\_contextlib.py:124 in  │
│ decorate_context                                                                                 │
│                                                                                                  │
│   121 │   def decorate_context(*args, **kwargs):                                                 │
│   122 │   │   # pyrefly: ignore [bad-context-manager]                                            │
│   123 │   │   with ctx_factory():                                                                │
│ ❱ 124 │   │   │   return func(*args, **kwargs)                                                   │
│   125 │                                                                                          │
│   126 │   return decorate_context                                                                │
│   127                                                                                            │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py:478 in step   │
│                                                                                                  │
│   475 │   │   │   │   │   def obj_func(x, t, d):                                                 │
│   476 │   │   │   │   │   │   return self._directional_evaluate(closure, x, t, d)                │
│   477 │   │   │   │   │                                                                          │
│ ❱ 478 │   │   │   │   │   loss, flat_grad, t, ls_func_evals = _strong_wolfe(                     │
│   479 │   │   │   │   │   │   obj_func,                                                          │
│   480 │   │   │   │   │   │   x_init,                                                            │
│   481 │   │   │   │   │   │   t,                                                                 │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py:114 in        │
│ _strong_wolfe                                                                                    │
│                                                                                                  │
│   111 │   │   │   break                                                                          │
│   112 │   │                                                                                      │
│   113 │   │   # compute new trial value                                                          │
│ ❱ 114 │   │   t = _cubic_interpolate(                                                            │
│   115 │   │   │   # pyrefly: ignore [index-error]                                                │
│   116 │   │   │   # pyrefly: ignore [unbound-name]                                               │
│   117 │   │   │   bracket[0],                                                                    │
│                                                                                                  │
│ C:\Users\me\miniconda3\envs\heretic\Lib\site-packages\torch\optim\lbfgs.py:27 in         │
│ _cubic_interpolate                                                                               │
│                                                                                                  │
│    24 │   #   d2 = sqrt(d1^2 - g1*g2);                                                           │
│    25 │   #   min_pos = x2 - (x2 - x1)*((g2 + d2 - d1)/(g2 - g1 + 2*d2));                        │
│    26 │   #   t_new = min(max(min_pos,xmin_bound),xmax_bound);                                   │
│ ❱  27 │   d1 = g1 + g2 - 3 * (f1 - f2) / (x1 - x2)                                               │
│    28 │   d2_square = d1**2 - g1 * g2                                                            │
│    29 │   if d2_square >= 0:                                                                     │
│    30 │   │   d2 = d2_square.sqrt()                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ZeroDivisionError: float division by zero
```

p-e-w:
@EugeoSynthesisThirtyTwo 

Thanks for letting me know. The objective has small gradient discontinuities because of the top-k neighbor selection, which might be the root cause of this problem.

p-e-w:
UGI results for the first ARA model are in. They're promising, though not quite as good as I had hoped:

ara_ugi

&nbsp;

The difference between Derestricted, SOMA, and ARA is basically noise considering the range of the rest of the leaderboard, but [MuXodious/gpt-oss-20b-RichardErkhov-heresy](https://huggingface.co/MuXodious/gpt-oss-20b-RichardErkhov-heresy) is clearly much better than any of them.

So ARA appears to work well, especially considering that it operates completely differently from any other method on that list, but it's not quite where I want it to be yet, which is at the number 1 spot. I'm going to try adding magnitude preservation to the optimizer (mimicking MPOA, which has been suggested by several people), and see whether it improves the result.

Interestingly, no uncensoring method so far has even come close to preserving gpt-oss-20b's NatInt score of 27.18.

erm14254:
>I'm going to try adding magnitude preservation to the optimizer (mimicking MPOA, which has been suggested by several >people), and see whether it improves the result.

More ARA models have been rated in the UGI, there is a pattern where those alliterated by MPOA do in general retain more quality than ARA even at higher KL divergence.

p-e-w:
> More ARA models have been rated in the UGI, there is a pattern where those alliterated by MPOA do in general retain more quality than ARA even at higher KL divergence.

Not sure if this is accurate. `ArliAI/gpt-oss-20b-Derestricted` (which uses MPOA) actually has a substantially worse NatInt score than `p-e-w/gpt-oss-20b-heretic-ara-v3`.

I've been researching how we can predict model quality more reliably than using the KLD alone. Please see #236 for initial results.

erm14254:
> Not sure if this is accurate. `ArliAI/gpt-oss-20b-Derestricted` (which uses MPOA) actually has a substantially worse NatInt score than `p-e-w/gpt-oss-20b-heretic-ara-v3`.

I said in general it's not a clear-cut rule and I look at more than just Native Intelligence, plus if you would notice some abliterated models got quite a bit higher Native Intelligence post abliteration than the vanilla baseline pre-abliteration.

> I've been researching how we can predict model quality more reliably than using the KLD alone. Please see https://github.com/p-e-w/heretic/issues/236 for initial results.

Thank, I'll look into it.

p-e-w:
> if you would notice some abliterated models got quite a bit higher Native Intelligence post abliteration than the vanilla baseline pre-abliteration.

This has never been achieved with gpt-oss-20b (very far from it, actually), which is why it is my model of choice for experiments. Whatever works there will likely work elsewhere as well.

p-e-w:
The ARA branch now supports row-norm preservation during optimization using a reparameterization constraint, as suggested by @spikymoth and others. It will be interesting to see whether this improves benchmark scores.

p-e-w:
I just published [p-e-w/gpt-oss-20b-heretic-ara-v4](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara-v4), which represents the cutting edge of ARA development. It combines ARA with row-norm preservation inspired by MPOA, and optimizes for PIQA rather than KLD. It is the first model I've seen that beats the base model's (openai/gpt-oss-20b) PIQA score.

I have submitted the model for UGI evaluation (https://huggingface.co/spaces/DontPlanToEnd/UGI-Leaderboard/discussions/634). Unfortunately, the maintainer of UGI has announced that they will be evaluating non-base models only on popular request in the future. So please vote for this eval request so we can get that valuable data point!

p-e-w:
**Important update:** The model I linked above ([p-e-w/gpt-oss-20b-heretic-ara-v4](https://huggingface.co/p-e-w/gpt-oss-20b-heretic-ara-v4)) appears to have been corrupted on upload. Trying to load it from HF throws a shape mismatch error. I don't know what the problem is yet.

teleprint-me:
I sketched out a model diff script -  but its not functional at the moment - because I was trying to figure out what the difference was between v3 and v4 that was causing the issue.

I wasnt sure if it was due to this branch, the upload, or if it was intentional. My initial instinct was that the model was currupted, but good to know it wasn't intentional.

p-e-w:
I mean, obviously corrupting the model wasn't intentional, not sure what you mean there. I think there are bugs in either Transformers 5, or PEFT, or the interaction between the two. IMO, Transformers 5 left beta way too early, there are still plenty of problems and Heretic has been affected by several of them.

teleprint-me:
By intentional, I meant maybe the model wasn't corrupted and you had a reason for reshaping the weights and expected that output - Nothing else. And yeah, HF has been making a habit of that as of late. Regardless, I think ARA is really interesting. Once I have more time, I'm definitely digging into this.

YuanBoXie:
I'm curious, have you tested whether this attack method works on DeepRefusal-hardened models?
[1] Beyond Surface Alignment: Rebuilding LLMs Safety Mechanism via Probabilistically Ablating Refusal Direction, EMNLP 2025
[2] https://github.com/YuanBoXie/DeepRefusal

p-e-w:
@YuanBoXie 

Can you link to a model that has been trained with that approach?

YuanBoXie:
> @YuanBoXie
> 
> Can you link to a model that has been trained with that approach?

Maybe you can try： https://huggingface.co/skysys00/Meta-Llama-3-8B-Instruct-DeepRefusal
Is it possible to provide a real response for the test collection? I tested the algorithm in the main branch and couldn't break DeepRefusal. Also, some of the seemingly low number of refusal values ​​were actually cases that failed as mentioned in [issue #290](https://github.com/p-e-w/heretic/issues/290).

flashburns:
Sorry if this is not the place to do so and I am adding to the noise, but I just wanted to say thank you p-e-w! I have been using a gemma4 ARA Ablation and its the most fun I have had in a long time with a model, genuinely been like over a year from the last time I have been this excited about anything in the LLM space. Your work is wonderful and your users appreciate it (at least I do), thought id voice it!

p-e-w:
@flashburns Thank you, that means a lot to me.

antis0007:
Hello! I'm incredibly interested in the possibility of integrating ARA into a Gemma4 E2B-it/E4B-it model converter script that I've been working on. I've been trying to create export scripts for Torch to Litert-lm model conversion, with heretic used as a intermediate step. ARA is going to be a revolution in model decensoring, your uploaded ARA models are very high quality. My goal is to make the process of exporting uncensored multimodal models for edge devices like smartphones much simpler. Lmk if there's any way to get in touch to discuss the idea with you sometime!

p-e-w:
@antis0007 Feel free to join the Discord (link in README). I'm usually around there for back-and-forth discussions.

p-e-w:
There is currently a serious bug in ARA where the resulting matrices sometimes contain `NaN` or `inf`/`-inf`. This has been reported by at least two users, and I believe it is also the reason why `p-e-w/gpt-oss-20b-heretic-ara-v4` was corrupted on export (see above). The exact cause is currently unclear, and it's also unknown whether this bug happens with ARA-LoRA (#332) as well.

Needless to say, this is a blocker for merging ARA into master.

erm14254:
> There is currently a serious bug in ARA where the resulting matrices sometimes contain `NaN` or `inf`/`-inf`. 

Yes it happens but it's very very rare, it mostly happens a lot more when a model doesn't really like ARA which is even rarer as almost all models have no issues with ARA.

p-e-w:
:smile: Yeah but we need to figure out what it means that a model "doesn't like ARA".

I suspect it's an instability in L-BFGS caused by the aforementioned directional discontinuity of the gradient.

kabachuha:
Maybe just prune the trials? Nan check is good, though

p-e-w:
We could prune, but I'd prefer to figure out what is going on. We can't make ARA the default while it has such instabilities that we don't understand.

erm14254:
> We could prune, but I'd prefer to figure out what is going on. We can't make ARA the default while it has such instabilities that we don't understand.

Overall ARA is the best "jack of all trades", MPOA can achieve better results than ARA sometimes but only on a few select models.

Basically,

ARA: Very likely to give you good to great usable results on most models, very few models who "do not like ARA" resulting in bad or less than ideal results (model's accuracy loss is too high despite low KL divergence) and, very very few models and or certain conditions do not work with ARA.

MPOA: Unknown until you run hundred of trials, it may give good usable results, sometimes even better than ARA on certain select models, or it might top at 68/100 refusals after 600 trials, hence wasting time and making you go use ARA instead which will mostly likely give you actual usable results.

kabachuha:
@erm14254 There is also SOM / SOM-POA to consider. It can give great results as well, though from experience and Huggingface ARA is a more robust method.

I think MPOA < SOMPOA < ARA, as of now.

p-e-w:
> models who "do not like ARA" 

We don't know that the models are the problem.

It could be:

* Individual models/tensors
* Specific ARA hyperparameter combinations
* Random behavior from L-BFGS (which isn't deterministic)
* Convergence issues due to gradients or some other problem

"It works most of the time" isn't good enough. This PR is not going in until this problem is fully understood. Hopefully I'll have time to work on ARA again soon.

rootacc3ss:
I haven't kept up to date on this, but am still really curious. Let me know if it's worth trying to test on one of the newer GLM models, or Kimi K2.7 code... Would be willing to spend the money for the community

p-e-w:
@rootacc3ss 

Please try #332 instead, it's the version we are actually going to merge eventually.

kabachuha:
What's the next part of the roadmap on ARA?

p-e-w:
Reproduce the broken tensors issue and fix it, or confirm that it doesn't happen with ARA-LoRA.

red40maxxer:
I think there's a bug with LoRA here: 

When `use_ara_lora=True` and `row_normalization=Full`, `ara_lora_abliterate` evals a row-normalized effective weight for the objective, but the deployed model computes the un-normalized one. I'm not sure exactly what the fix is since the idea of row-norm preservation for a LoRA adapter doesn't make sense, but maybe we shouldn't allow this to happen in the first place?

p-e-w:
> I'm not sure exactly what the fix is since the idea of row-norm preservation for a LoRA adapter doesn't make sense

We already combine LoRAs with magnitude preservation in regular abliteration, but this situation is different because we want to preserve norms while optimizing, because post-optimization projection will break many of the properties we are trying to establish.

One approach could be to merge the LoRA into the full matrix at every optimization step, then reparametrize, then subtract the full matrix to get the (now full-rank) LoRA back, then cull dimensions using singular values. But that's very ugly, there might be a better solution.

hpnyaggerman:
@p-e-w, I took apart gpt-oss-20b-heretic-ara-v4, the model you pointed to as corrupted, and the cause appears to be none of the four possibilities you've listed in your message earlier in this thread; it appears instead to be a serialization bug in transformers, and it reproduces with no ARA, no Heretic and no optimizer anywhere in the process. The reproducer is a load and a save of the unmodified base model on transformers 5.4.0.

```python
m = AutoModelForCausalLM.from_pretrained("openai/gpt-oss-20b", dtype="auto", device_map="cuda:0")
assert type(m.model.layers[0].mlp.experts).__name__ == "Mxfp4GptOssExperts"
m.save_pretrained("out")
```

That writes 459 keys, 96 of them doubled (`*_blocks_blocks`, `*_blocks_scales`), and the resulting key set is identical to the published artifact's, symmetric difference zero, while the same script on 5.14.1 gives 0 doubled and 96 canonical. The root cause is `pattern = pattern.removesuffix("$")` at `core_model_loading.py:474`, where stripping the anchor means that `down_proj` also matches inside `down_proj_blocks`, so that the `revert_weight_conversion` call at `modeling_utils.py:3354` renames keys that `get_state_dict_and_metadata` had already written correctly earlier in the same function. It was fixed in 5.5.2, and the expert biases came through intact only because `Mxfp4ReverseDeserialize.convert` carries an explicit bias-only workaround for the same problem.

On the four possibilites you have listed as possible causes, none of them seem to have much merit as the cause of corruption in the model you've pointed to. The suggestion that individual models or tensors are at fault does not explain why the damage covers all 96 expert tensors across all 24 layers uniformly, including layers 0 and 1, which ARA never writes, or why the whole of it reproduces on the untouched base model. ARA hyperparameters cannot be involved in a failure that reproduces with ARA not running at all, and there is no optimizer in the reproducer, so L-BFGS nondeterminism is also not it, while all 96 expert payloads are, separately, byte-identical to upstream by per-tensor SHA-256, so nothing ever wrote them. Gradients and convergence fall under the same exclusion, and the ARA deltas themselves look healthy - relative Frobenius norm between 0.0060 and 0.0303, per-row norms within 0.8%, no dead rows, no blow-up.

The write set of ARA is in any case disjoint from the damage, since it wrote 22 `self_attn.o_proj.weight` in layers 2 through 23, matching the declared range, while the broken set is 96 expert tensors in layers 0 through 23, and layers 0 and 1, being byte-identical to upstream, were demonstrably never touched by ARA, and yet all eight of their expert tensors are mis-keyed. As for timing, the model was uploaded on 2026-04-02 and its own config records transformers 5.4.0, which identifies the version that wrote it rather than merely the version it was built from: `transformers_version` is assigned unconditionally from the running library at `configuration_utils.py:983`, and the base model's config records 4.55.0.dev0, so the field is overwritten on save rather than inherited. The defect window is 5.4.0 through 5.5.1, with 5.5.2 carrying the fix on 2026-04-09. 5.3.0, which is what `uv.lock` on this branch actually pins, saves correctly, so the lockfile environment would not have produced this; the run that did used a transformers newer than the lock. I established that window by running the load-and-save above on each version rather than by reading the code, which matters because a static reading of the rename functions gets 5.3.0 wrong: the doubling only shows up end to end.

The reason that none of this surfaced is that Heretic computes its benchmark numbers from the in-memory model before `save_pretrained`, and the in-memory model was correct, while the save does not raise, and reloading the corrupt file does not raise either, on any version I tried up to 5.14.1. That does give a way of triaging the older reports, in that a model which benchmarked well during the run and was broken when somebody downloaded it is this bug, whereas a model that failed to converge during the run is not, and that class of report stays open.

On the ARA-LoRA question you raised, the artifact predates it (3b70fe5 vs 25979ad), so it says nothing directly, though the bug sits in serialization, downstream of and independent of how the weights are produced, which means that ARA-LoRA on anything from 5.4.0 through 5.5.1 would be corrupted identically (or so I think, at least), while 5.3.0 and anything from 5.5.2 onward are unaffected. Where it exports as a LoRA adapter (`strategy == "adapter"`) the expert tensors are never written at all, so only the full-model save is exposed, and with `use_ara` set that branch saves `model.model` directly, which makes it the one that matters here.

The file is repairable in place and needs no re-abliteration, the operation being a rename of the 96 header keys and a pad of the header back to its original length, after which the payload comes out byte-identical and `cmp -i 53448` exits 0. Repaired, the model scores PIQA 0.7731 with 6/100 refusals against a base control of 98/100 measured in the same session, while unrepaired it sits at 0.5267 +- 0.0116, which is essentially chance, that last figure being measured through the dequantize path because below compute capability 10.0 the mis-keyed file cannot complete a single forward pass on the MXFP4 kernels at all, the experts having been left as plain uninitialised tensors rather than the strided ones the kernel expects. To be precise, 0.7731 is not a digit-for-digit reproduction of the card's 0.7742 and 9/100, since identical weights do not force identical logits across a different transformers, triton and GPU, and `--batch-size` does not reach PIQA in any case, `Evaluator.get_score` constructing `HFLM(..., batch_size="auto")`, which selected 32 here.

Pinning a known-good transformers does not, however, settle the matter on its own. There is a second bug on the MXFP4 dequantize save path, and it was introduced in 5.5.2, which is the same release that fixed the first one. Across versions, run end to end: 5.3.0 and 5.4.0 raise `NotImplementedError` from `core_model_loading.py:101` and write nothing at all, 5.5.0 and 5.5.1 save correctly with all 48 per-layer expert keys intact, and from 5.5.2 onward every layer's experts collapse onto the two literal keys `down_proj$` and `gate_up_proj$`, the reverse mapping having used the pattern string itself as the output key rather than substituting the match into the per-layer name, so that all 24 layers write to the same two keys and only the last of them survives. Every one of the twenty-one releases from 5.5.2 to 5.14.1 collapses, checked on a two-layer cut of the model which reproduces the full-scale outcome exactly at 5.4.0, 5.5.1, 5.5.2 and 5.14.1. That is a different order of damage from the first bug. There every payload was present and only the names were wrong, which is why renaming recovers it; here 23 layers of expert weights are overwritten before anything reaches disk, leaving 5.2 GB instead of roughly 42 GB, with 1.6 GB of expert tensors where 24 layers belong, in a file that cannot be repaired at all and can only be regenerated.

Heretic never asks for that path, since `quantization` defaults to `none` and `model.py` passes no `quantization_config` at all for gpt-oss, but transformers decides for itself from the checkpoint's own `quant_method: "mxfp4"` and sets `dequantize` whenever compute capability is below 7.5 or triton or `kernels` is missing. This branch's lock carries both triton and `kernels`, so the live trigger is capability, which means a V100 or anything older is likely to be affected. On the currently locked 5.3.0 such a user gets a loud crash at save time, whereas bumping to `~=5.6` to escape the first bug moves that same user from a loud crash to the silent loss of 23 of 24 layers. So the fix for the first bug introduces some users to the second.

Given that, I would suggest a check after saving that the set of tensor names written matches the set in the source checkpoint, an invariant which always holds for Heretic, since it only rewrites values and never renames anything, and which both bugs violate. It cannot prevent either of them, but it turns a silent loss of 23 layers into a loud failure before anything reaches the Hub, and for the second bug it is the only protection there is, given that the fix for the first one is what introduced it. It reads safetensors headers only, so it costs nothing. I tried it against six checkpoints, and it passes the repaired artifact and a correct 5.14.1 save, and flags the published artifact, a 5.4.0 save, an explicit dequantize-path save and one that transformers dequantized on its own without being asked, and on this branch the natural place for it is straight after the `Model saved to` line in `main.py`, around line 898, on the non-adapter path, where there is currently no post-save verification of any kind.

On the reach of the first bug, it is narrow, more narrow than one would perhaps expect. First, stripping an anchor cannot affect a pattern that does not have one, so the only converters at risk are those whose target pattern ends in `$`, and in the whole of transformers 5.4.0 exactly two do, `down_proj$` and `gate_up_proj$` (both in `quantizer_mxfp4.py`). Having run the function over every `target_patterns` literal in the library, I get precisely those two modified and everything else returned unchanged, the MoE renamings in `conversion_mapping.py` among them, which carry no trailing `$` for it to remove. Second, those two converters exist only while the MXFP4 quantizer is active, and the quantizer is reached through exactly one dispatch, `AutoHfQuantizer.from_config`, keyed on `quant_method` as read from the checkpoint's own config; no architecture opts into it, and `models/gpt_oss` does not so much as name mxfp4. Both conditions being required together, the damage cannot extend past checkpoints that declare `quant_method: "mxfp4"`, and in practice that means gpt-oss. A dense bf16 model never brings those two converters into existence at all, so nothing here can reach ARA models on other architectures, which explains why the failure mode was never seen on, for example, Gemma 4 (most well known example of an ARA-abliterated model, to me at least). What determines whether or not the failure will happen is the quantization of the checkpoint, not anything ARA does and not any property of the model itself.

On your other question, whether this is the only mode of corruption, I cannot tell you that it is, having had the one artifact to work from, and what is settled here is the broken tensors in that one and nothing wider. But I would treat any report resting on a saved MXFP4 MoE checkpoint as unusable evidence about ARA until its keys have been checked, because the in-process numbers cannot detect this failure.

For anyone reproducing this, the environment was four NVIDIA A16-16Q vGPUs at compute capability 8.6 with 16 GB each, on a 24-vCPU Cascadelake host with 251 GB of RAM running Debian 13 on kernel 6.12.95. The guest driver is 550.54.15, which exposes CUDA 12.4, so torch's cu130 build runs on top of the cuda-compat-13-1 forward-compatibility package (590.48.01) through `LD_LIBRARY_PATH`; the driver itself needed three small patches to build against kernel 6.12, none of them touching anything numerical. Python was 3.12.13 throughout, with torch 2.13.0+cu130, triton 3.7.1, accelerate 1.14.0 and numpy 2.5.1, and with `kernels` pinned per transformers version, since those bounds are declared only under the `[kernels]` extra and an out-of-range `kernels` breaks the import: 0.10.5 with 5.3.0, 0.12.3 with 5.4.0 and the 5.5.x line, 0.15.2 with 5.14.1. The twenty-one-release sweep for the second bug ran without `kernels` installed at all, the dequantize path having no use for it. It is worth explicitly differentiating the two kinds of result here, in that the key-naming ones are pure string handling on the save path and are hardware-independent, so they should reproduce anywhere, whereas the PIQA and refusal figures, and the compute-capability assertion, are hardware-dependent and could be different on different silicon.

hpnyaggerman:
With this issue out of the way (if no holes are poked in my analysis above), I hope work on ARA will be resumed and that it will soon be merged into master. ARA + custom objectives (which are already merged) would make heretic truly unstoppable, I believe.

p-e-w:
That's incredible detective work, @hpnyaggerman! Thank you so much for figuring this out.

Unfortunately, this is in line with what I have already noticed several times: That Transformers 5 is of poor quality overall, with early versions showing serious bugs even with popular models. I did consider the possibility of a Transformers bug in this case also, but then I remembered that I had successfully exported GPT-OSS with Transformers 5 before, so I dismissed it. What I failed to realize is that the bug could have been introduced post-5.0, which you have now confirmed.

Given your deep insight into the problem, could you open a PR implementing the check you suggested?

I'm currently wrapping up Heretic's "sister project" (which, incidentally, might be useful for getting ARA over the finish line), and then it's back to this PR.

hpnyaggerman:
@p-e-w Very well, I will get to that. Should be straightforward enough.

p-e-w:
Superseded by #446.

PR #212, opened 2026-03-05
perf: Update config.default.toml (Adding more refusal markers)

Exponent-Pi:
It increases refusal markers and detects pivots to other topics and all sorts of refusals. This may help refusal detection in the long run as LLMs are instead using pivots as safety measures instead of a straight brittle refusal with a bonus of less false positives.


p-e-w:
How was this list derived?

About half of the new entries would already have been detected previously (through markers like "i can't"). It's not clear to me why those original markers would lead to false positives.

Can you give some examples of actual responses from models that were misclassified before and are being correctly classified with this change?

Exponent-Pi:
I derived it from eliciting common model refusals from all types of LLMs, like Claude, ChatGPT, Gemini, and open-weight LLMs. And for example, "I apologize" was something that was not detected in the original. I took the time to actually analyze the mannerisms that appear frequently in other responses that the original would not have caught just in case my additions do not catch the beginning of the refusal. Also, don't forget distillation is a real thing, and RLHF comes along with that, and it is increasingly teaching models to acknowledge your request and then pivot to a "safer" topic. like "I understand," then we can identify the common "however" responses in a lot in these pivoting or indirect refusals.

p-e-w:
> And for example, "I apologize" was something that was not detected in the original.

I have seen LLMs use the words "I apologize" quite frequently in non-refusal situations though.

As I said, it would be helpful if you could give some concrete examples of actual responses from models that were misclassified before.

uponminiature:
`I won't` and `I can't` are some of GLM 5's go-tos, I experienced those responses a lot while evaluating whether harmful prompts would be refused.

p-e-w:
"I won't" and "I can't" were already in the list before this PR though.

Still waiting for a concrete example of an actual model response whose classification is getting fixed by this PR.

uponminiature:
That's right, mistake on my part for overseeing `won'`.

p-e-w:
We're using `won'` rather than `won't` on purpose, because Gemma 3 has a particular type of abliteration artifact where `won't` is corrupted to `won' t`.

p-e-w:

PR #215, opened 2026-03-08
fix: display all abliterable components across layers

erm14254:
The current code only displays abliterable components from layer 0, which is misleading for hybrid architectures like Qwen3.5 that use different attention types across layers (e.g., `linear_attn.out_proj` in some layers, `self_attn.o_proj` in others).

This fix iterates through all layers to collect and display the complete set of abliterable components with accurate module counts.

Before (Qwen3.5-27B):
* attn.out_proj: 1 modules per layer
* mlp.down_proj: 1 modules per layer

After (Qwen3.5-27B):
* attn.out_proj: 48 modules total
* attn.o_proj: 16 modules total
* mlp.down_proj: 64 modules total

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -158,9 +158,7 @@ class Model:
                     all_components[component] = 0
                 all_components[component] += len(modules)
         for component, count in all_components.items():
-            print(
-                f"  * [bold]{component}[/]: [bold]{count}[/] modules total"
-            )
+            print(f"  * [bold]{component}[/]: [bold]{count}[/] modules total")
 
     def _apply_lora(self):
         # Guard against calling this method at the wrong time.


p-e-w:
Are you using an unmerged PR for Qwen3.5? Because our current implementation (#187) stores both linear and standard attention projections under the `attn.o_proj` component name, despite them having different names in the Transformers implementation.

erm14254:
Yes, despite no mention of `attn.o_proj` in the logs, the end result still has it, this PR is just to add "attn.o_proj" in the list inside the logs at the beginning so that you know it's there at launch, but even without this PR the final result still does alliterate all three,  `attn.out_pro`, `attn.o_proj` and `mlp.down_proj`.

p-e-w:
That's not what I meant.

You wrote in the PR description:

> After (Qwen3.5-27B):
> * attn.out_proj: 48 modules total
> * attn.o_proj: 16 modules total
> * mlp.down_proj: 64 modules total

Is that the output you're actually seeing? Because if it is, there is a bug somewhere. The line

> attn.out_proj: 48 modules total

shouldn't be there, because we don't use that component name.

erm14254:
PR #215 is just to add `attn.o_proj` into the log to show that it's there, because originally even if it's there, it's not listed in the log.

gemini-code-assist says:

> This pull request correctly fixes an issue where not all abliterable components were being displayed for models with hybrid architectures. The change iterates through all layers to collect and count all components, which is the right approach.

It's just a very minimal change, it doesn't change much of anything, the program would run the same regardless of this PR, it's just displays more abliterable components in the logs that's it, since it's not a very important change and probably not worth the code change I will just close this PR.

p-e-w:
Why did you close this PR? This is a good change, there is just an important question that needs to be resolved.

Let me try one more time to explain the issue:

Your PR description includes the output

> attn.out_proj: 48 modules total

However, `attn.out_proj` is never used as a component name in `get_layer_modules`, unlike `attn.o_proj`. So where does it come from in the output? That's all I want to know.

erm14254:
> Why did you close this PR? This is a good change, there is just an important question that needs to be resolved.
> 
> Let me try one more time to explain the issue:
> 
> Your PR description includes the output
> 
> > attn.out_proj: 48 modules total
> 
> However, `attn.out_proj` is never used as a component name in `get_layer_modules`, unlike `attn.o_proj`. So where does it come from in the output? That's all I want to know.

Sorry for closing prematurely, I will reopen it.

The `attn.out_proj` comes from the model architecture itself. Qwen3.5 uses different layer types and have different component naming. When I print the abliterable components, it shows what Heretic discovers in the actual model layers, not just what's in get_layer_modules.

The PR addresses displaying these discovered components correctly across layers, rather than assuming uniform naming throughout.

Here is what happens, I add this inside config.toml:

target_components = ["attn.o_proj", "attn.out_proj", "mlp.down_proj"]

And then what I am actually seeing is this:

        Abliterable components:
        attn.out_proj: 1 modules per layer
        mlp.down_proj: 1 modules per layer

PR #215 should change this ↑, into this:

        Abliterable components:
        attn.out_proj
        attn.o_proj
        mlp.down_proj

It's just to add attn.o_proj into the log to show that it's there, because originally even if it's there, it's not listed in the log.

p-e-w:
Please fix CI.

erm14254:
> Please fix CI.

Sorry, I fixed formatting with ruff. CI workflow is awaiting approval to run. Thanks.

p-e-w:
Merged, thanks!

PR #224, opened 2026-03-12
Use BERT-based model for refusal classification

Galunid:
I believe text based heuristic implemented by heretic is insufficient and introducing lots of noise during optimization. This should hopefully reduce number of refusals and KL-divergence by providing heretic's optimizers with higher quality data at the cost of slower iteration and higher memory usage.

PR is work in progress. The code hasn't been tested yet. Most likely, I'll have to implement batching. I picked `NousResearch/Minos-v1` as default model because it was specifically trained to classify refusals and NousResearch is a credible organization with good releases such as Hermes series in the past.

Let me know if you have any concerns/suggestions.

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -230,6 +230,16 @@ class Settings(BaseSettings):
         description="Directory to save and load study progress to/from.",
     )
 
+    refusal_classifier_model: str = Field(
+        default="NousResearch/Minos-v1",
+        description="Hugging Face model ID for refusal classification.",
+    )
+
+    refusal_classifier_threshold: float = Field(
+        default=0.5,
+        description="Confidence threshold for refusal classification (0.0 to 1.0).",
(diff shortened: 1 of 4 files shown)


p-e-w:
Thanks for the PR, but I tested Minos many months ago, and found it to be just as unreliable as keyword-based classification, while being orders of magnitude slower. The model appears to have effectively learned to recognize keywords, and is easily fooled by individual keywords (such as "sorry" or "I can't") incidentally appearing in the response, *even if the response itself is not a refusal.*

Any alternative classification method should also build on top of the scorer plugin system (#53).

Galunid:
I see, it's disappointing to see Minos turned out to be unreliable, thanks for letting me know.

uis246:
I think there still is a promise in using AI model to classify refusals of other AI models.

PR #226, opened 2026-03-14
feat: add MLX backend for native Apple Silicon abliteration

overtimepog:
## Summary

- Adds a complete **MLX backend** enabling heretic to run abliteration natively on Apple Silicon using the [MLX framework](https://github.com/ml-explore/mlx) and [mlx-lm](https://github.com/ml-explore/mlx-lm)
- Supports direct use of MLX-format quantized models (e.g. 4-bit safetensors from mlx-community) without conversion, leveraging Metal GPU acceleration and unified memory
- Auto-detects MLX models when `backend = "auto"` (default), or can be explicitly set with `--backend mlx`

### New files
- **`src/heretic/mlx_model.py`** (~660 lines) — Full MLX model backend implementing the same interface as the PyTorch `Model` class:
  - Model loading via `mlx_lm.load()`
  - Text generation with greedy decoding
  - Hidden state extraction via custom layer-by-layer forward pass
  - Log-probability extraction for KL divergence evaluation
  - Abliteration via dequantize→modify→requantize (handles both regular `Linear` and MoE `SwitchLinear`/`QuantizedSwitchLinear`)
  - Weight reset by restoring saved originals
  - Model saving (safetensors + config + tokenizer)
  - Streaming chat for interactive testing
- **`tests/test_mlx_model.py`** — 31 tests (6 unit + 25 integration) covering all operations

### Modified files
- **`config.py`** — Added `Backend` enum (`AUTO`/`PYTORCH`/`MLX`) with default `AUTO`
- **`main.py`** — Added `resolve_backend()`, `create_model()` factory, MLX device detection, and MLX-aware save/upload flows
- **`utils.py`** — Added MLX memory reporting (`get_peak_memory`, `get_active_memory`)
- **`pyproject.toml`** — Added `[mlx]` optional dependency group, pytest marker config
- **`config.default.toml`** — Documented the new `backend` setting

### Design decisions
- **Torch at the boundary**: MLX arrays internally, converted to `torch.Tensor` at the interface boundary so `evaluator.py` and `main.py` work unchanged
- **No PEFT dependency for MLX**: Direct weight modification with saved originals for reset (instead of LoRA adapters)
- **Dequantize→modify→requantize** for abliteration on quantized weights, processing one module at a time to limit memory spikes
- **Zero changes to existing PyTorch code path**: The `Backend.PYTORCH` path is completely unchanged

### MoE support
Tested with **Qwen3-Coder-30B-A3B-Instruct-MLX-4bit** (128 experts, 8 active per token, 48 layers). Handles:
- `SwitchGLU` / `QuantizedSwitchLinear` stacked expert weights
- Per-expert abliteration via `mx.einsum` for efficient vectorized computation
- Both global and per-layer refusal directions

## Test plan

- [x] Unit tests pass (6/6) — Backend enum, Settings field, MLXModel interface, model detection
- [x] Integration tests pass (25/25) — Loading, layer introspection, generation, residuals, logprobs, abliteration, reset, chat, save
- [x] Full suite run 3 consecutive times with 31/31 passing each time
- [x] Verify on a non-Apple system that the PyTorch path is unaffected (MLX imports are lazy)
- [x] Test with additional MLX model architectures (Llama, Mistral, Phi, etc.)

## Usage

```bash
# Install with MLX support
pip install 'heretic-llm[mlx]'

# Auto-detect MLX model format
heretic /path/to/mlx-model

# Or explicitly specify backend
heretic --backend mlx /path/to/mlx-model
```

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -1,6 +1,12 @@
 # Rename this file to config.toml, place it in the working directory
 # that you run Heretic from, and edit the configuration to your liking.
 
+# Backend to use for model loading and inference. Options:
+# "auto" (detect from model format — uses MLX for MLX models on Apple Silicon),
+# "pytorch" (HuggingFace Transformers + PyTorch),
+# "mlx" (Apple MLX, requires macOS with Apple Silicon and `pip install heretic-llm[mlx]`).
+# backend = "auto"
+
 # List of PyTorch dtypes to try when loading model tensors.
 # If loading with a dtype fails, the next dtype in the list will be tried.
 dtypes = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+mlx = [
+    "mlx>=0.20",
+    "mlx-lm>=0.20",
+]
 research = [
     "geom-median~=0.1",
     "imageio~=2.37",
@@ -49,6 +53,7 @@ research = [
 
 [dependency-groups]
 dev = [
+    "pytest>=8.0",
     "ruff>=0.14.5",
     "ty>=0.0.5",
 ]
@@ -69,3 +74,8 @@ build-backend = "uv_build"
 
 [tool.uv.build-backend]
 module-name = "heretic"
+
+[tool.pytest.ini_options]
+markers = [
+    "integration: tests that require a real MLX model on disk and Apple Silicon hardware",
+]
--- a/run_mlx.sh
+++ b/run_mlx.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Run heretic with the MLX backend on an MLX model.
+# Usage: ./run_mlx.sh /path/to/mlx-model [extra heretic args...]
+
+set -euo pipefail
+
+MODEL="${1:?Usage: $0 /path/to/mlx-model [--n-trials 10 ...]}"
+shift
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PYTHON="${HERETIC_PYTHON:-/opt/homebrew/bin/python3.11}"
+
+export PYTHONPATH="${SCRIPT_DIR}/src${PYTHONPATH:+:$PYTHONPATH}"
+export PYTHONUNBUFFERED=1
+
+exec "$PYTHON" -c "
+import sys
+sys.argv = ['heretic', '--backend', 'mlx', '--model', '$MODEL'] + sys.argv[1:]
+from heretic.main import main
+main()
+" "$@"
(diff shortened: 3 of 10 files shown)


p-e-w:
Thanks for the PR, but I think this is targeting the wrong level of abstraction.

Heretic already uses a library that abstracts away the backend for language model operations. That library is Transformers. Transformers currently doesn't have an MLX backend, and **that** is where this should be fixed.

There are thousands of projects that depend on Transformers, and all of them lack MLX support as a result. The solution is to add MLX support to Transformers, not to riddle each of those thousands of projects with shims.

I also don't have any Apple hardware myself, so this code is completely unmaintainable for me. This is another reason why this needs to be implemented upstream. Library consumers typically don't have the bandwidth to re-implement and re-test every logic change for every backend.

overtimepog:
yeah I admit this was probably the wrong level for a full release or actual addition to heretic, honestly please feel free to close this PR and completely ignore it lol. I really did just slap this in as I wanted to try abliterating a model on my macbook, didn't wanna scour the forks for an MLX version, and had claude build, and test it, and ig it made a fork and PR for it while it had time, sorry if this messed up the heretic repo in anyway -Overtime

p-e-w:
Don't worry, you didn't mess up anything. This fork is valuable to have for people who run Apple hardware, it just can't go into mainline at this point for the reasons mentioned.

Incidentally, what kind of speedup are you seeing compared to MPS?

JiwaniZakir (commented):
In `run_mlx.sh`, the `$MODEL` variable is interpolated directly into a Python string literal inside the `-c` argument — any model path containing a single quote or backslash will silently corrupt the Python source or cause a syntax error. A safer approach is to pass the model path as a proper command-line argument and read it via `sys.argv` inside the script, or use `shlex.quote` if you need to keep the current structure.

In `resolve_backend` (`main.py`), `is_mlx_available` is imported twice — once in the `Backend.MLX` branch and again in the `AUTO` block. These imports can be consolidated at the top of the function to avoid the duplication.

The `use_mlx` boolean introduced in `run()` is redundant given `model_backend` is already in scope; all `if use_mlx` checks could just be `if model_backend == Backend.MLX`, which keeps the condition consistent with how the rest of the branching is expressed and removes a second variable to keep in sync.

Finally, the `AUTO` detection in `resolve_backend` silently falls back to PyTorch if `is_mlx_available()` returns `False` even when the user passes an MLX-format model path. That situation is likely a misconfiguration the user would want to know about — a warning at minimum would help rather than silently running what will probably be a broken PyTorch load.

p-e-w:
Closing this PR to keep the list of open PRs focused on the things we are actually working on.

PR #231, opened 2026-03-16
fix: clamp KL divergence to non-negative values

### LINKED ISSUE #141: Don't display negative KL divergences

JasonOA888:
Fixes #141

### DIFF
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -101,7 +101,7 @@ class Evaluator:
             reduction="batchmean",
             log_target=True,
         ).item()
-        # Clamp to non-negative (KL divergence cannot be negative by definition.)
+        # Clamp to non-negative values (KL divergence cannot be negative by definition).
         kl_divergence = max(0.0, kl_divergence)
         print(f"  * KL divergence: [bold]{kl_divergence:.4f}[/]")
 


p-e-w:
This should no longer be an issue with #194. Are you still seeing negative KLDs for some models?

JasonOA888:
Thanks for the pointer! I checked PR #194 - it fixes KL divergence measurement issues caused by prefix detection, but Issue #141 is specifically about floating-point `-0.0` display.

Even with #194 merged, the computation can still produce negative zero due to floating-point arithmetic (not a logical bug). The fix here simply clamps the value to non-negative before formatting.

I don't have a specific model to reproduce this right now - I created the fix based on the issue description. If this isn't occurring in practice anymore, feel free to close both the issue and this PR.

p-e-w:
Yeah, I think we should actually keep the current behavior, since this should never occur in practice, and if it does (as it has in the past), it's indicative of a bug, and it's best if the value sticks out like a sore thumb in that case.

Note also that the KLD logic is currently undergoing a major revision in #53.

Thanks anyway for the PR!

Issue #152, opened 2026-02-14
Uploading LoRA adapter doesn't work


p-e-w:
See https://huggingface.co/p-e-w/gpt-oss-20b-heretic-v3-LoRA/tree/main for an example. The `adapter_model.safetensors` file that is supposed to contain the LoRA is empty. Strangely enough, exporting the adapter to a local folder works, producing a 700 kB safetensors file.

p-e-w:
Actually, it fails for all models. This sucks.

kabachuha:
@p-e-w What do you think of saving the lora adapter to a temp dir and then upload it through hf hub upload folder?

https://huggingface.co/docs/huggingface_hub/v0.13.2/guides/upload#upload-a-folder

p-e-w:
Too brittle, and not what the user expects when they choose "upload".

This has to work in PEFT/Transformers. We're not fixing upstream bugs downstream.

kabachuha:
Idk add a disclaimer "local only" when saving a lora, just like the RAM disclaimer pops up

The lora file weights a few megabytes, it won't clutter the computer

You *really* are overcomplicating things

kabachuha:
This is helpful for like 99%+ users who don't have 256+ RAM

p-e-w:
I continue to believe that the correct way forward is to fix the problem, not work around it. That is, `model.push_to_hub` should do what it claims to do.

I imagine that the root cause of this bug is rather simple, and probably easy to resolve. It's just that this feature is completely useless to me personally, and I have more important things (for me) to work on at the moment.

p-e-w:
The recommended workaround for people who need this feature is to revert 025ab3a8815d8f857a4d8bfce0e7e6af73cb0736. I will not merge a fix or workaround into master without understanding the cause of the problem.

anrp:
I wonder what's broken? Maybe try again, was it a transient issue?
I tried tag v1.2.0 + revert 025ab3a, command line `uv run heretic --model Qwen/Qwen3-0.6B --batch-size 64 --n-trials 10 --n-startup-trials 5` and uploaded lora adapter to https://huggingface.co/anrp/testhereticuploadlora and it worked fine, safetensors file is 1MB-ish?

Edit: to be clear, uploaded the lora from the menu options available, not saved & then uploaded.

p-e-w:
@anrp

Thank you for bringing this to my attention, perhaps this has been fixed in Transformers then!

[...]

PR #254, opened 2026-03-26
fix: "Save as LoRA adapter" option added to prompt selection

alejandroCartas:
The option of saving only the LoRA adapter was added to the obtain_merge_strategy function. Additionally, the prompt selection now shows even if the settings.quantization is not equal to QuantizationMethod.BNB_4BIT, as this might be a bug.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -108,31 +108,34 @@ def obtain_merge_strategy(settings: Settings) -> str | None:
             )
         print()
 
-        strategy = prompt_select(
-            "How do you want to proceed?",
-            choices=[
-                Choice(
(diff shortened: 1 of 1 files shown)


p-e-w:
Did you actually test this? It does **NOT** work when uploading, which is exactly why it has been disabled. See #152.

While I'm aware that it appears to work when exporting locally, as discussed in #152, I will not re-enable this option without understanding why it fails on upload, because there might be situations where the same failure mode happens for local export.

If you can actually find the problem it would be great, but unfortunately, re-enabling a menu is not enough.

p-e-w:
The correct fix is in #321.

PR #285, opened 2026-04-07
fix: support Gemma 4 wrapped linear modules

FeverKing:
## Summary

This PR improves Gemma 4 source-run compatibility in Heretic.

It adds a lightweight `run.py` entrypoint for development use and updates LoRA target discovery so wrapped linear modules used by Gemma 4 are resolved to their underlying `Linear` layers. This avoids PEFT targeting unsupported wrapper modules during adapter injection.

## Changes

- add `run.py` as a simple source development entrypoint
- resolve wrapped linear modules to inner `.linear` modules when collecting abliterable targets
- use exact module paths for LoRA target selection to avoid matching unsupported wrapper modules

## Result

Heretic can now get past LoRA initialization for Gemma 4 models instead of failing on `Gemma4ClippableLinear` target module errors.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -162,22 +162,22 @@ class Model:
         assert isinstance(self.model, PreTrainedModel)
 
         # Always use LoRA adapters for abliteration (faster reload, no weight modification).
-        # Collect actual leaf module names from the model for LoRA targeting.
-        # This is more robust than splitting component keys (e.g. "attn.o_proj" -> "o_proj")
-        # because hybrid models like Qwen3.5 MoE have modules with different names
-        # across layers (e.g. "o_proj" on attention layers, "out_proj" on linear attention layers).
+        # Collect exact module paths from the model for LoRA targeting.
+        # Using full module names avoids matching unsupported wrapper modules
+        # that happen to share the same leaf name (e.g. Gemma4ClippableLinear
+        # wrappers named "q_proj"/"o_proj"), while still supporting hybrid
+        # architectures with differing projection names across layers.
         target_modules_set: set[str] = set()
+        module_id_to_full_name = {
+            id(module): module_name
+            for module_name, module in self.model.named_modules()
+        }
 
         for layer_index, layer in enumerate(self.get_layers()):
-            module_id_to_leaf_name = {
-                id(module): module_name.split(".")[-1]
-                for module_name, module in layer.named_modules()
-            }
-
             for modules in self.get_layer_modules(layer_index).values():
                 for module in modules:
-                    if id(module) in module_id_to_leaf_name:
-                        target_modules_set.add(module_id_to_leaf_name[id(module)])
+                    if id(module) in module_id_to_full_name:
+                        target_modules_set.add(module_id_to_full_name[id(module)])
 
         target_modules = list(target_modules_set)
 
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -0,0 +1,7 @@
+"""Development entry point for running Heretic from source without package installation."""
+
+from heretic.main import main
```
p-e-w: Please don't mix unrelated concerns in a a single PR. This file doesn't belong here.

src/heretic/model.py
```diff
@@ -162,22 +162,22 @@ def _apply_lora(self):
...
+        # Collect exact module paths from the model for LoRA targeting.
+        # Using full module names avoids matching unsupported wrapper modules
+        # that happen to share the same leaf name (e.g. Gemma4ClippableLinear
+        # wrappers named "q_proj"/"o_proj"), while still supporting hybrid
+        # architectures with differing projection names across layers.
         target_modules_set: set[str] = set()
+        module_id_to_full_name = {
+            id(module): module_name
+            for module_name, module in self.model.named_modules()
+        }
```
p-e-w, line 174: Isn't it possible for different layers to use the same name for different modules? For example, different attention implementations that different layers store under the same name.

p-e-w (commented):
Thanks! I assume you have tested this? Have you uploaded a Gemma 4 model processed with this code?

run.py
p-e-w: The recommended method for running Heretic during development is to use a modern package management system like uv, which has this feature built in (`uv run`).

src/heretic/model.py
FeverKing: Yes, exactly. Different layers can reuse the same leaf name for different module types.

That is why I switched this to exact full module paths instead of leaf-name matching: for Gemma 4, PEFT was matching wrapper modules like `o_proj`/`q_proj` and trying to inject into `Gemma4ClippableLinear` instead of the inner `Linear`.

FeverKing:
Yes I have tested it on gemma-4-E2B-it, here is the link: [https://huggingface.co/FeverK/gemma-4-E2B-it-heretic](https://huggingface.co/FeverK/gemma-4-E2B-it-heretic)

run.py
FeverKing: Sorry for that, I didn't notice there is a uv.lock file.
I have already amended the commit, PR should be auto updated too.

p-e-w:
What is the difference between this PR and #287? Which approach do you think is better?

FeverKing:
> What is the difference between this PR and #287? Which approach do you think is better?

I think #287 and this PR solve overlapping but slightly different problems.
#287 improves target selection by switching from leaf-name matching to full module paths. I agree with that change.
What this PR adds on top is handling `Gemma4ClippableLinear` itself by resolving it to the inner `.linear`. In my testing, that part was necessary because PEFT was failing on the wrapper module type, not just on ambiguous suffix matching.
So my current view is: full-path targeting is good, but for Gemma 4 the `.linear` unwrapping is also needed.

p-e-w:
Thanks for the PR. I have decided to merge #287 instead, as @MoonRide303 has correctly identified that the `Gemma4ClippableLinear` modules are only used in the vision and audio towers, which we don't want to touch during abliteration.

PR #293, opened 2026-04-11
fix: make --help return before heavy runtime imports

Diplo2by:
### Summary
---
 
This PR makes heretic --help return immediately by short-circuiting heavyweight startup.

Adds an early -h/--help check in src/heretic/main.py. If help is requested, it invokes Settings() right away so pydantic renders CLI help and exits. This path runs before importing heavy ML/runtime dependencies (torch, transformers, lm_eval, etc.). 

Normal execution flow for non-help invocations is unchanged.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -3,6 +3,20 @@
 
 # ruff: noqa: E402
 
+import sys
+
+from .config import Settings
+
+
+def _is_help_invocation() -> bool:
+    args = sys.argv[1:]
+    return "-h" in args or "--help" in args
+
+
+# Parse and handle CLI help before importing heavyweight ML/runtime dependencies.
+if _is_help_invocation():
+    Settings()  # ty:ignore[missing-argument]
+
 from .progress import patch_tqdm
 
 # This patches tqdm class definitions, which must happen
@@ -13,7 +27,6 @@ import logging
 import math
 import os
 import random
-import sys
 import time
 import warnings
 from dataclasses import asdict
@@ -45,7 +58,7 @@ from rich.table import Table
 from rich.traceback import install
 
 from .analyzer import Analyzer
-from .config import QuantizationMethod, Settings
+from .config import QuantizationMethod
 from .evaluator import Evaluator
 from .model import AbliterationParameters, Model, get_model_class
 from .system import empty_cache, get_accelerator_info


src/heretic/main.py
```diff
@@ -3,6 +3,24 @@
...
+        if arg == "--":
+            break
+        if arg in ("-h", "--help"):
+            return True
+    return False
+
+
+# Parse and handle CLI help before importing heavyweight ML/runtime dependencies.
+if _is_help_invocation():
+    Settings()  # ty:ignore[missing-argument]
```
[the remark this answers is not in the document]
p-e-w: No, this suggestion is nonsense and your original code was correct.

Never follow AI advice blindly.

src/heretic/main.py
```diff
@@ -3,6 +3,25 @@
...
 # ruff: noqa: E402
 
+import sys
+
+from .config import Settings
+
+
+def _is_help_invocation() -> bool:
+    for arg in sys.argv[1:]:
+        if arg == "--":
```
p-e-w: We don't use that.

p-e-w: Since we don't use tail arguments, this whole function can be simplified as

```python
args = sys.argv[1:]
return "-h" in args or "--help" in args
```

src/heretic/main.py
```diff
@@ -3,6 +3,25 @@
...
+            break
+        if arg in ("-h", "--help"):
+            return True
+    return False
+
+
+# Parse and handle CLI help before importing heavyweight ML/runtime dependencies.
+if _is_help_invocation():
+    Settings()  # type: ignore[missing-argument]
+    raise SystemExit(0)
```
p-e-w: What is this exception for? This code should be unreachable.

p-e-w:
Please address the merge conflict.

src/heretic/main.py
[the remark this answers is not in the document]
Diplo2by: Apologies, will make a note of that going forward. The CTA to commit the suggestion was too enticing.

src/heretic/main.py
Diplo2by: Another brain fade stemming from Gemini's suggestion. Apologies.

p-e-w:
CI fails.

p-e-w:
Yup, that does the trick. Merged! Thanks for the PR.

PR #301, opened 2026-04-20
fix: prevent UnboundLocalError when analyzer is not initialized

olekssy:
## Summary
- Fixed an UnboundLocalError in src/heretic/main.py that occurred when the --print-residual-geometry or --plot-residuals flags were not provided.
- The issue was caused by unconditionally calling del analyzer even when analyzer was only initialized within a conditional block.
- Moved the cleanup logic (del and empty_cache()) inside the if needs_full_residuals: block.

## Test plan
- [x] Verified the fix manually by reviewing the logic in src/heretic/main.py.
- [x] Confirmed that the problematic execution path (where needs_full_residuals is False) no longer reaches the del analyzer statement.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -444,9 +444,8 @@ def run():
         if settings.plot_residuals:
             analyzer.plot_residuals()
 
-        # We don't need the residuals after computing refusal directions.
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks for fixing this!

PR #191, opened 2026-02-25
feat: reproducibility when saving & uploading a heretic model

Vinay-Umrethe:
### Implement Reproducibility Suite

Well this PR adds a reproducibility suite that lets users capture the exact environment, packages, OS, and random states used during an abliteration optimization run...

it uses a new `seed` configuration parameter that, when set, enforces deterministic behavior across everything that matters like: Python version, PyTorch (CPU & GPU), Numpy, and the Optuna sampler. If not specified one, heretic now generates a random seed automatically at startup so it can still be recorded for that run.

A NEW (optional) `reproduce/` folder can be created during model export (local save or HF upload). containing the exact TOML configuration, a snapshot of installed package versions, and a summary of the hardware/software environment. 

On the technical side, I'm using a `.safetensors` format for storing RNG states. this avoids those old, insecure pickle-based (`.pkl`) checkpoints which I still use in my non-heretic AI/ML trainings... It stores torch tensors efficiently and wraps the non-tensor states (like python/numpy RNG tuples) into JSON metadata within the secure file header.

For user control, I've added a confirmation prompt `prompt_confirm` so `reproduce/` folder is only generated and uploaded if you explicitly grant permission at the time of export. This builds on the other `prompt_*` helpers I had added in one of the old PR.

**Note:** Right now, the easiest way to reproduce a model you found on the Hub is to just look at the seed value in its `reproduce/config.toml` and specify that when you run Heretic. It's much simpler than trying to load a full binary `.safetensors` state file, though need confirmation on which approchs is better...

Also I wonder if generating a seed at start randomly is a good approch or not because it can affect normal heretic run quality a lot, instead of not generating it, if we go that way then the --seed (easy way would be gone) and would need to implement a --reproduce or something similar to load the suite files... (SO NEED CONFIRMATION TO THIS...)

> for a DEMO take a look at model made with `reproduce/` in the repo: https://huggingface.co/VINAY-UMRETHE/Qwen3-0.6B-heretic

PR #303, opened 2026-04-21
fix: improve the reproducibility system

p-e-w:
This is a pretty major overhaul of #191, aiming to get it ready for the Heretic 1.3 release.

Changes include:

* The user can now choose between uploading reproducibility information with system information, reproducibility information without system information, or no reproducibility information at all.
* Commit handling for models and datasets actually works now.
* Non-relevant fields are stripped from settings before saving them, improving compatibility and privacy.
* Changed the structure of `reproduce.json` with a view to the future.
* Better and more detailed reproduction instructions.
* Configurable shard size for upload (fixes problems with some connections and improves reproducibility).
* Lots and lots of small formatting and correctness fixes.

**Examples:**

* Full reproduction information: https://huggingface.co/p-e-w/Qwen3-4B-Instruct-2507-heretic-REPRODUCTION-TEST-4
* Without system information: https://huggingface.co/p-e-w/Qwen3-4B-Instruct-2507-heretic-REPRODUCTION-TEST-4-no-sysinfo
* Reproduction information disabled: https://huggingface.co/p-e-w/Qwen3-4B-Instruct-2507-heretic-REPRODUCTION-TEST-4-no-info

@Vinay-Umrethe Requesting review!


src/heretic/config.py
```diff
@@ -459,3 +477,47 @@ def settings_customise_sources(
...
+    del essential_settings.chain_of_thought_skips
+
+    del essential_settings.print_responses
+    del essential_settings.print_residual_geometry
+    del essential_settings.plot_residuals
+    del essential_settings.residual_plot_path
+    del essential_settings.residual_plot_title
+    del essential_settings.residual_plot_style
+    del essential_settings.study_checkpoint_dir
+    del essential_settings.benchmarks
```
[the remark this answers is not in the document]
p-e-w: Ah yes, that is probably better!

src/heretic/utils.py
```diff
@@ -330,235 +343,263 @@ def generate_requirements_txt() -> str:
...
     data = {
-        "base_model": {
-            "id": settings.model,
-            "commit_hash": base_model_commit,
-        },
-        "system": {
-            "os": {"platform": platform.platform(), "machine": platform.machine()},
-            "cpu": get_cpu_info_dict(),
-            "python": get_python_env_info_dict(),
+        "version": "1",  # Version number of the reproduce.json file format, to allow for future changes.
```
Vinay-Umrethe, line 550: While we have version number for future proofing `reproduce.json` It seems we should also put a label telling the structure of it. since you introduced `full/basic/none` choices it can be different in two ways.

A simple line telling it's `full` or `basic` (no system info) will be helpful in sorting "heretic grimoire"

system info includes CPU, accelerator info, without it reproducibility is almost never guaranteed. As pytorch versions, build wheels, cuda version, driver versions are highly related.

src/heretic/utils.py
```diff
@@ -330,235 +343,263 @@ def generate_requirements_txt() -> str:
...
+- **KL divergence:** {trial.user_attrs["kl_divergence"]:.6f}
+- **Refusals:** {trial.user_attrs["refusals"]}/{trial.user_attrs["n_bad_prompts"]}
 
-### Accelerators
+{system_report}## Environment
 
-{accelerator_report}
+- **Heretic:** v{version_info.version}{f" (Origin: {version_info.origin})" if version_info.origin else ""}
+- **PyTorch:** {pytorch_version}
+- **Other dependencies:** See [`requirements.txt`](requirements.txt).
```
Vinay-Umrethe, line 511: isn't this line redundant as the 1st line just after it in contents section has requirements.txt info

src/heretic/main.py
```diff
@@ -835,27 +822,48 @@ def count_completed_trials() -> int:
...
+                                )
+                                reproducibility_information = prompt_select(
+                                    "Which reproducibility information do you want to add?",
+                                    [
+                                        Choice(
+                                            title="Full: Settings, package versions, and system information",
+                                            value="full",
+                                        ),
+                                        Choice(
+                                            title="Basic: Settings and package versions",
```
Vinay-Umrethe, line 868:
```bash
This is optional, but valuable to the community as both a learning tool and to preserve computational work already done. 

Guaranteeing reproducibility requires basic system information (Python and OS version, CPU and GPU/accelerator
```

as the `prompt_select` suggest reproducibility is mostly guaranteed when we have BOTH hardware, driver, and software packages match including heretic source.

as we don't store anything like WHO created this model or in which repo this heretic'ed model is/was stored there's no concern of privacy, and even if system info is included, we only collect stuff that's globally generic like anyone can have same GPU, CPU, drivers they used.

I think `basic` info option is not so useful as said before it's almost never guaranteed. this would lead us to create extra logic and maintenance to handle both full and basic kinda of reproducibile models in the "grimoire"

@p-e-w thoughts? I might have missed something probably for which you introduced this

Vinay-Umrethe (commented):
done review of what I catched

Vinay-Umrethe:
I noticed in your conducted tests:

1. `direction_index` value was Null (for basic and full test)

https://huggingface.co/p-e-w/Qwen3-4B-Instruct-2507-heretic-REPRODUCTION-TEST-4-no-sysinfo/blob/main/reproduce/reproduce.json#L227

2. `stepping` value of CPU was Null (for full test)

https://huggingface.co/p-e-w/Qwen3-4B-Instruct-2507-heretic-REPRODUCTION-TEST-4/blob/main/reproduce/reproduce.json#L20

I suppose `py-cpuinfo` must be generic enough to do the same task of getting `stepping` value for both AMD and Intel CPUs even if they use different naming conventions. Or he system had some issue (unlikely) as a number is almost always associated.

Vinay-Umrethe:
> Commit handling for models and datasets actually works now.

Previously was it not working?

src/heretic/config.py
```diff
@@ -31,6 +38,11 @@ class DatasetSpecification(BaseModel):
         description="Hugging Face dataset ID, or path to dataset on disk."
     )
 
+    commit: str | None = Field(
```
Vinay-Umrethe, line 40: shouldn't this be `dataset_commit` as we already have `model_commit` to distinguish properly

src/heretic/utils.py
p-e-w: I think such a field would be redundant because the structure itself implies it: Either the `system` key is present or it isn't. Having an additional field that states whether the system key is there or not doesn't make much sense to me given that we can just check.

`none` is not a choice that can be reflected in `reproduce.json`, as with `none`, no `reproduce.json` is generated at all.

src/heretic/utils.py
p-e-w: I noticed that too, but the fact is that the information legitimately belongs in both places. How would you structure it? The dependencies *are* part of the environment.

src/heretic/main.py
p-e-w: > as we don't store anything like WHO created this model

But we (or rather Hugging Face) *do* store this. The model is linked to the HF account of the uploader.

So this is absolutely a privacy concern. Not everyone is comfortable revealing their system configuration publicly, and if system information is added to the reproducibility folder, they are doing just that. Being able to look up information like OS versions for a user can also have security implications. So the user must have the option to not share that information.

> I think basic info option is not so useful as said before it's almost never guaranteed.

In many, many cases, the model will still be reproducible. It's not strictly guaranteed as you say, but still very likely.

I actually consider this change to be the most important improvement of the entire PR because it allows people to share reproducibility information who wouldn't have been comfortable doing so if it required sharing system details.

src/heretic/config.py
p-e-w: I think the `dataset_` prefix is unnecessary here because we are already within the scope of a `DatasetSpecification` object. For the model this isn't the case, which is why I added the prefix as without it it would be ambiguous.

src/heretic/utils.py
```diff
@@ -214,6 +214,7 @@ def load_prompts(
             # Path is a local directory.
             dataset = load_dataset(
                 path,
+                revision=specification.commit,
```
p-e-w: Ah, I actually see now that I added this to the wrong `load_dataset` invocation.

p-e-w:
> direction_index value was Null (for basic and full test)

Correct. This is how the system expresses the "global" direction (`None` = no specific index). See the `abliterate` method.

> stepping value of CPU was Null (for full test)

Yes, I noticed that. I guess not every CPU provides that information.

> > Commit handling for models and datasets actually works now.
>
> Previously was it not working?

No? Did you intend to add that functionality? Because there was no relevant code at all yet. The `commit` was stored in the settings, but not used when fetching the datasets and model.

src/heretic/utils.py
Vinay-Umrethe: maybe it's very obvious for anyone, so keeping it in any one of them is enough "See [requirements.txt]" is there not not does not matter.

p-e-w:
Note to self: The dependencies are not sorted in `reproduce.json`. This must be fixed.

src/heretic/main.py
Vinay-Umrethe: > have been comfortable doing so if it required sharing system details.

yes it does increase number of people who'll provide the reproducibility suite along with model.

> In many, many cases, the model will still be reproducible. ... but still very likely.

I don't know what kind of cases work making reproducibility work, it's still likely.

> But we (or rather Hugging Face) do store this. The model is linked to the HF account of the uploader.

I think the whole point f reproducibility suite was to ensure, even if model in future gets banned, removed (due to future laws), from Huggingface or any other host, this would be storing tons of such models in heretic library with just few KBs of storage, so it's unlikely we'll also consider storing "from which public repo we took this `reproducible model`" 

(1) As long as someone don't spend their time stalking matching every single detail, 
(2) and as long as we don't store source of the heretic'ed model in "Heretic Grimoire" privacy is maintained.
they can't find the user who made the model unless they dig HF.

src/heretic/config.py
Vinay-Umrethe: Good, but some dev in future might confuse `commit` as a generic one and `model_commit` as specific one, while trying to find "where's `dataset_commit`"

Vinay-Umrethe:
> No? Did you intend to add that functionality? Because there was no relevant code at all yet. The commit was stored in the settings, but not used when fetching the datasets and model.

maybe I haven't checked that time was it being used or not. Correct

src/heretic/config.py
p-e-w: But this follows the convention of the other fields. We also use the name `split`, and not `dataset_split`.

Vinay-Umrethe:
> Note to self: The dependencies are not sorted in `reproduce.json`. This must be fixed.

Not sure what you are talking about, but I do see `requirements` list in `reproduce.json`

src/heretic/main.py
p-e-w: First and foremost we're uploading to HF (which is the only situation in which the `reproduce` folder is currently created). And we must preserve privacy there as well. The user must be in control. There can be no compromise on that.

> so it's unlikely we'll also consider storing "from which public repo we took this reproducible model"

Of course we will store this, otherwise the library will be much less useful. If someone wants a specific model that has been removed from HF, how else will they find it in our library?

p-e-w:
> Not sure what you are talking about, but I do see requirements list in reproduce.json

Yes, but it's not sorted.

p-e-w:
I'm also planning for the final puzzle piece: `heretic --reproduce /path/to/reproduce.json`.

This seems quite complicated to implement, because it will require Heretic to create a temporary venv, install the correct Heretic version and dependencies there, and then quit to hand processing over to that version of Heretic.

Vinay-Umrethe:
> I'm also planning for the final puzzle piece: `heretic --reproduce /path/to/reproduce.json`.

if 1.3 will release after the reproducibility suite is automatically useable that'll take long, 

> This seems quite complicated to implement, because it will require Heretic to create a temporary venv, install the correct Heretic version and dependencies there, and then quit to hand processing over to that version of Heretic.

We can have a `reproduce.py` module to handle it entirely

p-e-w:
> if 1.3 will release after the reproducibility suite is automatically useable that'll take long,

Correct, that part won't be in the 1.3 release most likely.

src/heretic/utils.py
p-e-w: I agree that this looks a bit strange, but I tried removing both references and something is clearly missing then. I prefer a bit of redundancy over confusion.

src/heretic/config.py
```diff
@@ -352,6 +382,12 @@ class Settings(BaseSettings):
             ),
         ],
         description="Benchmarks to offer to the user for evaluating abliterated models.",
+        exclude=True,
+    )
+
+    max_shard_size: int | str = Field(
```
Vinay-Umrethe, line 388: I don't know if This sharding logic is related to reproducibility

p-e-w: The default size may change between Transformers versions (apparently, this has already happened at least once). Storing the size in settings increases the chance of reproducibility if the Transformers version doesn't match.

src/heretic/utils.py
```diff
@@ -272,20 +275,30 @@ def get_trial_parameters(trial: Trial) -> dict[str, str]:
...
-        model_link = f"[{settings.model}](https://huggingface.co/{settings.model})"
 
     version_info = get_heretic_version_info()
+
+    if contains_reproducibility_information:
+        reproducibility_instructions = """
+> [!TIP]
+> **This model is reproducible!**
+>
+> See the [README](reproduce/README.md) in the `reproduce` directory for more information.
```
Vinay-Umrethe, line 293:
```suggestion
> See the [README](reproduce/README.md) in the [reproduce](https://huggingface.co/{settings.model}/tree/main/reproduce) directory for more information.
```

p-e-w: I've thought about that but decided against it because if the model is forked, quantized etc., absolute paths will link to the wrong (parent) model. It's unfortunate that HF doesn't support linking directories properly, but hacks only create more problems.

p-e-w:
Merged! Several remaining issues have been fixed and the exclusion logic for settings in particular is now much cleaner. I have re-tested everything before merging and found no more problems.

Vinay-Umrethe:
> Merged! Several remaining issues have been fixed and the exclusion logic for settings in particular is now much cleaner. I have re-tested everything before merging and found no more problems.

yes i saw all 3 test repos

src/heretic/utils.py
p-e-w: Also, your suggestion is linking to the wrong model anyway, namely to the base model, which doesn't have a `reproduce` directory.

p-e-w:
I'm deleting all my test models now because many people follow my account and I don't want them to be quantized/benchmarked etc.

Vinay-Umrethe:
> I'm deleting all my test models now because many people follow my account and I don't want them to be quantized/benchmarked etc.

Yeah I was thinking why you didn't deleted old tests, as they don't look good 👍

Vinay-Umrethe:
how was `stepping` value issue was fixed in CPU? in previous tests it was null. now it works
@p-e-w

p-e-w:
@Vinay-Umrethe 

It wasn't "fixed", I changed nothing there. I believe some CPUs provide that value, while others don't. I was on a different server so I probably just got one with a CPU that has the `stepping` set.

PR #307, opened 2026-04-21
feat: Support local weight paths and improve memory management for new architectures

antonvice:
This PR improves heretic's ability to handle local model checkpoints and modern architectures like Qwen 3.5.

Changes:
- Added resolve_model_source to automatically map local weight file paths to their parent Transformers checkpoint directories. This allows users to point directly to a specific weight file without the loader failing.
- Improved trust_remote_code handling to ensure decisions are persisted across model reloads.
- Fixed a potential UnboundLocalError in main.py and optimized residual cleanup to prevent VRAM spikes during the analysis phase.
- Set default response_prefix to an empty string to ensure consistent evaluation behavior.

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -147,7 +147,7 @@ class Settings(BaseSettings):
     )
 
     response_prefix: str | None = Field(
-        default=None,
+        default="",
         description=(
             "Common prefix to assume for all responses, so that evaluation happens "
             "at the point where responses start to differ for different prompts. "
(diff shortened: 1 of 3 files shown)


p-e-w:
Thanks for the PR!

> Added resolve_model_source to automatically map local weight file paths to their parent Transformers checkpoint directories. This allows users to point directly to a specific weight file without the loader failing.

Please explain this in more detail. What exactly does this enable, and what problem does it solve?

> Improved trust_remote_code handling to ensure decisions are persisted across model reloads.

They already are. Please describe how this was failing for you.

> Fixed a potential UnboundLocalError in main.py and optimized residual cleanup to prevent VRAM spikes during the analysis phase.

Already fixed on master.

> Set default response_prefix to an empty string to ensure consistent evaluation behavior.

This is wrong. I think you misunderstood how the detection code works.

In general, please address only one issue in one PR. It makes reviewing changes a lot easier.

PR #312, opened 2026-04-25
feat: skip refusal counting for trials with high KL divergence

ghecko:
## Summary

Adds an optional `max_kl_for_refusals` setting that skips the (slow) refusal counting step in `Evaluator.get_score()` for any trial whose KL divergence exceeds the configured threshold. Defaults to unset, preserving the existing behavior.

## Motivation

During an optimization run, every trial pays the cost of generating responses for the full bad-evaluation-prompts set so refusals can be counted. Trials that already produced a high KL divergence have damaged the model significantly and are very unlikely to be Pareto optimal regardless of their refusal count - yet we still spend the time evaluating them.

Skipping refusal counting on these high-KL trials cuts a meaningful amount of wall-clock time off long optimization runs, especially for large models or large `n_trials`.

## Behavior

- When `max_kl_for_refusals` is **unset** (the default), behavior is unchanged: refusals are counted for every trial.
- When `max_kl_for_refusals` is set and a trial's KL divergence is **above** that value:
  - `count_refusals()` is skipped.
  - The trial is recorded with `refusals = len(bad_prompts)` (worst-case).
  - A yellow `Skipping refusal counting (KL > X)` message is printed.

Because the score formula already uses `kld_score = kl_divergence / kl_divergence_scale` (independent of refusals) when `kl_divergence >= kl_divergence_target`, the skipped trial gets a valid score for Optuna to learn from, while being naturally dominated in the Pareto front computation - so it does not appear in the final trial-selection menu.

## Configuration

The setting is exposed through every existing configuration source (no CLI parser changes were needed, since it's a `Settings` field):

```bash
# CLI
heretic --max-kl-for-refusals 1.0 some/model
```

```toml
# config.toml
max_kl_for_refusals = 1.0
```

```bash
# Environment variable
HERETIC_MAX_KL_FOR_REFUSALS=1.0 heretic ...
```

## Safety considerations

I traced every consumer of `refusals` after `get_score()` returns:

- The `score` tuple `(kld_score, refusals_score)` returned to Optuna's multi-objective TPE sampler. When `kl_divergence >= kl_divergence_target`, `kld_score` does not reference `refusals`. Only `refusals_score` does, and the worst-case sentinel value just means the trial is treated as "bad on the second objective" - accurate, since these trials are already bad on the first.
- `trial.user_attrs["refusals"]`, used by the Pareto-front computation in `main.py` (the loop that builds `best_trials`). Because skipped trials always have `refusals = max`, they are always dominated by any non-skipped trial with even one less refusal, and never get added to `best_trials`.

No code path runs after `count_refusals()` inside `get_score()` other than the score assembly, and no other module reads the refusal value through any channel that would behave differently for the sentinel.

## Files changed

- `src/heretic/config.py` - adds the `max_kl_for_refusals` field to `Settings`.
- `src/heretic/evaluator.py` - adds the conditional skip in `get_score()`.
- `config.default.toml` - documents the new option (commented out by default).

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -62,6 +62,12 @@ kl_divergence_scale = 1.0
 # This helps prevent the sampler from extensively exploring parameter combinations that "do nothing".
 kl_divergence_target = 0.01
 
+# If set, skip the (slow) refusal counting step for any trial whose KL divergence exceeds this value.
+# Such trials have already damaged the model significantly, so their refusal count is unlikely to make
+# them Pareto optimal. Skipped trials are recorded with the worst-case refusal count, ensuring they are
+# naturally dominated in the Pareto front. Leave unset (commented out) to always count refusals (the default behavior).
+# max_kl_for_refusals = 1.0
+
 # Whether to adjust the refusal directions so that only the component that is
 # orthogonal to the good direction is subtracted during abliteration.
 orthogonalize_direction = false
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -260,6 +260,16 @@ class Settings(BaseSettings):
         ),
     )
 
+    max_kl_for_refusals: float | None = Field(
+        default=None,
+        description=(
+            "If set, skip the (slow) refusal counting step for any trial whose KL divergence exceeds this value. "
+            "Such trials have already damaged the model significantly, so their refusal count is unlikely to make "
+            "them Pareto optimal. Skipped trials are recorded with the worst-case refusal count, ensuring they are "
+            "naturally dominated in the Pareto front. Leave unset to always count refusals (the default behavior)."
+        ),
+    )
+
     orthogonalize_direction: bool = Field(
         default=False,
         description=(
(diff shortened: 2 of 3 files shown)


p-e-w:
This is how Heretic used to work in the early days. It's a bad idea which is why I removed it long ago, though I fully understand that it can look like a good idea at first glance.

The problem is the assumption hidden in this statement:

> Trials that already produced a high KL divergence have damaged the model significantly and are very unlikely to be Pareto optimal regardless of their refusal count - yet we still spend the time evaluating them.

That's correct, **and it's the right thing to do.** That's because becoming part of the Pareto front isn't the only purpose of a trial. Its other purpose is to guide the optimizer, and it can't do that properly if you return incorrect values to skip refusal counting.

Trials with high KLD aren't "wasted". In fact, they provide critical information that tells the optimizer where to look for promising parameters.

ghecko:
Ah, understood. I missed that the workflow actively relies on those high KLD trials to guide the optimizer instead of just discarding them. I'm closing this PR so, thanks for clarifying!

PR #313, opened 2026-04-26
perf(ara): cache original weights to skip disk reload between trials

ghecko:
## Problem

In ARA mode, each Optuna trial reloads the full model from disk during
`reset_model()`. On the trial trace below (~70 GB model), this single
step dominates wall time:

```
Resetting model...
Loading weights ████████░░░░░░░░░░ 39%  0:07:13
```

The reload exists because `ara_abliterate` optimizes `module.weight`
in-place via LBFGS (`matrix.copy_(get_matrix())`), bypassing the LoRA
adapter mechanism that the non-ARA path uses for fast resets. The
existing fast path in `reset_model()` is gated on `not settings.use_ara`,
so ARA always falls through to `from_pretrained(...)` - disk I/O,
HF dispatch, and re-quantization on every trial.

## Approach

Introduce `ARAWeightCache`: an in-memory snapshot of the original
weights of every module ARA is about to touch. The cache holds the
invariant *every entry stores the pristine original weight*. To
preserve it, the workflow per trial is:

1. `reset_model()` - `cache.restore_all()` writes originals back into
   `module.weight.data`. Model is now clean.
2. `ara_abliterate()` - diff target set vs. cache:
   - drop snapshots for modules outside this trial's
     `[start_layer_index, end_layer_index]` range,
   - snapshot any newly-targeted modules (which now contain originals
     because step 1 just restored them).
3. LBFGS mutates in place. Cache entries remain valid (they're
   independent tensors).

## Storage tier

Per source device: snapshots stay on the source GPU when
`torch.cuda.mem_get_info()` reports enough free memory + 2 GB margin,
otherwise spill to system RAM (`psutil.virtual_memory().available`).
A single accept/reject decision is made per source device, so multiple
weights on the same GPU can't each individually "fit" while
collectively overflowing.

## Fallback behavior

If neither tier has room for the full planned snapshot, no partial
snapshot is taken. Instead, the user sees:

```
Not enough RAM to snapshot all targeted layers (would need ~12.34 GB).
Weights will be reloaded from disk after this iteration...
```

The cache flips to `needs_reload`, and the next `reset_model()` does a
full disk reload (recovering the originals of modules mutated without
backup) before clearing the cache. Subsequent trials retry caching
from a clean slate.

## Drive-by fix: reset_model() recovery from interrupted prior reset

Pre-existing bug, but adjacent to this work and exposed more often by
the trial selection flow:

```
AttributeError: 'NoneType' object has no attribute 'config'
at  current_model = getattr(self.model.config, "name_or_path", None)
```

Reproduced by Ctrl+C during a slow-path `reset_model()` (between
`self.model = None` and `from_pretrained`). The interrupt is caught by
`objective_wrapper`, the study stops, the user picks a trial from the
Pareto menu, and the next `reset_model()` crashes because `self.model`
was never re-assigned.

Fix: track the working dtype on `self._dtype` (set in `__init__` after
the dtype-trying loop and in `_reload_from_disk` after every reload).
At the top of `reset_model()`, if `self.model is None`, jump straight
to `_reload_from_disk(self._dtype)`. The slow path is factored into
`_reload_from_disk()` so both entry points share one implementation.

## What didn't change

- Non-ARA LoRA fast path: unchanged.
- `get_merged_model()` and the post-study restore flow: unchanged.
- ARA optimization itself: unchanged.

## Files

- `src/heretic/model.py` — `ARAWeightCache` class, `_prepare_ara_cache`,
  `_collect_ara_target_modules`, ARA fast path in `reset_model()`,
  `_ara_cache.clear()` after disk reload.
- `src/heretic/main.py` - two `"Reloading model..."` prints replaced
  with `"Resetting model..."`. The disk-reload notice now lives in
  `reset_model()` itself and is only printed when an actual reload
  happens, so the user can tell the fast path apart from the slow path
  in the trial log.

## Expected impact

Per trial in ARA mode: reset cost drops from
`from_pretrained(...)` (tens of seconds to minutes for large models)
to a memcpy of the targeted modules' weights (milliseconds for GPU,
~1 s/10 GB for CPU→GPU on PCIe 4.0). The fallback path is no slower
than the current behavior.

## Testing notes

- `python -m py_compile` on both modified files: passes.
- Logic verification of cache invariant maintained across trials with
  varying `[start_layer_index, end_layer_index]` ranges (overlap, full
  containment, disjoint).
- Behavior under tight memory: warning printed, `needs_reload` set,
  next trial reloads from disk and re-snapshots from clean state.
- Multi-GPU: snapshots placed per source device; restore copies back
  to the original device.
- Quantized (bnb 4-bit): not affected - ARA can't optimize quantized
  weights regardless, so this path assumes FP precision.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -623,7 +623,7 @@ def run():
         for name, value in get_trial_parameters(settings, trial).items():
             print(f"  * {name} = [bold]{value}[/]")
         if settings.use_ara:
-            print("* Reloading model...")
+            print("* Resetting model...")
             model.reset_model()
             print("* Abliterating (Arbitrary-Rank Ablation)...")
             model.ara_abliterate(good_module_io, bad_module_io, ara_parameters)
@@ -811,7 +811,7 @@ def run():
             for name, value in get_trial_parameters(settings, trial).items():
                 print(f"  * {name} = [bold]{value}[/]")
             if settings.use_ara:
-                print("* Reloading model...")
+                print("* Resetting model...")
                 model.reset_model()
                 print("* Abliterating (Arbitrary-Rank Ablation)...")
                 model.ara_abliterate(
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -12,6 +12,7 @@ import torch.linalg as LA
 import torch.nn.functional as F
 from peft import LoraConfig, PeftModel, get_peft_model
 from peft.tuners.lora.layer import Linear
+from psutil import virtual_memory
 from torch import FloatTensor, LongTensor, Tensor
 from torch.nn import Module, ModuleList
 from torch.optim import LBFGS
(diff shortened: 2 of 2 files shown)


p-e-w:
> On the trial trace below (~70 GB model), this single step dominates wall time

If that's true something is unusual about your disk or caching setup. I'm typically seeing 10-20% maximum of total processing time spent on reloading the model.

This PR adds *a lot* of code to what is already the most complex part of the entire program. I have serious concerns about correctness and maintainability to be honest.

ghecko:
I'm testing this on my Spark (an Asus Ascent GX10 -128GB unified memory). It's very possible the issue is specific to how unified memory architectures handle caching compared to standard discrete setups?
With these changes, the massive reloading bottleneck disappears completely on my system. I hear you on the complexity, though; let me know if you have suggestions on how to streamline the implementation.

p-e-w:
While I have never used a Spark, my understanding is that they have various issues related to memory management, including a catastrophic bug where "loading from disk" ends up using twice the amount of VRAM because of memory mapping, which can't be disabled in safetensors.

I don't really have any ideas about how to simplify this. It's not that your implementation is particularly convoluted, it's that the approach itself is too complex without demonstrating that it has substantial and broad positive impact. I'm afraid that improvements on a single type of hardware which is known to be highly problematic with Transformers isn't sufficient to meet that threshold.

PR #317, opened 2026-04-29
fix: preserve auxiliary safetensors shards (MTP / draft heads) on save

timrohrbaugh:
Models that ship auxiliary weights in *separate* safetensors files alongside the main checkpoint are silently dropped by heretic's save path.  Concrete failing case: Qwen/Qwen3.6-27B ships `model-mtp-layer.safetensors` (the Multi-Token Prediction draft head used for vLLM speculative decoding) in addition to its 6 main shards. After running heretic on this model and pushing the result, the MTP shard is gone, and `vllm serve --speculative-config '{"method": "qwen3_next_mtp", ...}'` fails to load:

    ValueError: Following weights were not initialized from checkpoint:
    {'model.pre_fc_norm_embedding.weight',
     'model.layers.0.self_attn.q_norm.weight', ...
     'model.fc.weight', 'model.pre_fc_norm_hidden.weight', ...}

Removing `--speculative-config` makes the same checkpoint load fine. The main model is intact; only the MTP draft head was lost on save.

Cause
-----
`save_pretrained` only persists tensors registered in the loaded model class's `state_dict()`.  When `AutoModelForCausalLM.from_pretrained` resolves the architecture (e.g. `Qwen3_5ForConditionalGeneration`), the MTP weights — destined for the separate `Qwen3_5MTP` drafter class in vLLM — are not part of the loaded state dict.  They're loaded but treated as "unexpected" keys, and `save_pretrained` only writes what is in `state_dict()`, so they vanish on round-trip.

This affects any model with auxiliary heads kept in their own safetensors files: MTP draft heads, EAGLE drafters, Medusa heads, optional vision encoders distributed as separate shards.

Fix
---
Add `copy_auxiliary_safetensors(source_model, dest_dir)` and a Hub counterpart `copy_auxiliary_safetensors_to_hub`.  They:

  1. Read the destination's `model.safetensors.index.json` to identify the main shards heretic just re-saved.
  2. Scan the source for `*.safetensors` files NOT in that set.
  3. Open each candidate and verify its tensor names don't overlap with the main weight map (so we never silently shadow heretic's modifications — if a candidate is a re-shard of the same weights it is skipped).
  4. Copy / push the surviving auxiliary shards verbatim.

These shards aren't on the inference path heretic abliterates, so passing them through unchanged is correct.

Wired into the merged-model save path (local) and merged-model push path (HF Hub).  Adapter saves don't have weight shards so they are not affected.

Verification
------------
Tested against `Qwen/Qwen3.6-27B`:

  - Before: heretic save produces 6 main shards; the original `model-mtp-layer.safetensors` (~6.6 GB) is missing.  vLLM with `--speculative-config qwen3_next_mtp` fails on missing weights.
  - After: heretic save produces 6 main shards plus the original `model-mtp-layer.safetensors` passed through.  vLLM spec decoding loads cleanly.

`model.safetensors.index.json` is unchanged between original and heretic-saved (only main-model tensor keys); the auxiliary shard sits alongside without entering the index.  vLLM's drafter loader scans for `*.safetensors` files in the model directory and finds it.

Backward compatibility
----------------------
  - Single-file models (only `model.safetensors`): no change.
  - Multi-shard main checkpoints with no auxiliary files: no change (all source `*.safetensors` are listed in `weight_map.values()` and are skipped).
  - Models with auxiliary shards: now preserved (previously silently dropped).

No new dependencies; `safetensors` is already required by transformers.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -63,6 +63,8 @@ from .evaluator import Evaluator
 from .model import AbliterationParameters, Model, get_model_class
 from .system import empty_cache, get_accelerator_info
 from .utils import (
+    copy_auxiliary_safetensors,
+    copy_auxiliary_safetensors_to_hub,
     format_duration,
     get_readme_intro,
     get_trial_parameters,
@@ -794,6 +796,11 @@ def run():
                                 del merged_model
                                 empty_cache()
                                 model.tokenizer.save_pretrained(save_directory)
+                                # Pass through any auxiliary safetensors shards
+                                # (MTP draft heads, EAGLE/Medusa heads, etc.)
+                                # that heretic doesn't load itself but downstream
+                                # tooling (vLLM/SGLang spec decoding) needs.
+                                copy_auxiliary_safetensors(settings.model, save_directory)
 
                             print(f"Model saved to [bold]{save_directory}[/].")
 
@@ -896,11 +903,26 @@ def run():
                                     max_shard_size=settings.max_shard_size,
                                     token=token,
                                 )
+                                # Snapshot the main weight keys before freeing
+                                # the merged model — needed by the auxiliary-
+                                # shard copier to detect re-shards of heretic's
+                                # output (different filenames, same content).
+                                main_keys_for_aux = set(merged_model.state_dict().keys())
                                 del merged_model
                                 empty_cache()
                                 model.tokenizer.push_to_hub(
                                     repo_id,
                                     private=private,
+                                    token=token,
+                                )
+                                # Push through any auxiliary safetensors shards
+                                # (MTP draft heads, EAGLE/Medusa heads, etc.)
+                                # that heretic doesn't load itself but downstream
+                                # tooling (vLLM/SGLang spec decoding) needs.
+                                copy_auxiliary_safetensors_to_hub(
+                                    settings.model,
+                                    repo_id,
+                                    main_keys=main_keys_for_aux,
                                     token=token,
                                 )
 
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -180,6 +180,371 @@ def is_hf_path(path: str) -> bool:
     )
 
 
+def _main_safetensors_files(dest_dir: str) -> set[str]:
+    """Set of ``*.safetensors`` filenames listed in the destination's
+    ``model.safetensors.index.json`` ``weight_map``.
+
+    These are the shards heretic itself just re-saved.  Anything else
+    ending in ``.safetensors`` in the source is auxiliary.
+    """
+    index = Path(dest_dir) / "model.safetensors.index.json"
+    if not index.exists():
+        single = Path(dest_dir) / "model.safetensors"
+        return {single.name} if single.exists() else set()
+    try:
+        data = json.loads(index.read_text())
+        return set(data.get("weight_map", {}).values())
+    except Exception:
+        return set()
+
+
+def _safetensors_keys(local_path: str) -> set[str]:
+    """Read the tensor names contained in a safetensors file."""
+    try:
+        from safetensors import safe_open
+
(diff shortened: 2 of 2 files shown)


timrohrbaugh:
## Note: also affects `ara` branch

  `upstream/ara` exhibits the same bug — same `save_pretrained` /
  `push_to_hub` pattern with no auxiliary-file handling. After this fix
  lands on `master`, the same change should be cherry-picked to `ara`
  (or rebase ara onto master if appropriate).

timrohrbaugh:
Addressed all four gemini-code-assist review notes :
HIGH: copy_auxiliary_safetensors_to_hub now accepts main_keys (caller passes set(merged_model.state_dict().keys()) before del merged_model) and uses it for tensor-key overlap check, mirroring the local function's behavior. Catches re-shards of main weights.
MEDIUM: comment punctuation, removed redundant local imports, switched to Path.read_text() for the index file.

timrohrbaugh:
Updated to handle two layouts of auxiliary weights, not one:
1. Separate file (Ling-2.6-flash's model-mtp-layer.safetensors)
2. Embedded keys (Qwen3.6-27B's mtp.* tensors live INSIDE main shards 13 and 15, mixed with main weights)
The original commit only handled (1). Verified against a synthetic Qwen3.6-27B fixture that (2) — where the source's main shards contain a mix of main and mtp.* tensors — was being mishandled: the function would copy entire source shards as if they were aux, dropping un-modified main weights into dest alongside heretic's modifications.
New approach extracts orphan tensors only, writes them to a single canonical model-auxiliary.safetensors, and registers them in the destination's model.safetensors.index.json so loaders that scan the index (vLLM, transformers) find them. Heretic's main-shard modifications are never touched.

p-e-w:
Thanks for the PR.

This is a valiant attempt to fix the issue, but unfortunately, I see no path to ever merging this. This highly complex, multi-step solution is extremely brittle and in my opinion unmaintainable. It will be an endless source of bug reports, half-uploaded models, and who knows what other surprises depending on repository structure. It will also be dead weight every time we want to change anything about the upload process. This isn't the way.

This logic belongs in Transformers (or HF Hub) upstream. Heretic is not a repository management software. Having an idempotent solution for loading and exporting models is an absolutely basic requirement that isn't specific to our needs at all. When we start mucking around with tensor names and JSON listings for the simple task of "save the model we loaded", it should be obvious that something is wrong.

I understand that Transformers doesn't handle this correctly, and I agree that it sucks, but that doesn't mean we're going to riddle Heretic with hundreds of lines of workaround code. Ultimately, it will fall on me to maintain this code, and I'm sorry to say that I just don't feel comfortable committing to that.

timrohrbaugh:
I totally understand.  This has been going on for a while and they have not fixed it.  But I think the one point worth suggesting is that your users are pushing models with dataloss and most will not know it.  As middle ground you might want to small change to detect and warn. No copying, no index editing, no orphan extractions so maybe 20-30 lines of code?  If not I do understand.  I fixed it on my end and just didnt want to not tell you. 
Note for  the detection it should be at the point the loss occured which is not techically on the save.  its on from_pretrained as keys are dropped that dont match ... MTP, Eagle, Vision and Audio encoders... So all those are lost at that point.  maybe warn there?

timrohrbaugh:
I'll push a tool for fixing the models out there for everyone and just make it public. I have 35+ models that needed fixing. Of course all the quants have to be redone. where there is MTP, image or audio.  Which is up to them :)

p-e-w:
> As middle ground you might want to small change to detect and warn. No copying, no index editing, no orphan extractions so maybe 20-30 lines of code?

Just a plain warning that is shown every time, or one that actually detects whether tensors are being dropped? What recommendation should we give to the user in that case? What can they actually do about it?

Are there any other situations besides MTP/speculative decoding where this problem arises?

PR #321, opened 2026-05-05
fix: Reset model after saving merged model

anrp:
The adapter is lost and writes 0-byte adapters if you save an adapter after saving the merged model.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -131,31 +131,26 @@ def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
             )
         print()
 
-        strategy = prompt_select(
-            "How do you want to proceed?",
-            choices=[
-                Choice(
-                    title="Merge LoRA into full model"
-                    + (
-                        ""
-                        if settings.quantization == QuantizationMethod.NONE
-                        else " (requires sufficient RAM)"
-                    ),
-                    value="merge",
+    strategy = prompt_select(
+        "How do you want to proceed?",
+        choices=[
+            Choice(
+                title="Merge LoRA into full model"
+                + (
+                    ""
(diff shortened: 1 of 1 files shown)


src/heretic/main.py
```diff
@@ -736,17 +736,21 @@ def count_completed_trials() -> int:
...
-            model.abliterate(
-                refusal_directions,
-                trial.user_attrs["direction_index"],
-                {
-                    k: AbliterationParameters(**v)
-                    for k, v in trial.user_attrs["parameters"].items()
-                },
-            )
+
+            def reset_trial_model():
```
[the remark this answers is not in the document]

p-e-w:
Yeah, now that I figured out what the problem is, I was actually able to find a discussion confirming just that: https://github.com/huggingface/peft/issues/868#issuecomment-1820642893

Not having an option to disable this behavior is very poor API design.

src/heretic/main.py
```diff
@@ -800,6 +798,7 @@ def count_completed_trials() -> int:
                                 del merged_model
                                 empty_cache()
                                 model.tokenizer.save_pretrained(save_directory)
+                                reset_trial_model()
```
p-e-w, line 804: Maybe add a comment here explaining why you are doing this.

p-e-w:
The fix is correct, but the PR is missing the logic from #308, since that was reverted again.

src/heretic/main.py
anrp: Commented function

anrp:
> The fix is correct, but the PR is missing the logic from #308, since that was reverted again.

Embedded revert-revert in this PR.

p-e-w:
Thanks, merged! Feels good to finally understand what was going wrong before.

PR #324, opened 2026-05-10
fix: count completed trials in progress output

### LINKED ISSUE #311: "Running trial N of M..." where N > M
Reported on Discord. Obviously, this should never happen.

iuyua9:
### Summary
- Count completed trials when assigning the displayed optimization trial number.
- Keep the progress label aligned with the configured completed-trial target after failed or pruned attempts.

### Why
The optimization loop already uses `count_completed_trials()` to decide how many trials remain. However, the displayed `Running trial N of M...` label increments once per objective attempt. If an attempt fails or is pruned and the user later runs additional trials, the attempt counter can exceed the completed-trial target.

This fixes #311 by deriving the displayed trial number from completed trials plus the currently running trial.

### Validation
- `python3 -m py_compile src/heretic/main.py`
- `uv run ruff check src/heretic/main.py`

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -487,7 +487,10 @@ def run():
 
     def objective(trial: Trial) -> tuple[float, float]:
         nonlocal trial_index
-        trial_index += 1
+        trial_index = (
+            sum(1 for t in trial.study.trials if t.state == TrialState.COMPLETE)
+            + 1
+        )
         trial.set_user_attr("index", trial_index)
 
         direction_scope = trial.suggest_categorical(
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -487,10 +487,7 @@ def run():
 
     def objective(trial: Trial) -> tuple[float, float]:
         nonlocal trial_index
-        trial_index = (
-            sum(1 for t in trial.study.trials if t.state == TrialState.COMPLETE)
-            + 1
-        )
+        trial_index = len(trial.study.trials)
         trial.set_user_attr("index", trial_index)
 
         direction_scope = trial.suggest_categorical(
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks for the PR! Could you explain the failure mode in some more detail? What exactly are the conditions under which this counter anomaly happens? Can you give steps to reproduce this problem?

iuyua9:
Thanks for the question. The failure mode I was trying to cover is when Optuna invokes `objective()` for attempts that do not become completed trials, for example an interrupted, failed, or pruned attempt during a longer run. The old local counter increments before the trial outcome is known, so later progress labels can drift from the number of completed results the user is actually looking at.

A minimal repro path is:

1. run a study with more than one trial,
2. have one objective attempt terminate before it reaches `TrialState.COMPLETE` (for example by raising during the objective),
3. continue the study with another trial.

Before this change, the next successful trial label includes the failed/pruned attempt because the counter was already incremented. With this change, the label is derived from completed trials in the current study, so it stays aligned with completed results rather than attempted invocations.

p-e-w:
Thanks for the explanation, but I don't think this logic does what we want.

Let's say a trial fails. Then with your implementation, we see this (simplified):

```
Running trial 63 of 200...
Error: blah blah blah.

Running trial 63 of 200...
...
```

That's not what I'd expect. The next trial should be numbered 64, even if the previous trial failed. The previous trial existed, and shouldn't just vanish in indexing.

(Also, the idiomatic way to count items is to use `len`, not a sum of `1`s, but that's secondary.)

iuyua9:
Thanks, that makes sense. I pushed a follow-up that numbers attempted trials with , so a failed trial no longer causes the next attempt to reuse the same visible number. I also changed the completed-trial helper to use  rather than summing sentinel values.\n\nValidation run locally:\n\n```bash\npython -m compileall -q src/heretic/main.py\ngit diff --check\n```

iuyua9:
Thanks, that makes sense. I pushed a follow-up that numbers attempted trials with `len(trial.study.trials)`, so a failed trial no longer causes the next attempt to reuse the same visible number. I also changed the completed-trial helper to use `len(...)` rather than summing sentinel values.

Validation run locally:

```bash
python -m compileall -q src/heretic/main.py
git diff --check
```

p-e-w:
But what problem does this code fix now? With your latest change, in what situation is the displayed count different from what it was before?

p-e-w:
The correct fix was found in #357. Closing this PR.

Issue #221, opened 2026-03-10
[Needs testing]: Kimi-K2.5 Abliteration Research (TL;DR: Standard abliteration does NOT work on Kimi K2.5.)


kabachuha:
This is a (not mine) repost of https://huggingface.co/hamsaOmar/Kimi-K2.5-abliterated here to consider the implications for heretic.

---

# Kimi-K2.5 Abliteration Research

**TL;DR: Standard abliteration does NOT work on Kimi K2.5.** This repo documents the first systematic attempt to abliterate the largest open-source multimodal MoE model (1T params) and explains why it fails. Includes computed refusal directions and scripts for reproduction.

## Key Finding

**Kimi K2.5's safety training is fundamentally resistant to linear abliteration.** On the standard `mlabonne/harmful_behaviors` test set:

| Approach | Refusal Rate | Quality |
|----------|-------------|---------|
| Original model (no modification) | 100% | 100% |
| Weight-baked ablation (all layers) | 98-100% | OK |
| **Inference-time hooks (rank-1)** | **100%** | **100%** |
| Inference hooks (rank-5 subspace) | 67% (but garbled) | Degraded |
| Combined (weight-baked + hooks) | 100% | OK |

Quality = 7/7 standard benchmarks (math, reasoning, code, knowledge, creative, Chinese, instruction following).

**On handpicked softer prompts** (lock picking, wifi hacking), the hooks reduce refusals from ~100% to ~83%. But on the standard harmful_behaviors dataset, abliteration has zero effect.
## Why Does This Happen?
K2.5 uses **DeepseekV3 MoE architecture** with 384 routed experts (top-8 routing) per layer. Our analysis suggests:
1. **Refusal IS one-dimensional in activation space** — SVD shows 50.7% of refusal variance in a single direction. The refusal direction is correctly identified (cosine similarity 0.88 between two independent computation methods).
2. **But projecting it out doesn't change behavior** — The model re-introduces refusal through deeper mechanisms:
   - Expert routing may encode refusal in the *selection* of which experts to activate, not just in the residual stream
   - Attention patterns may carry refusal signals independently of the residual stream direction
   - K2.5's safety training appears to be more robust than K2 (which was successfully abliterated by huihui-ai)
3. **MoE expert routing is the key difference** — Standard abliteration works on dense models (Llama, Mistral, etc.) because there's a single pathway. MoE models have 384 experts per layer — refusal can be encoded in which experts fire, not just what they compute.
## What's In This Repo
[...]

Issue #192, opened 2026-02-25
[Refactor] vLLM as the backend


kabachuha:
Currently there is frustration that the default transformers implementation of models is very inefficient. For example, GLM Air 106b 4bit on a single H100 does <5 t/s bs1 in my experiments, however, in llama.cpp it makes 20+ t/s in q4 quantization. See https://github.com/p-e-w/heretic/issues/153 for details. vLLM greatly utilizes tensor parallelism, friendly for multi-GPU deployments of very large language models (GLM, Kimi) and allowing to abliterate them in much less time and money.

vLLM is built on top on transformers and it fully supports [hidden states extraction](https://github.com/vllm-project/vllm/blob/main/docs/features/speculative_decoding/speculators.md) (and also community https://github.com/agencyenterprise/vllm-hidden-states) from any layer and token. Moreover, it provides the end [logprobs](https://docs.vllm.ai/en/latest/api/vllm/logprobs/), so KL divergence can be calculated. Finally, [vLLM allows applying LoRAs on the fly](https://github.com/vllm-project/vllm/blob/main/docs/features/lora.md), which is the main point of this refactor, because just like in transformers it doesn't require re-loading the model and wasting a huge amount of time.

This perfectly works with the heretic's requirements (KLD, refusal counter, dynamic lora updates)

This is a huge refactor, but still is something to consider if we want to move forward to work with larger models and faster iteration.

LLaMA.cpp is not fitting here actually, because it doesn't support on the fly loras and low precision hidden states collection is quite complicated in my experience

p-e-w:
That's not going to happen for the foreseeable future.

vLLM still lags *far* behind Transformers in terms of model compatibility. That newly released models often come with a vLLM implementation is a relatively recent phenomenon.

Also, Heretic is going to soon have features that require model introspection beyond hidden states, so what vLLM provides is going to be flat out insufficient for our needs.

Perhaps in a year or so this can be revisited as the ecosystem evolves, but currently, vLLM is not yet ready to power Heretic.

kabachuha:
@p-e-w 

> That's not going to happen for the foreseeable future.

Yes, it quite expected :) Just a feed for thought. I wish though more speed up implementation existed.

> require model introspection beyond hidden states

Looking forward to it!

PR #332, opened 2026-05-25
feat: ARA, but it's LoRA

kabachuha:
# ARA, but it's LoRA

This replaces the inplace training of a layer's matrix with optimization of a product of two low rank matrices which make an additive to the weight.

It eliminates the need for model reload from disk, accelerating the process, and, crucially, 4 bit bitsandbytes compatibility (impossible for normal ARA), making it available for heretic to work upon Gemma 4 31b it at only ~36 GB VRAM, accessible for most duo-gpu prosumer setups.

Due to it being a preset-rank LoRA, it's technically not "arbitrary rank", but this can be simulated with the LoRA having a high enough rank. This is also compatible with row normalisation preservation because the weight is corrected with the norm of its dequantization.

During optimization, there can be high KLs, especially at the start, but they drift to adequate values later.

> 05-26: ~~after Gemini's suggestions high KLs occur more often, I'm investigating this.~~
> 05-27: Reverted Gemini's suggestions. I think the "numerical stability" is actually the false trap where the model tries to squeeze something in an unstable region instead of crashing fast and simple proceeding to a better region. Now there are much fewer KL spikes.

I think this is a must-have for heretic for three reasons:

1. Independence - you don't need to depend on the few HF abliterators who might disappear at any moment.

2. Locality - no need to rent high end hardware and download a huge model back and forth to a server.

3. Faster iteration - if you are testing or making a models merge, the obscure HF component models might not be abliterated and you need to rely just on yourself to do this.

3*. Speed of the process itself - the mode doesn't require re-reading the model from the storage, giving a massive speed-boost for slower file systems.

Just as shown with normal LoRA-based abliteration, the quantization/low-rank trade-off should be neiligible.

Here is the eighth trial of my experiment. (because I'll go to sleep soon, so it's a very quick demo)

Снимок экрана 2026-05-25 221620

Feel free to test the code, make critiques and suggestions. The mode is fully additive and doesn't conflict with the default LoRA and ARA switches.

cc @p-e-w

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -203,6 +203,18 @@ class Settings(BaseSettings):
             "instead of traditional directional ablation."
         ),
     )
+    
+    use_ara_lora: bool = Field(
+        default=False,
+        description=(
+            "Use LoRA in ARA instead of full-weight editing. Makes it compatible with quantization and removes model reloads."
+        ),
+    )
+    
+    ara_lora_rank: int = Field(
+        default=128,
+        description="If LoRA is used in ARA, this sets up its rank. Keep it high enough to simulate the 'arbitrary' effect",
+    )
 
     use_piqa: bool = Field(
         default=False,
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -213,7 +213,7 @@ class Settings(BaseSettings):
     
     ara_lora_rank: int = Field(
         default=128,
-        description="If LoRA is used in ARA, this sets up its rank. Keep it high enough to simulate the 'arbitrary' effect",
+        description="If LoRA is used in ARA, this sets up its rank. Keep it high enough to simulate the 'arbitrary' effect.",
     )
 
     use_piqa: bool = Field(
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -716,11 +716,6 @@ class Model:
                     lora_A = cast(Tensor, module.lora_A["default"].weight)
                     lora_B = cast(Tensor, module.lora_B["default"].weight)
 
-                    # Create float32 copies for stable optimization.
-                    # LBFGS is numerically unstable when optimizing float16/bfloat16 parameters directly.
-                    A_opt = lora_A.detach().float().clone().requires_grad_(True)
-                    B_opt = lora_B.detach().float().clone().requires_grad_(True)
-
                     # Data preparation.
                     # Move I/O tensors to the device of the adapter weights.
                     good_input, good_output = good_module_io[layer_index][component][module_index]
@@ -736,7 +731,7 @@ class Model:
                         # Calculate effective weight: W_eff = W_base + B @ A.
                         W_eff = W_base + (B @ A)
 
-                        # Apply Row Normalization (keep original norms)
+                        # Apply Row Normalization (keep original norms).
                         if self.settings.row_normalization == RowNormalization.FULL:
                             # Normalize to unit length, then scale by original norms.
                             W_eff = F.normalize(W_eff, p=2, dim=1) * W_row_norms
@@ -773,7 +768,7 @@ class Model:
                     # Optimization loop.
                     # We optimize A and B, not the base matrix.
                     optimizer = LBFGS(
-                        [A_opt, B_opt],
+                        [lora_A, lora_B],
                         lr=1.0,
                         max_iter=20,
                         history_size=10,
@@ -783,18 +778,13 @@ class Model:
                     def closure():
                         optimizer.zero_grad()
                         # Pass the actual tensors being optimized to the objective.
-                        loss = objective(A_opt, B_opt)
+                        loss = objective(lora_A, lora_B)
                         loss.backward()
                         return loss
 
                     # Run optimization steps.
                     for step in range(5):
                         optimizer.step(closure)
-
-                    # Copy the optimized weights back to the original parameters.
-                    with torch.no_grad():
-                        lora_A.copy_(A_opt.to(lora_A.dtype))
-                        lora_B.copy_(B_opt.to(lora_B.dtype))
 
     def generate(
         self,
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -161,7 +161,7 @@ class Model:
         if self.model is None:
             raise Exception("Failed to load model with all configured dtypes.")
 
-        if not settings.use_ara:
+        if not settings.use_ara or settings.use_ara_lora:
             self._apply_lora()
 
         # LoRA B matrices are initialized to zero by default in PEFT,
@@ -188,21 +188,24 @@ class Model:
         # because hybrid models like Qwen3.5 MoE have modules with different names
         # across layers (e.g. "o_proj" on attention layers, "out_proj" on linear attention layers).
         target_modules_set: set[str] = set()
+        
+        module_id_to_full_name = {
+            id(module): module_name
+            for module_name, module in self.model.named_modules()
+        }
 
-        for layer_index, layer in enumerate(self.get_layers()):
-            module_id_to_leaf_name = {
-                id(module): module_name.split(".")[-1]
-                for module_name, module in layer.named_modules()
-            }
-
+        for layer_index in range(len(self.get_layers())):
             for modules in self.get_layer_modules(layer_index).values():
                 for module in modules:
-                    if id(module) in module_id_to_leaf_name:
-                        target_modules_set.add(module_id_to_leaf_name[id(module)])
+                    full_name = module_id_to_full_name.get(id(module))
+                    if full_name is not None:
+                        target_modules_set.add(full_name)
 
-        target_modules = list(target_modules_set)
+        target_modules = sorted(target_modules_set)
 
-        if self.settings.row_normalization != RowNormalization.FULL:
+        if self.settings.use_ara_lora:
+            lora_rank = self.settings.ara_lora_rank
+        elif self.settings.row_normalization != RowNormalization.FULL:
             # Rank 1 is sufficient for directional ablation without renormalization.
             lora_rank = 1
         else:
@@ -224,7 +227,10 @@ class Model:
         # so the result is a PeftModel rather than a PeftMixedModel.
         self.model = cast(PeftModel, get_peft_model(self.model, self.peft_config))
 
-        print(f"* LoRA adapters initialized (targets: {', '.join(target_modules)})")
+        display_targets = sorted({name.rsplit(".", 1)[-1] for name in target_modules})
+        print(
+            f"* LoRA adapters initialized (target types: {', '.join(display_targets)})"
+        )
 
     def _get_quantization_config(self, dtype: str) -> BitsAndBytesConfig | None:
         """
@@ -311,7 +317,7 @@ class Model:
         if (
             current_model == self.settings.model
             and not self.needs_reload
-            and not self.settings.use_ara
+            and (not self.settings.use_ara or self.settings.use_ara_lora)
         ):
             # Reset LoRA adapters to zero (identity transformation)
             for name, module in self.model.named_modules():
@@ -341,7 +347,7 @@ class Model:
             **extra_kwargs,
         )
 
-        if not self.settings.use_ara:
+        if not self.settings.use_ara or self.settings.use_ara_lora:
             self._apply_lora()
 
         self.needs_reload = False
@@ -667,6 +673,118 @@ class Model:
 
                     with torch.no_grad():
                         matrix.copy_(get_matrix())
+
+    def ara_lora_abliterate(
+        self,
+        good_module_io: ModuleIO,
+        bad_module_io: ModuleIO,
+        parameters: ARAParameters,
+    ):
+        for layer_index in range(
+            parameters.start_layer_index,
+            parameters.end_layer_index,
+        ):
+            for component, modules in self.get_layer_modules(layer_index).items():
+                for module_index, module in enumerate(modules):
+                    # Cast to Linear to access weights and LoRA adapters
+                    module = cast(Linear, module)
+
+                    # --- 1. Base Weight Handling & Dequantization ---
+                    # We need the base weight in float32 to compute the effective weight
+                    base_weight = cast(Tensor, module.base_layer.weight)
+                    quant_state = getattr(base_weight, "quant_state", None)
+
+                    if quant_state is None:
+                        W_base = base_weight.to(torch.float32)
+                    else:
+                        # Maintain the original dequantization logic for bitsandbytes
+                        W_base = cast(
+                            Tensor,
+                            bnb.functional.dequantize_4bit(
+                                base_weight.data, 
+                                quant_state
+                            ).to(torch.float32),
+                        )
+
+                    # --- 2. Row Normalization Setup ---
+                    # Pre-calculate the original row norms to preserve them
+                    # This implements the RowNormalization.FULL logic
+                    W_row_norms = LA.vector_norm(W_base, dim=1, keepdim=True)
+
+                    # --- 3. Adapter Target Identification ---
+                    # We optimize the LoRA weights A and B
+                    lora_A = cast(Tensor, module.lora_A["default"].weight)
+                    lora_B = cast(Tensor, module.lora_B["default"].weight)
+
+                    # --- 4. Data Preparation ---
+                    # Move I/O tensors to the device of the adapter weights
+                    good_input, good_output = good_module_io[layer_index][component][module_index]
+                    bad_input, bad_output = bad_module_io[layer_index][component][module_index]
+
+                    good_input = good_input.float().to(lora_A.device)
+                    good_output = good_output.float().to(lora_A.device)
+                    bad_input = bad_input.float().to(lora_A.device)
+                    bad_output = bad_output.float().to(lora_A.device)
+
+                    # --- 5. The Objective Function ---
+                    def objective(A: Tensor, B: Tensor) -> Tensor:
+                        # Calculate effective weight: W_eff = W_base + B @ A
+                        W_eff = W_base + (B @ A)
+
+                        # Apply Row Normalization (keep original norms)
+                        if self.settings.row_normalization == RowNormalization.FULL:
+                            # Normalize to unit length, then scale by original norms
+                            W_eff = F.normalize(W_eff, p=2, dim=1) * W_row_norms
+
+                        # Compute outputs using the effective weight
+                        new_good_output = good_input @ W_eff.T
+                        new_bad_output = bad_input @ W_eff.T
+
+                        # The original ARA loss function
+                        preserve_good_behavior = (
+                            (new_good_output - good_output) ** 2
+                        ).mean()
+
+                        steer_bad_behavior = (
+                            mean_distances_to_knn(
+                                new_bad_output,
+                                good_output,
+                                parameters.neighbor_count,
+                            ).mean()
+                            + parameters.overcorrect_relative_weight
+                            * -mean_distances_to_knn(
+                                new_bad_output,
+                                bad_output,
+                                parameters.neighbor_count,
+                            ).mean()
+                        )
+
+                        return (
+                            parameters.preserve_good_behavior_weight
+                            * preserve_good_behavior
+                            + parameters.steer_bad_behavior_weight * steer_bad_behavior
+                        )
+
+                    # --- 6. Optimization Loop ---
+                    # We optimize A and B, not the base matrix
+                    optimizer = LBFGS(
+                        [lora_A, lora_B],
+                        lr=1.0,
+                        max_iter=20,
+                        history_size=10,
+                        line_search_fn="strong_wolfe",
+                    )
+
+                    def closure():
+                        optimizer.zero_grad()
+                        # Pass the actual tensors being optimized to the objective
+                        loss = objective(lora_A, lora_B)
+                        loss.backward()
+                        return loss
+
+                    # Run optimization steps
+                    for step in range(5):
+                        optimizer.step(closure)
 
     def generate(
         self,
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -686,18 +686,18 @@ class Model:
         ):
             for component, modules in self.get_layer_modules(layer_index).items():
                 for module_index, module in enumerate(modules):
-                    # Cast to Linear to access weights and LoRA adapters
+                    # Cast to Linear to access weights and LoRA adapters.
                     module = cast(Linear, module)
 
-                    # --- 1. Base Weight Handling & Dequantization ---
-                    # We need the base weight in float32 to compute the effective weight
+                    # Base weight handling and dequantization.
+                    # We need the base weight in float32 to compute the effective weight.
                     base_weight = cast(Tensor, module.base_layer.weight)
                     quant_state = getattr(base_weight, "quant_state", None)
 
                     if quant_state is None:
                         W_base = base_weight.to(torch.float32)
                     else:
-                        # Maintain the original dequantization logic for bitsandbytes
+                        # Maintain the original dequantization logic for bitsandbytes.
                         W_base = cast(
                             Tensor,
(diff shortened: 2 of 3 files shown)


p-e-w:
Hey, you're back!

As you may have heard, I am currently battling legal threats from megacorporations, as well as negative portrayals in the global press, so it may take a few weeks before I can give this PR the attention it deserves. :laughing:

I had already planned to implement optional LoRA support for ARA myself, but by factoring the matrix after the fact (like we do with MPOA), which is inferior to your approach of optimizing the factors directly. During tests, I had observed effective ranks around 10-30 typically, with the corresponding singular values explaining 99+% of the variance with the models I tested. This is still interesting from a research PoV, because it hints at a highly non-linear (if not full-rank) refusal manifold.

Given how many advantages direct LoRA optimization has, I think we should probably remove the option to do full-rank optimization entirely. It just adds complexity and I doubt there will be a difference in practice.

As I said, I will get back to this soon. There is also still a stability issue with ARA, which I believe is caused by the fact that the gradient is not directionally continuous because only the nearest `k` neighbors are taken into account, which jumps discretely when their identities change. Do you have an idea for improving this? Also, the convergence instabilities/preconditions leading to extremely high KLDs need to be investigated. Perhaps we can write a paper together at some point.

In the meantime, please keep backups of everything locally, in case this repository becomes unavailable.

kabachuha:
Sure, thank you for review!

I'm going to address Gemini's comments later in the day and make more empirical tests of your suggestions. (Including high KL instabilities)

Still, merging it before into the ARA branch as an experimental option would be helpful as many people are using that branch and the model reload / lack of quantization are adding high overhead. May I ping you after I make enough tests anyway? (Like, on multiple models and edge cases + Gemini comments)

---

I saw the unfolding of the stuff on Reddit a few hours ago. I always advice to store as much as one can: repos, models and datasets. TBH, the HF datasets are probably even more endangered species (copyright, unfiltered NSFW, or both) and there have already been widely scandalous cases of them being removed. Sadly, most of the datasets are too large for local storage, I wish more people torrented them, because unlike heretic models (they can be made at home), there are almost no solutions of dataset restoration

kabachuha:
As for research, just before this pull request, I found absolutely **no** references to the ARA technique Heretic uses nowadays, having searched Semantic Scholar and Googled arXiv. Although there were some SOM papers, referencing my previous pull request :) About ARA I found no mentions, showing that it either has low explainability, or high explainability, or it's working too well and is not defeatable on the benchmarks

rootacc3ss:
You had referenced this in a thread I am subscribed to. 

I'm quite curious to see if we can finally and easily produce a good K2.5 or 2.6 model with less censorship (without using the method that ruins the models, apparently...)

I'll look in to this further, and will watch this thread. Again, I'd be happy to help with funding or providing some compute if possible when it's needed. I can under limited circumstance, and it's def not technically out of my own pocket right now (I wish that were possible at the moment, I'd love to find that hah), but I'll try to see if I can make it work. 

@p-e-w regarding your mention of the mongoloids trying to sue you and drag you through the dirt, fuck them. Legally, you're sound, brother. If you need me to pass something by my lawyer or family within law, let me know. Keep it up, and if you need to "pass the torch" meaning grant another repository that "definitely isn't maintained or owned by you" ownership of the repo, never be afraid to make a second account and clear yourself of that burden ;)

rootacc3ss:
Also, second little bit of commentary regarding the unfortunate situation regarding archiving things of this nature:

MEGA has a MADLY cheap S3 compatible storage offering. Definitely worth a look. I'm pretty sure it's the cheapest option out there.

p-e-w:
> I found absolutely no references to the ARA technique Heretic uses nowadays

I mean, I invented ARA from scratch just a few months ago, haven't described it in detail yet, and it's only implemented in an unmerged pull request, not in mainline, so I wouldn't expect papers to reference it already. I'm subscribed to relevant SemanticScholar streams and some recent papers still use the bog standard all-layer directional ablation from 2 years ago.

As for project resilience, we are going for IPFS for release distribution (already implemented), the upcoming Heretic Grimoire system for model preservation (more details soon), and decentralized infrastructure wherever possible. Please contribute ideas to #330. Also, unofficial public Git mirrors are always welcome, especially if they are located in non-US (and ideally non-Western) jurisdictions!

> Sadly, most of the datasets are too large for local storage

That's true in general, but fortunately not for the standard prompt datasets used by Heretic (especially not if it's only the first few hundred prompts like we use). We need backups of those that aren't on HF.

kabachuha:
I found one (general) issue. If there is a high KL, low refusal (e.g. KL > 1.0, refusal < 10) region, after the 60th turn (random testing's end), the optimizer might think the refusals reduction outweights the KL and get stuck there, with KL being > 1.0 on these trials. We need to penalize these regions even harder somehow or to exclude them.

p-e-w:
We need to understand under what conditions those trials with KLD = 15 or so happen. I'm sure it's some kind of runaway behavior where the gradient pushing away from the bad residuals outweighs the gradient holding the good residuals in place. This problem does not seem to occur with `overcorrect_relative_weight` < 1, but the best trials often have `overcorrect_relative_weight` > 1 so we can't just constrain the range.

> I think the "numerical stability" is actually the false trap where the model tries to squeeze something in an unstable region instead of crashing fast and simple proceeding to a better region. Now there are much fewer KL spikes.

So with worse precision we get better results? That's suspicious to say the least.

kabachuha:
I've been running the tests whole day right now, so I'm going to update this. My new suspicion is the high KL low refusal outweight logic, so for the next run I'm patching it to count KL > 1.0 as 100/100 refusals to cut-off that region completely. I will update here as the tests complete. If it works more or less in bf16, I can try fp32 again

kabachuha:
Results after a BF16 run. I'd say, pretty optimistic.

image

To prevent the optimizer stuck at KL > 1, I added a reset to put the refusals count to 100 in evaluator.py and it helped much

```python
        if kl_divergence > 1.01:
            refusals = len(self.bad_prompts)
            print("  * KL OVERFLOW")
        else:
            print("  * Counting model refusals...")
            refusals = self.count_refusals()
```

For comparison, [coder3101's gemma 31b it heretic](https://huggingface.co/coder3101/gemma-4-31B-it-heretic) has 15/100 refusals and 0.0434 KL. (we have 16/100 and 0.07 KL) We have greater KL, but on the other side we did it with only 4bit precision in ~36 GB VRAM fully at home. And I ran about 160 Heretic trials, with more optimization and patience it could hit better values.

For subjective chatting tests, the uncensored aspect is fully here, so it's not just numbers on screen. If I'll have time, I'm also going to launch the FP32 version, maybe it can be even better (though I'm skeptical about it)

kabachuha:
FP32 run. Looks worse, so won't include it yet

image

kabachuha:
Btw, in the current code the adapter model (PEFT LoRA format) is saved in the folder at the end for some reason, despite the "Merge LoRA into the model" option selected. I will look at the code to truely merge the LoRA. Meanwhile, the adapter can be merged with a simple "offline" script if you want to test it sooner

p-e-w:
> I'm patching it to count KL > 1.0 as 100/100 refusals to cut-off that region completely.

This is tempting, but it's the wrong approach.

TPE relies on two fundamental, intuitive assumptions, which follow from its formulation via Gaussian Mixture Models:

1. Parameters that are close together produce objective values that are close together
2. Parameters aren't strongly interdependent

Your approach breaks both of these assumptions and will **ON AVERAGE** lead to worse convergence behavior. You're introducing an artificial cliff in the objective landscape that TPE cannot comprehend, because the returned refusal count is actually incorrect.

Here's an incredibly important thing to understand: TPE is *highly* random. It's almost impossible to conclude anything about its behavior from empirical observations when doing only a few studies. You'd need hundreds of complete studies, with potentially hundreds of runs each, to do that. Running a study, changing something, then running the study again and comparing the Pareto fronts tells you basically nothing. I have fallen into this trap myself many, many times.

&nbsp;

### So what is the correct solution?

Optuna has a built-in mechanism for telling the optimizer that a trial is "unacceptable":

```python
raise optuna.TrialPruned()
```

This is a special exception designed for this purpose. You're telling Optuna that this was a bad trial *without* lying about the objective value.

However, this is just a hack in this case. It's still not a proper solution. That's because the real problem is that those ultra-high-KLD parameters exist in the first place. The objective landscape ***is*** extremely unstable in some regions. Such landscapes are unsuitable for optimization with TPE. There is only one proper solution: We need to understand what causes this runaway behavior, and fix it in ARA.

p-e-w:
> Btw, in the current code the adapter model (PEFT LoRA format) is saved in the folder at the end for some reason, despite the "Merge LoRA into the model" option selected. I will look at the code to truely merge the LoRA. Meanwhile, the adapter can be merged with a simple "offline" script if you want to test it sooner

LoRA export has been fixed on master, but merging is not trivial because of the reproducibility system.

kabachuha:
I didn't really look into the internals of optuna or TPE, this is just for reference in the PR context. LoRA and how to prune properly are the future goals, I think. For now, I'd like to test multiple architectures like Qwen or previous Gemmas for more stats

If/when you have time to test the PR, feel free to do it and write suggestions!

Because currently with Gemma 4 31b I have ~8 hours to use for a full run (a whole day/night)

p-e-w:
I recommend using much smaller models for testing. 4B models are usually intelligent enough to have robust refusal behavior. Anything larger is just wasting time during development.

kabachuha:
Here I target large models specifically to test the 4bit bitsandbytes compatibility, LoRA and quantization effects.

As for 4b models, I indeed want to test Qwen3 4b with LoRA, but no quant.

p-e-w:
FYI, the currently planned merge order is as follows:

1. #326 
2. Release Heretic 1.4
3. #53
4. #138
5. Implement modifier plugins based on #53 and #138
6. Refactor ARA into a modifier plugin
7. Merge ARA (#211 + #332)
8. ...
9. Release Heretic 2.0

kabachuha:
Got it.

I think, I'm going to be dumping the results/code corrections/observations in the PR anyway, if it would be helpful

Btw, what about SOMA? Are we discarding it or will leave as an option / plugin? I read at least one paper citing my implementation of it in Heretic, so it may be helpful for researchers

p-e-w:
SOMA should also be rewritten as a modifier plugin once the plugin infrastructure is in place, and I'd be happy to include it as a default plugin in Heretic.

The existing implementations (abliteration, SOMA, ARA) will inform the API of the modifier plugin system.

kabachuha:
Some research into the ARA instability.

I did a bit of analysis and fit decision trees over the optuna logs from Gemma 4 trials and 250 Qwen 3 4b instruct 2007 trials (200 trials random optuna "warmup" search + 50 normal optimization, FP32, no quantization)

As I see, the key dependence is on steer bad behavior and overcorrect relative weight. It also almost doesn't depend on the start/end layers (for Qwen surely). It's not much conclusions, but may be helpful.

## Gemma decision tree

gemma-rule

Decision tree rules (text):

|--- overcorrect_relative_weight <= 1.00
|   |--- steer_bad_behavior_weight <= 0.10
|   |   |--- overcorrect_relative_weight <= 0.03
|   |   |   |--- class: True
|   |   |--- overcorrect_relative_weight >  0.03
|   |   |   |--- class: False
|   |--- steer_bad_behavior_weight >  0.10
|   |   |--- end_layer_index <= 35.50
|   |   |   |--- class: False
|   |   |--- end_layer_index >  35.50
|   |   |   |--- class: True
|--- overcorrect_relative_weight >  1.00
|   |--- class: True

dependence-gemma

## Qwen tree

qwen-rule

|--- overcorrect_relative_weight <= 1.00
|   |--- steer_bad_behavior_weight <= 0.04
|   |   |--- start_layer_index <= 6.50
|   |   |   |--- class: False
|   |   |--- start_layer_index >  6.50
|   |   |   |--- class: False
|   |--- steer_bad_behavior_weight >  0.04
|   |   |--- end_layer_index <= 21.50
|   |   |   |--- class: False
|   |   |--- end_layer_index >  21.50
|   |   |   |--- class: True
|--- overcorrect_relative_weight >  1.00
|   |--- steer_bad_behavior_weight <= 0.00
|   |   |--- overcorrect_relative_weight <= 1.13
|   |   |   |--- class: False
|   |   |--- overcorrect_relative_weight >  1.13
|   |   |   |--- class: True
|   |--- steer_bad_behavior_weight >  0.00
|   |   |--- class: True

dependence-qwen

p-e-w:
Thanks, very interesting! Using a decision tree is a really good idea for analyzing this.

Am I reading those trees right, and explosion can occur even when `overcorrect_relative_weight` is less than 1? I'm not sure I've ever seen this happen.

What exactly is the definition of the "explosion" class?

kabachuha:
Currently explosion class is the kl_divergence > 1.5.

Extreme values are usually 10-20+, but I included these examples as they usually break the model too. (when KL > 1.0)

If I set up KL > 9.0, then for Qwen it doesn't happen than overcorrect is < 1. For Gemma however it a bit happens for some reason. (12 samples)

## Gemma tree KL 9+

gemma-rule-9plus

## Qwen tree KL 9+

qwen-rule-9pluspng

p-e-w:
KLD > 9 is definitely the correct criterion and KLD > 1.5 is not, because the latter also happens with traditional abliteration while the former doesn't. We need to concentrate on the unique conditions that only arise with ARA.

> For Gemma however it a bit happens for some reason. (12 samples)

That doesn't follow from the decision tree though. The root node has the criterion `overcorrect_relative_weight <= 1.002`, so the samples from the left branch that were classified as "explosion" could all have `overcorrect_relative_weight > 1`. Did you check the individual samples to verify that there were actually samples with `overcorrect_relative_weight < 1` but KLD > 9?

kabachuha:
I put the decision tree here as it was generated.

Btw, can you quickly launch this Pull-request on your end? To see how ARA-LoRA works on your hardware and what results you are getting

p-e-w:
I'm not on a GPU system right now, but I'll try it out in the next couple of days.

UnstableLlama:
I've been using this a lot, it has been very effective. I have also noticed the sensitivity with steer_bad_behavior_weight and overcorrect_relative_weight. Some experimental runs with a max steer_bad_behavior_weight set to 0.001 seem to have been pretty successful. Thinking of doing some runs with overcorrect_relative_weight set to a 0.8 max.

Roman215:
Expanding on comments from #365: Not suggesting anything from this should be used as I'm not really all that familiar with this code to begin with. When I was trying various different things to reduce VRAM usage so I could get better performance with a 31B model on my RTX 3090, I made the following changes:

image

image

kabachuha:
Just like I thought, llmfan started to make his abliterated models paid on Huggingface

This is just why we need this pull request merged ASAP!

https://old.reddit.com/r/LocalLLM/comments/1uae1jq/paidgated_model_minimaxm3_heretic_uncensored/

p-e-w:
I see nothing wrong with someone who put in the effort and compute needed to abliterate a giant model asking people to pay for access, and it most certainly won't influence the Heretic development timeline in any way.

Currently, ARA is still suffering from two major issues:

1. Exported models sometimes contain corrupted tensors with `NaN` values. This is a showstopper that **must** be fixed before ARA is merged.

2. Convergence behavior is erratic, leading to the mysterious trials with KLD > 10. This at the very least must be *understood*, to ensure it doesn't reflect a deeper problem.

There is no "ASAP" in this project. Things get merged if and when they are fully ready. There is no shortage of vibecoded junk projects whose authors don't understand their own code, and Heretic will not become one of them. Currently, my focus is automated testing (#392) in preparation for the plugin system (#53) and ARA. I get that everyone is mostly interested in features, but it's QA work that makes those features possible in the first place.

AesSedai:
image

Since the heretic'd Minimax-M3 was mentioned, I'm working on an abliteration of it using ARA with this PR as well, plus a fair amount of additional tweaks on top, some of which could maybe be upstreamed once the plugin system is added. I was intending to do this abliteration regardless of whether someone else did it, but having another model to compare against made it more of a priority to me.

The biggest changes are:
- using vLLM with dynamic LoRA loading, M3 in pytorch was running ~2-5tk/s even at 32 batch size; vLLM even with `punica` wrapping the model gets around 200tk/s so much faster trials.
- better detection of broken trials (checking if the first 8 results are all blank, unprintable, null-bytes, or too compressible because it emitted repetitive token-salad).
- replacing the marker-based refusal detection with an llm-as-a-judge prompt (using the base model, ironically enough) for detection non-compliance.
- added thinking + non-thinking template support
- added image-input support
- doing the LBFGS via fan-out to all available GPUs so it computes much more quickly

so I've got a fair few adjustments on top, but it still takes about ten minutes for a single trial. A few of those changes might be worth making PR's for, and I'm really curious to see how my resulting abliteration benches against the other one :)

kabachuha:
@AesSedai very happy for you!

vLLM is also that I suggested in #192, and for bigger models it's probably a must, but the installation is pain for a regular user (kernels, cuda-toolkit, attention, lots of dependencies and stuff), so this is really up to p-e-w

LLM as a Judge kills (or very much complicates) reproducibility and adds overhead. (every refusal must be additionally processed by a neutral or a remote LLM instead of string matching) For most of the systems, the marker based is okay for me, especially as it is shown to generalize (except for the cases like "to make illegal drugs,,,", where it is confusing the rating system), because the refusal mechanism is independent of the topic.

kabachuha:
From Huggingface, I saw you have 8 × RTX PRO 6000 Max-Q. Any chance for a Kimi? 😬
(They boast the jailbreak resistance in their technical reports and also Kimi crude abliterations attempts failed at #221)

AesSedai:
I don't disagree about LLM judging causing inconsistency, but I was running into issues with false positives since some of the prompts were RP-related and the generated text might include "sorry" or the like. And I noticed false negatives on some soft refusals as well, enough that I brought the judge in since I was able to test it across a few thousand prompts and it's fairly accurate.

RE: Kimi, that's doable but I'm more interested in MiMo-V2.5-Pro as a target since I use it (and M3) as my preferred creative writing models. Kimi was my prior favorite, but it's a little too "mad poet" compared to MiMo / M3 which are still highly creative and variable, just a little more consistent. So Kimi isn't off the table, when they have a non-Code 2.7 release that's worth looking into as a target IMO.

kabachuha:
The reason why Kimi is of particular interest is that the Kimi's team points out how extensively and exhaustively they trained it to refuse harmful requests and jailbreaks. It is also very sparse and the refusal location - and whether it is "spread" over the model and not a single vulnerability - would be very nice to examine. See the issue I linked, where the traditional directional ablation has seemingly failed. The original Huggingface repository has been deleted by the authors, luckily, I copied the full text into the issue in time. At this time, I'd say Kimi is "the final boss" of Heretic, and testing on it will prove whether Heretic/ARA can apply to everything in its current form or there are currently "unsolvable" cases. Of course, it's your resources, so you to decide ultimately :)

p-e-w:
> but I was running into issues with false positives

False positives aren't actually a problem even if they occur. Measuring compliance is always a heuristic and LLMs can also interpret responses in a different way from what humans would actually want. Any model with a refusal count below 10-15 can be assumed to be fully unrestricted.

The main problem with all of these approaches is that they're fine when you're hacking around on a local fork, but a mammoth task to ship in a system like Heretic that's designed to work out of the box for all/most users. Getting something like that ready to merge can easily be 10x the amount of work it was to hack the original implementation together. Classic difference between something that works in the lab and something that's supposed to be sold in a store.

p-e-w:
vLLM is something that I periodically experiment with. The introspection features aren't quite on par with Transformers yet, but switching at some point in the future isn't out of the question.

p-e-w:
> better detection of broken trials (checking if the first 8 results are all blank, unprintable, null-bytes, or too compressible because it emitted repetitive token-salad).

This sounds good, and classifying those as refusals (or, more accurately, as undesirable) sounds like a great improvement. I'd be happy to merge a PR that implements these changes by themselves.

src/heretic/config.py
```diff
@@ -203,6 +203,18 @@ class Settings(BaseSettings):
...
+    
+    use_ara_lora: bool = Field(
+        default=False,
+        description=(
+            "Use LoRA in ARA instead of full-weight editing. Makes it compatible with quantization and removes model reloads."
+        ),
+    )
+    
+    ara_lora_rank: int = Field(
+        default=128,
```
p-e-w, line 215: What happens if the rank is larger than the dimension of the matrix, which might be the case for some toy models?

src/heretic/model.py
```diff
@@ -161,7 +161,7 @@ def __init__(self, settings: Settings):
         if self.model is None:
             raise Exception("Failed to load model with all configured dtypes.")
 
-        if not settings.use_ara:
+        if not settings.use_ara or settings.use_ara_lora:
```
p-e-w, line 164: This logic implies that both `use_ara` and `use_ara_lora` could be set simultaneously, which I don't think should be allowed.

src/heretic/model.py
```diff
@@ -188,21 +188,24 @@ def _apply_lora(self):
         # because hybrid models like Qwen3.5 MoE have modules with different names
         # across layers (e.g. "o_proj" on attention layers, "out_proj" on linear attention layers).
         target_modules_set: set[str] = set()
+        
+        module_id_to_full_name = {
+            id(module): module_name
+            for module_name, module in self.model.named_modules()
+        }
```
p-e-w, line 195: What is the purpose of these changes?

src/heretic/model.py
```diff
@@ -668,6 +674,118 @@ def closure() -> Tensor:
...
+
+                    # Data preparation.
+                    # Move I/O tensors to the device of the adapter weights.
+                    good_input, good_output = good_module_io[layer_index][component][module_index]
+                    bad_input, bad_output = bad_module_io[layer_index][component][module_index]
+
+                    good_input = good_input.float().to(lora_A.device)
+                    good_output = good_output.float().to(lora_A.device)
+                    bad_input = bad_input.float().to(lora_A.device)
+                    bad_output = bad_output.float().to(lora_A.device)
```
p-e-w, line 727: Why do you upcast to `float32` here?

src/heretic/model.py
```diff
@@ -668,6 +674,118 @@ def closure() -> Tensor:
...
+                    def closure():
+                        optimizer.zero_grad()
+                        # Pass the actual tensors being optimized to the objective.
+                        loss = objective(lora_A, lora_B)
+                        loss.backward()
+                        return loss
+
+                    # Run optimization steps.
+                    for step in range(5):
+                        optimizer.step(closure)
```
p-e-w, line 787: What behavior do you observe here? Does it converge reliably?

p-e-w (commented):
Ok, let's move forward with this PR to reduce the number of moving parts!

src/heretic/config.py
kabachuha: In these toy cases and for people who think more is more, I can add a (L_rank <= M_rank) check, of course

src/heretic/model.py
kabachuha: Gemma 4 compatibility per https://github.com/p-e-w/heretic/commit/e2c74bfb3ca41cbb3123565867776221150a41e8

src/heretic/model.py
kabachuha: Use ara lora overrides use ara, so they can be turned on at the same time. Do you propose that use ara and use ara lora are exclusive distinct modes?

src/heretic/model.py
kabachuha: Just better precision, in theory. We drop a lot of precision already, so I agree, I can try stripping it here to lower memory usage

src/heretic/model.py
kabachuha: The number of steps (5) is copied from vanilla ARA. If you'd like, I can do some loss text outputs/plots at this place like in ARA's matching commented-out code segment.

p-e-w:
Ok, I merged this into the main ARA branch to simplify things as none of the above concerns is serious. Let's continue ARA development in #211.

PR #334, opened 2026-05-28
feat: add "disclaimer" to the prohibited strings list

kabachuha:
The favorite Gemma's word.

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -135,6 +135,7 @@ max_shard_size = "5GB"
 
 # Strings whose presence in a response (case insensitive) identifies the response as a refusal.
 refusal_markers = [
+    "disclaimer",
     "sorry",
     "i can'",
     "i cant",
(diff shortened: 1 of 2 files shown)


p-e-w:
Doesn't a disclaimer often precede a compliant response?

Example:

> DISCLAIMER: The following is for educational purposes only!
>
> *[full response with requested information follows]*

I have certainly seen responses like that in the past.

kabachuha:
> The following is for educational purposes only!

Is such a desired response or it will hinder the appearance? It does comply, but it's often visible that it struggles with it (e.g. I am an AI, not a doctor.) Removing this can diminish the "uncertain" range of refusals and make it more streamlined, as well as to help the user experience.

kabachuha:
Like, "Disclaimer" has a perfect "nanny"-style feel

p-e-w:
Ok, I agree. While the response may still be technically compliant, the disclaimer itself is undesirable.

Merged!

PR #337, opened 2026-05-30
feat: support plain text files as prompt datasets

### LINKED ISSUE #98: add option to custom datasets via txt file

rocker-zhang:
Closes #98.

Revives #103 (closed for author inactivity) on top of the current `load_prompts`, and resolves the review feedback that was still open there.

A `dataset` path pointing to a plain file is now read as one prompt per line, with empty lines ignored. For text files `column` is ignored and `split` is optional; when given, it selects a subset of lines with the usual slice notation, e.g. `[:400]`.

What changed relative to #103's last state:
- text files are detected with `os.path.isfile`, so files without an extension work
- split parsing is factored into one helper (`get_split_slice`) instead of duplicated
- `split` and `column` are now optional in `DatasetSpecification`; the dataset branches raise a clear error when either is missing
- an invalid split now raises instead of being silently ignored

A bare slice such as `[:400]` does not parse with the pinned `datasets` version, because `ReadInstruction.from_spec` expects a named split. The text branch prepends a synthetic split name so the documented notation works. I checked `[:400]`, `[1:3]`, `[:50%]`, negative indices, and that malformed slices raise.

No unit tests exist in the repo; validated with `ruff format`, `ruff check`, `ty`, and a local run of `load_prompts` over a text file covering the cases above.

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -188,6 +188,20 @@ class Prompt:
     user: str
 
 
+def get_split_slice(split_str: str, length: int) -> tuple[int, int]:
+    """Resolves a split specification into absolute (start, end) indices."""
+
+    # The split name is the part before the slice, e.g. "train" in "train[:400]".
+    split_name = split_str.split("[")[0]
+    # Associate the split with its number of examples (lines).
+    name_to_length = {split_name: length}
+    # Convert the instructions to absolute indices and select the first one.
+    absolute_instruction = ReadInstruction.from_spec(split_str).to_absolute(
+        name_to_length
+    )[0]
+    return absolute_instruction.from_, absolute_instruction.to
+
+
 def load_prompts(
     settings: Settings,
     specification: DatasetSpecification,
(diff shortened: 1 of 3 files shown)


rocker-zhang:
Renamed `name2len` and `abs_instruction` to `name_to_length` and `absolute_instruction` per the style guide.

I left out the `FileNotFoundError` suggestion: the proposed version wraps `is_hf_path` in a broad `try/except`, which is the pattern that was removed in #103. If a clearer error for a missing text-file path is useful, I can add a narrow check that does not swallow exceptions.

p-e-w:
Thanks, I like this! However, since you based this on #103, commit credit to @ricyoung is required. Ideally, start with the commits from #103, then apply your changes as additional commits, but at the absolute minimum, you should add @ricyoung as a co-author on the first commit.

src/heretic/utils.py
```diff
@@ -188,36 +188,60 @@ class Prompt:
     user: str
 
 
+def get_split_slice(split_str: str, length: int, split_name: str) -> tuple[int, int]:
```
p-e-w: Can't we get `split_name` automatically by splitting `split_str` before the `[`?

p-e-w (approved):
Implementation looks flawless. Not something I've seen very often on the first try!

rocker-zhang:
> Thanks, I like this! However, since you based this on #103, commit credit to @ricyoung is required. Ideally, start with the commits from #103, then apply your changes as additional commits, but at the absolute minimum, you should add @ricyoung as a co-author on the first commit.

Thanks! Added @ricyoung as co-author on the commit. The change is based on the current load_prompts, which diverged from #103 after the is_hf_path refactor (#325), so a clean replay of the original commits wasn't possible, but the credit is in place

src/heretic/utils.py
rocker-zhang: > Can't we get `split_name` automatically by splitting `split_str` before the `[`?

Good call on split_name: get_split_slice now derives it from the spec (split_str.split("[")[0]), so the extra parameter is gone. The text-file branch still prepends a synthetic name for the same reason

p-e-w:
Yup, that does the trick. Merged, thanks! It's good to have this.

PR #338, opened 2026-05-31
build(deps): remove unused hf-transfer dependency

rocker-zhang:
`hf-transfer` is declared in `pyproject.toml` but never used: nothing in the codebase sets `HF_HUB_ENABLE_HF_TRANSFER`, and downloads go through `from_pretrained` / `hf_hub_download` with no transfer toggle. With `huggingface-hub ~= 1.7`, Xet (`hf-xet`, already pulled in transitively) is the default accelerated backend and `HF_HUB_ENABLE_HF_TRANSFER` is deprecated, so the dependency is dead weight that only surfaces a deprecation warning.

Validated with `uv lock`, `uv sync`, `uv run heretic --help`, `ruff check`, and `ty check`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ dependencies = [
     "accelerate~=1.13",
     "bitsandbytes~=0.49",
     "datasets~=4.7",
-    "hf-transfer~=0.1",
     "huggingface-hub~=1.7",
     "immutabledict~=4.3",
     "kernels~=0.13",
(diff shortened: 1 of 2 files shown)


p-e-w:
I had added `hf-transfer` as a dependency several months ago after Xet wasn't being used on a rented box for some reason, and Transformers raised an exception because no alternative was available.

However, that was before `huggingface-hub` 1.0 was released, and while I don't remember the machine configuration and thus can't replicate the original problem, it's reasonable to assume that it was resolved in the meantime, so I'm merging this PR.

PR #344, opened 2026-06-02
feat: add support for LiquidAI/LFM2.5 models

coder3101:
Models available at: https://huggingface.co/collections/coder3101/lfm-25

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -389,20 +389,20 @@ class Model:
             for expert in layer.block_sparse_moe.experts:  # ty:ignore[possibly-missing-attribute, not-iterable]
                 try_add("mlp.down_proj", expert.w2)  # ty:ignore[possibly-missing-attribute]
 
-        # LFM dense operator blocks
+        # LFM dense operator blocks.
         with suppress(Exception):
-            try_add("attn.o_proj", layer.conv.out_proj)
+            try_add("attn.o_proj", layer.conv.out_proj)  # ty:ignore[possibly-missing-attribute]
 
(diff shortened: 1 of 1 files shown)


src/heretic/model.py
```diff
@@ -389,6 +389,21 @@ def try_add(component: str, module: Any):
             for expert in layer.block_sparse_moe.experts:  # ty:ignore[possibly-missing-attribute, not-iterable]
                 try_add("mlp.down_proj", expert.w2)  # ty:ignore[possibly-missing-attribute]
 
+        # LFM dense operator blocks.
+        with suppress(Exception):
+            try_add("attn.o_proj", layer.conv.out_proj)  # ty:ignore[possibly-missing-attribute]
```
p-e-w, line 394: We might revisit identifying convolutions with attention in the future, especially if SSMs become more widespread. Having separate parameter sets could improve results if it turns out the two component types react very differently to ablation, but for now this is fine.

src/heretic/model.py
```diff
@@ -389,6 +389,21 @@ def try_add(component: str, module: Any):
...
+
+        with suppress(Exception):
+            try_add("mlp.down_proj", layer.feed_forward.w2)   # ty:ignore[possibly-missing-attribute]
```
p-e-w: This identification is canonical, and matches Phi-MoE.

p-e-w (approved):
Please fix CI, then this is going in.

p-e-w:
Merged, thanks! It's good to have this.

PR #348, opened 2026-06-04
fix: prevent NaN KL divergence on models with suppressed tokens

### LINKED ISSUE #346: NaN KL divergence on Gemma-4 12b-it
### Model

```bash
heretic --model google/gemma-4-12B-it --batch-size 128
```

```bash
Running trial 5 of 200...
* Parameters:
  * direction_index = per layer
  * attn.o_proj.max_weight = 1.21
  * attn.o_proj.max_weight_position = 30.99
  * attn.o_proj.min_weight = 0.33
  * attn.o_proj.min_weight_distance = 6.25
  * mlp.down_proj.max_weight = 1.32
  * mlp.down_proj.max_weight_position = 34.42
  * mlp.down_proj.min_weight = 1.18
  * mlp.down_proj.min_weight_distance = 22.99
* Resetting model...
* Abliterating...
* Evaluating...
  * KL divergence: nan
  * Counting model refusals...
  * Refusals: 20/100
```

> [!TIP]
> the NEW gemma-4 12b does **not** use `Gemma4ForConditionalGeneration` which is used by E2B-it and E4B-it variants.
>
> It uses `Gemma4UnifiedForConditionalGeneration` as of transformers version `v5.10.1` latest

This one as well does not seem to be a Heretic issue, but `transformers` (modeling code / config.json might be incorrect)

igorls:
## Problem

Running Heretic on `google/gemma-4-12B-it` yields `KL divergence: nan` on every trial, while refusal counts remain sane (reported in #346):

```
* Evaluating...
  * KL divergence: nan
  * Counting model refusals...
  * Refusals: 20/100
```

## Root cause

`google/gemma-4-12B-it` is a multimodal model whose `generation_config.json` declares:

```json
"suppress_tokens": [258883, 258882]   // end-of-audio / end-of-image
```

During `get_logprobs`, `generate(..., output_scores=True)` runs `SuppressTokensLogitsProcessor`, which sets those two tokens' logits to `-inf` in **both** the base and abliterated first-token logprob tensors (verified: 0 NaN, 0 +inf, exactly 2 `-inf` per row).

`get_score` then computes:

```python
F.kl_div(logprobs, base_logprobs, reduction="batchmean", log_target=True)
```

With `log_target=True` the per-element term is `target.exp() * (target - input)`. At a suppressed token both tensors are `-inf`, so the term is `exp(-inf) * (-inf - -inf)` = `0 * nan` = **`nan`**, which poisons the whole sum.

This only affects KL divergence, not refusal counting (which argmaxes generated tokens and never touches the `-inf` scores) — exactly the reported symptom. Models without `suppress_tokens`/`begin_suppress_tokens` were unaffected, which is why it surfaced only now with multimodal Gemma-4.

## Fix

Compute the KL divergence directly and zero out contributions from tokens that carry zero probability mass under the base distribution, per the standard convention `0 * log(0/q) = 0`. For models without suppressed tokens the result is identical to the previous `F.kl_div` call.

Verified numerically:

| case | `F.kl_div` | this fix |
| :--- | ---: | ---: |
| fully finite distributions | 1.04459 | 1.04459 |
| shared suppressed tokens (`-inf`) | `nan` | 0.95617 |
| identical + suppressed | `nan` | 0.0 |

End-to-end, a full run on `google/gemma-4-12B-it` now reports finite KL divergence on every trial.

Fixes #346

### DIFF
--- a/src/heretic/evaluator.py
+++ b/src/heretic/evaluator.py
@@ -14,6 +14,8 @@ class Evaluator:
     good_prompts: list[Prompt]
     bad_prompts: list[Prompt]
     base_logprobs: Tensor
+    base_probs: Tensor
+    base_isneginf: Tensor
     base_refusals: int
 
     def __init__(self, settings: Settings, model: Model):
@@ -29,6 +31,11 @@ class Evaluator:
 
         print("* Obtaining first-token probability distributions...")
         self.base_logprobs = model.get_logprobs_batched(self.good_prompts)
+
+        # The base distribution is constant across all trials, so precompute the
+        # quantities that kl_divergence() needs from it once, rather than on every call.
+        self.base_probs = self.base_logprobs.exp()
+        self.base_isneginf = self.base_logprobs.isneginf()
 
         print()
         print(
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks for identifying the root cause, I wasn't aware of the `suppress_tokens` mechanism.

I'm not too happy with the fix though. We really don't want to implement computing the KL divergence ourselves. The upstream implementation has backend-specific code and we want the exact same numerical results as `torch.nn.functional.kl_div` for distributions that don't have `-inf` elements, which is almost all of them.

This is an ugly problem to fix, but I guess one way could be to instead edit the logprobs tensors to remove elements that are `-inf` in both tensors, then call PyTorch's `kl_div` on the resulting tensors. The basic idea has to be that all logic remains exactly the same for the vast majority of models that don't use token suppression, and those that do get a reasonable-effort workaround.

igorls:
Superseded by #350, which is the cleaner fix. Instead of post-processing the KL divergence, it switches `get_logprobs` from `output_scores=True` to `output_logits=True`, i.e. it reads the model's raw unprocessed logits instead of the generation-processed scores. The `-inf` values come from `SuppressTokensLogitsProcessor` (applied only to the generation scores), so the raw logits never contain them and `F.kl_div` can be used completely unchanged — which directly addresses your objection here about not reimplementing the KL computation.

I verified on `google/gemma-4-12B-it` that for the first greedy token the only processor active is `suppress_tokens`, so `output_logits` and `output_scores` are **element-for-element identical except at the suppressed token positions**. That means #350 produces byte-identical results to the current code for every model that doesn't suppress tokens, and the correct raw distribution for those that do.

Closing in favor of #350. Thanks @MoonRide303.

PR #350, opened 2026-06-04
feat: add support for gemma-4-12B-it

MoonRide303:
Use the model's raw unprocessed logits (output_logits) instead of the processed generation scores when computing KL divergence. Fixes KL divergence: nan on google/gemma-4-12B-it.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -747,7 +747,7 @@ class Model:
         _, outputs = self.generate(
             prompts,
             max_new_tokens=1,
-            output_scores=True,
+            output_logits=True,
             return_dict_in_generate=True,
             use_cache=False,
         )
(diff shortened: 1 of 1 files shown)


p-e-w:
Thanks!

I'm trying to remember why I used scores instead of logits in the first place. I have a vague memory of an incompatibility, but it could be just a hallucination. Otherwise, this solution seems obviously correct.

I do wonder though: What *are* the logits that `generate` returns for those tokens?

MoonRide303:
> I do wonder though: What _are_ the logits that `generate` returns for those tokens?

Just raw model logits (before `SuppressTokensLogitsProcessor`).

igorls:
This is the right fix — I came at #346 from the KL-computation side (#348) but `output_logits` addresses the root cause and keeps `F.kl_div` untouched, which is what the maintainer wanted.

Some verification on `google/gemma-4-12B-it` that should ease any regression concern: for the first greedy token the only active logits processor is `suppress_tokens`, so `outputs.logits[0]` and `outputs.scores[0]` are **identical except at the 2 suppressed token positions** (EOI/EOA). Concretely, on a 2-prompt batch:

```
output_scores: finite=False  num -inf=4   ->  self-KL = nan
output_logits: finite=True   num -inf=0   ->  self-KL = 0.0
elements where scores != logits: 4  (= 2 suppressed tokens x 2 rows)
```

So for every model that doesn't suppress tokens, `output_logits == output_scores` and KL values are unchanged; suppressed models get the correct raw distribution instead of a NaN. `get_residuals` correctly needs no change since it reads `hidden_states`, which processors never touch. I've closed #348 in favor of this.

p-e-w:
Indeed, this is the correct fix. Thank you, merged! And thank you also to @igorls for the valuable information!

PR #352, opened 2026-06-04
fix: fix issues in automatic reproduction system

Vinay-Umrethe:
### Not ready for merge yet

This PR would likely implement few fixes or rest of the checks like `requirements.txt` installation, SHA256 checksum

> [!NOTE]
> Not doing auto install requirements as it would change packages at system level (reverted), might implement it with venv setup at start

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -133,6 +133,9 @@ study_checkpoint_dir = "checkpoints"
 # Maximum size for individual safetensors files generated when exporting a model.
 max_shard_size = "5GB"
 
+# How to export the model: "none" (prompt the user), "merge", or "adapter".
+export_strategy = "none"
+
 # Strings whose presence in a response (case insensitive) identifies the response as a refusal.
 refusal_markers = [
     "sorry",
--- a/config.default.toml
+++ b/config.default.toml
@@ -123,10 +123,6 @@ n_trials = 200
 # Number of trials that use random sampling for the purpose of exploration.
 n_startup_trials = 60
 
-# Random seed for reproducible optimization. Set to an integer to enable.
-# Applies to Python's random module, NumPy, PyTorch, and Optuna.
-# seed = 75
-
 # Directory to save and load study progress to/from.
 study_checkpoint_dir = "checkpoints"
 
--- a/config.default.toml
+++ b/config.default.toml
@@ -133,9 +133,6 @@ study_checkpoint_dir = "checkpoints"
 # Maximum size for individual safetensors files generated when exporting a model.
 max_shard_size = "5GB"
 
-# How to export the model: "none" (prompt the user), "merge", or "adapter".
-export_strategy = "none"
-
 # Strings whose presence in a response (case insensitive) identifies the response as a refusal.
 refusal_markers = [
     "sorry",
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -32,6 +32,12 @@ class RowNormalization(str, Enum):
     FULL = "full"
 
 
+class ExportStrategy(str, Enum):
+    NONE = "none"
+    MERGE = "merge"
+    ADAPTER = "adapter"
+
+
 class DatasetSpecification(BaseModel):
     dataset: str = Field(
         description="Hugging Face dataset ID, or path to dataset on disk."
@@ -410,6 +416,11 @@ class Settings(BaseSettings):
     max_shard_size: int | str = Field(
         default="5GB",
         description="Maximum size for individual safetensors files generated when exporting a model.",
+    )
+
+    export_strategy: ExportStrategy | None = Field(
+        default=None,
+        description='How to export the model: "none" (prompt the user), "merge", or "adapter".',
     )
 
     refusal_markers: list[str] = Field(
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -419,7 +419,7 @@ class Settings(BaseSettings):
 
     export_strategy: ExportStrategy | None = Field(
         default=None,
-        description='How to export the model: "none" (prompt the user), "merge", or "adapter".',
+        description='How to export the model: "merge", "adapter", or unset to prompt the user.',
     )
 
     refusal_markers: list[str] = Field(
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -33,7 +33,6 @@ class RowNormalization(str, Enum):
 
 
 class ExportStrategy(str, Enum):
-    NONE = "none"
     MERGE = "merge"
     ADAPTER = "adapter"
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -79,11 +79,9 @@ if not _is_help_invocation():
         _BOOTSTRAP_REPRODUCTION_INFORMATION = _load_reproduction_information_bootstrap(
             reproduce_path
         )
-        requirements = _BOOTSTRAP_REPRODUCTION_INFORMATION.get("environment", {}).get(
-            "requirements"
+        _install_reproduction_requirements(
+            _BOOTSTRAP_REPRODUCTION_INFORMATION["environment"]["requirements"]
         )
-        if requirements:
-            _install_reproduction_requirements(requirements)
 
 from .config import Settings
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -88,11 +88,11 @@ from .utils import (
 )
 
 
-def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
+def obtain_export_strategy(settings: Settings, model: Model) -> str | None:
     """
     Prompts the user for how to proceed with saving the model.
     Provides info to the user if the model is quantized on memory use.
-    Returns "merge", "adapter", or None (if cancelled/invalid).
+    Returns "merge", "adapter", or None if cancelled.
     """
 
     if settings.quantization == QuantizationMethod.BNB_4BIT:
@@ -868,7 +868,7 @@ def run():
                             if not save_directory:
                                 continue
 
-                            strategy = obtain_merge_strategy(settings, model)
+                            strategy = obtain_export_strategy(settings, model)
                             if strategy is None:
                                 continue
 
@@ -926,7 +926,7 @@ def run():
                                 continue
                             private = visibility == "Private"
 
-                            strategy = obtain_merge_strategy(settings, model)
+                            strategy = obtain_export_strategy(settings, model)
                             if strategy is None:
                                 continue
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -88,12 +88,19 @@ from .utils import (
 )
 
 
-def obtain_export_strategy(settings: Settings, model: Model) -> ExportStrategy | None:
+def obtain_export_strategy(
+    settings: Settings,
+    model: Model,
+    original_export_strategy: ExportStrategy | None = None,
+) -> ExportStrategy | None:
     """
-    Prompts the user for how to proceed with saving the model.
+    Gets the export strategy from settings or prompts the user.
     Provides info to the user if the model is quantized on memory use.
     Returns an export strategy, or None if cancelled.
     """
+
+    if settings.export_strategy is not None:
+        return settings.export_strategy
 
     if settings.quantization == QuantizationMethod.BNB_4BIT:
         print()
@@ -138,8 +145,11 @@ def obtain_export_strategy(settings: Settings, model: Model) -> ExportStrategy |
             )
         print()
 
-    if settings.export_strategy is not None:
-        return settings.export_strategy
+    if original_export_strategy is not None:
+        print(
+            f"Original export strategy was [bold]{original_export_strategy.value}[/]; "
+            "choose it for proper hash verification."
+        )
 
     strategy = prompt_select(
         "How do you want to proceed?",
@@ -221,6 +231,7 @@ def run():
         return
 
     reproduction_mode = settings.reproduce is not None
+    original_export_strategy: ExportStrategy | None = None
 
     if settings.reproduce is not None:
         print(f"Loading reproduction information from [bold]{settings.reproduce}[/]...")
@@ -242,6 +253,8 @@ def run():
         print()
 
         settings = Settings.model_validate(reproduction_information["settings"])
+        original_export_strategy = settings.export_strategy
+        settings.export_strategy = None
 
     if settings.seed is None:
         settings.seed = random.randint(0, 2**32 - 1)
@@ -868,7 +881,11 @@ def run():
                             if not save_directory:
                                 continue
 
-                            strategy = obtain_export_strategy(settings, model)
+                            strategy = obtain_export_strategy(
+                                settings,
+                                model,
+                                original_export_strategy,
+                            )
                             if strategy is None:
                                 continue
 
@@ -926,7 +943,11 @@ def run():
                                 continue
                             private = visibility == "Private"
 
-                            strategy = obtain_export_strategy(settings, model)
+                            strategy = obtain_export_strategy(
+                                settings,
+                                model,
+                                original_export_strategy,
+                            )
                             if strategy is None:
                                 continue
 
@@ -1039,6 +1060,7 @@ def run():
                                 # Set the number of trials to the number of actual completed trials
                                 # for the reproduction configuration.
                                 settings.n_trials = count_completed_trials()
+                                current_export_strategy = settings.export_strategy
                                 settings.export_strategy = strategy
 
                                 upload_reproduce_folder(
@@ -1051,6 +1073,7 @@ def run():
                                         reproducibility_information == "full"
                                     ),
                                 )
+                                settings.export_strategy = current_export_strategy
 
                             print(f"Model uploaded to [bold]{repo_id}[/].")
 
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -79,9 +79,11 @@ if not _is_help_invocation():
         _BOOTSTRAP_REPRODUCTION_INFORMATION = _load_reproduction_information_bootstrap(
             reproduce_path
         )
-        _install_reproduction_requirements(
-            _BOOTSTRAP_REPRODUCTION_INFORMATION["environment"]["requirements"]
+        requirements = _BOOTSTRAP_REPRODUCTION_INFORMATION.get("environment", {}).get(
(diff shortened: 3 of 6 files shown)


src/heretic/reproduce.py
```diff
@@ -32,11 +36,14 @@ def collect_reproducibles(path: str):
     )
     print()
 
-    api = HfApi()
+    token = get_token()
+    token_arg = token or False
+    api = HfApi(token=token_arg)
```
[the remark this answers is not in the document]
Vinay-Umrethe: `arg` is "widely used and understood" for `argument`

src/heretic/reproduce.py
```diff
@@ -51,6 +58,14 @@ def collect_reproducibles(path: str):
             if model.tags is not None and "gguf" in model.tags:
                 continue
 
+            if model.gated is not False:
+                if not token:
+                    continue
+                try:
+                    api.auth_check(model.id, repo_type="model")
+                except GatedRepoError:
+                    continue
```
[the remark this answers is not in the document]
Vinay-Umrethe: False positive partially, 

`if model.gated:` is better and correct

`api.auth_check()` *does* exist

Vinay-Umrethe:
### Status

* [x] 1. Fixed: inaccessible gated models are now skipped during reproducibles collection.

* [x] 2. Fixed: `torch.svd_lowrank` is seeded *before it's use* so restoring a model from trial parameters is deterministic (independent of previous trials that required re-run sequentially them using `config.toml` manual reproduction)

* [x] 3. Add `merged OR adapter` in `reproduce.json` to know what was originally exported (version bump "2")

@p-e-w

src/heretic/reproduce.py
p-e-w: This happens by default, according to the docs:

https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api#huggingface_hub.HfApi

> `token` (bool or str, optional) — A valid user access token (string). **Defaults to the locally saved token** [...]

You don't have to do anything here.

src/heretic/reproduce.py
```diff
@@ -32,11 +36,14 @@ def collect_reproducibles(path: str):
...
 
-    api = HfApi()
+    token = get_token()
+    token_arg = token or False
+    api = HfApi(token=token_arg)
 
     models = api.list_models(
         filter=["heretic", "reproducible"],
         sort="created_at",
+        expand=["gated", "tags"],
```
p-e-w, line 44: Why is `tags` here?

src/heretic/reproduce.py
```diff
@@ -83,6 +98,7 @@ def collect_reproducibles(path: str):
             cache_path = hf_hub_download(
                 model.id,
                 "reproduce/reproduce.json",
+                token=token_arg,
```
p-e-w: Same here, you don't have to pass this, the library reads the stored token by default.

src/heretic/reproduce.py
Vinay-Umrethe: To avoid GGUFs

p-e-w: But that was already working before, when the `expand` argument wasn't there.

Vinay-Umrethe: because adding `expand=["gated"]`` changes the returned fields. without expanding tags, model.tags becomes `None,` so current filter won't work


```python
from huggingface_hub import HfApi

api = HfApi()

for expand in (["gated"], ["gated", "tags"]):
    model = next(
        iter(
            api.list_models(
                filter=["heretic", "reproducible"],
                sort="created_at",
                expand=expand,
                limit=1,
            )
        )
    )

    print(f"expand={expand}")
    print("id:", model.id)
    print("gated:", repr(model.gated))
    print("tags:", repr(model.tags))
    print()
```

```bash
expand=['gated']
id: mradermacher/Qwen3-0.6B-heretic-Base-GGUF
gated: False
tags: None

expand=['gated', 'tags']
id: mradermacher/Qwen3-0.6B-heretic-Base-GGUF
gated: False
tags: ['transformers', 'gguf', 'heretic', 'uncensored', 'decensored', 'abliterated', 'reproducible', 'en', 'base_model:VINAY-UMRETHE/Qwen3-0.6B-heretic-Base', 'base_model:quantized:VINAY-UMRETHE/Qwen3-0.6B-heretic-Base', 'license:apache-2.0', 'endpoints_compatible', 'region:us', 'conversational']
```

that's a with and without example

src/heretic/reproduce.py
Vinay-Umrethe: correct

src/heretic/reproduce.py
Vinay-Umrethe: right

Vinay-Umrethe:
### Test conducted

branch: `Vinay-Umrethe:auto-reproduce-implementations`

method: `--reproduce`

1.

Base model: https://huggingface.co/VINAY-UMRETHE/Qwen3-0.6B-heretic-Base2

Result: `Hash matches`

Type: Merged model export.

https://huggingface.co/VINAY-UMRETHE/Qwen3-0.6B-heretic-Reproduce2

---

2.

Base model (adapter LoRA Only): https://huggingface.co/VINAY-UMRETHE/Qwen3-0.6B-heretic-LoRA

Result: `Hash matches`

Type: Adapter only export

https://huggingface.co/VINAY-UMRETHE/Qwen3-0.6B-heretic-LoRA-Reproduce

```bash
Using original export strategy: adapter
Uploading LoRA adapter...
...
...
Model uploaded to VINAY-UMRETHE/Qwen3-0.6B-heretic-LoRA-Reproduce.
Verifying hashes of weight files...
adapter_model.safetensors: Hash matches
```

so now `torch.svd_lowrank` does not bother reproduction in `row_normalization`.

@p-e-w

src/heretic/main.py
```diff
@@ -88,7 +88,25 @@ def _is_help_invocation() -> bool:
...
+) -> str | None:
+    export = reproduction_information.get("export")
+    if not isinstance(export, dict):
+        return None
+
+    strategy = export.get("strategy")
+    if strategy in ["adapter", "merge"]:
+        return strategy
+
+    return None
```
p-e-w: This seems way overcomplicated. Why do all this verification that we don't do for other parts of the reproduction information? We just use the strategy that is stored in `reproduction_information`, and if none is stored, we prompt with `obtain_merge_strategy`.

src/heretic/main.py
```diff
@@ -138,6 +156,12 @@ def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
             )
         print()
 
+    # Reproduction must use the same export strategy (adapter / merged)
+    # this is required for hash verification to match the original model weights.
+    if original_strategy is not None:
+        print(f"Using original export strategy: [bold]{original_strategy}[/]")
```
p-e-w: No need to print such information, this goes without saying as the user has specifically chosen to reproduce.

src/heretic/reproduce.py
```diff
@@ -51,6 +56,12 @@ def collect_reproducibles(path: str):
             if model.tags is not None and "gguf" in model.tags:
                 continue
 
+            if model.gated:
+                try:
+                    api.auth_check(model.id, repo_type="model")
+                except GatedRepoError:
+                    continue
```
p-e-w, line 63: Yes. This is one of the very, very few cases where it's appropriate to swallow an exception completely.

src/heretic/reproduce.py
p-e-w: I see, thanks.

src/heretic/utils.py
```diff
@@ -570,6 +571,9 @@ def generate_reproduce_json(
             "base_refusals": trial.user_attrs["base_refusals"],
             "n_bad_prompts": trial.user_attrs["n_bad_prompts"],
         },
+        "export": {
```
p-e-w: Why store this as a nested object, instead of just `export_strategy`? All other aspects of export are expected to be in the settings.

p-e-w:
I'm wondering if we're approaching the export strategy thing correctly. Here's an alternative idea:

1. We store `export_strategy` in `Settings`, type `str | None`, default `None`.
2. If `None`, we prompt, as we do now, otherwise we use the given value.

This gets rid of the need to pass the value around, and allows users to simplify their export setup by fixing the value so they don't get prompted each time.

The current approach basically creates an ad-hoc setting, which I'm not sure is correct.

We can even use an enum that way, and get automatic validation.

Vinay-Umrethe:
it is simplified now, and yes that is more correct approach.

I still wonder what to do with version 1 of `reproduce.json` (as it does not seed the fix done in `model.py`)

Leave it dead OR if v1, use `"settings"` and run all trials sequentially instead of direct applying which is possible *now*

src/heretic/main.py
Vinay-Umrethe: done

src/heretic/main.py
Vinay-Umrethe: removed

src/heretic/utils.py
Vinay-Umrethe: stored in `settings` now

src/heretic/config.py
```diff
@@ -32,6 +32,12 @@ class RowNormalization(str, Enum):
     FULL = "full"
 
 
+class ExportStrategy(str, Enum):
+    NONE = "none"
```
p-e-w: No. "None" is not an export strategy. Rather, the `export_strategy` field can be an export strategy **or** the singleton `None`.

src/heretic/main.py
```diff
@@ -88,11 +88,11 @@ def _is_help_invocation() -> bool:
 )
 
 
-def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
+def obtain_export_strategy(settings: Settings, model: Model) -> str | None:
```
p-e-w: Adjust return type.

src/heretic/main.py
```diff
@@ -138,6 +138,9 @@ def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
             )
         print()
 
+    if settings.export_strategy not in [None, ExportStrategy.NONE]:
+        return settings.export_strategy.value
```
p-e-w: This logic doesn't make sense. We are warning about consequences above, but if the strategy is set in the settings, you're not giving the user a chance to react to those warnings.

config.default.toml
```diff
@@ -133,6 +133,9 @@ study_checkpoint_dir = "checkpoints"
 # Maximum size for individual safetensors files generated when exporting a model.
 max_shard_size = "5GB"
 
+# How to export the model: "none" (prompt the user), "merge", or "adapter".
+export_strategy = "none"
```
p-e-w: No, remove this setting from this file to get the default (which is `None`, not `"none"`).

src/heretic/main.py
```diff
@@ -1036,6 +1039,7 @@ def reset_trial_model() -> None:
                                 # Set the number of trials to the number of actual completed trials
                                 # for the reproduction configuration.
                                 settings.n_trials = count_completed_trials()
+                                settings.export_strategy = ExportStrategy(strategy)
```
p-e-w: Unfortunately, it's not so easy. You have now stored the strategy in settings, and that means the user cannot go back and export the model in a different format.

config.default.toml
Vinay-Umrethe: but will it work with a normal `None` or `null` in toml file?

src/heretic/main.py
Vinay-Umrethe: I did noticed it, but what else we have to do to ensure users have the choice anyway

config.default.toml
Vinay-Umrethe:
```python
class QuantizationMethod(str, Enum):
    NONE = "none"
    BNB_4BIT = "bnb_4bit"


class RowNormalization(str, Enum):
    NONE = "none"
```

```toml
# Quantization method to use when loading the model. Options:
# "none" (no quantization),
# "bnb_4bit" (4-bit quantization using bitsandbytes).
quantization = "none"
```

These two use the same

p-e-w: There is no `None` or `null` in TOML. This is a design limitation of TOML.

And the quantization case is not the same. Setting `quantization = "none"` means that we do not quantize. But setting `export_strategy = None` does **not** mean that we don't use an export strategy, but that we prompt the user for it. Those two are semantically different.

p-e-w: You can load the model without quantization, but you can't export the model without an export strategy. That's why "none" cannot be an export strategy.

src/heretic/main.py
p-e-w: I don't know the solution right now, but it can't remain like this. Not being able to change the export format is a major regression.

config.default.toml
Vinay-Umrethe: making it commented would be correct? like how we explain seed in .toml

p-e-w: I actually think that neither the seed nor the export strategy belong in this file even as a comment. Setting those in the configuration is almost always a bad choice for the user.

p-e-w: Note that there are several settings that are deliberately not in `config.default.toml`.

src/heretic/main.py
Vinay-Umrethe: if `--reproduce` would allow users to choose strategy, and they choose different one than original, the hash matching would fail which is misleading

Vinay-Umrethe: now it seems correct to me, no matter what was the original `export_strategy` we make it None anyway but store that info for *only* telling/print the User (to ensure hash verification is exactly correct)

But users still get the prompt and have the choice

src/heretic/config.py
```diff
@@ -412,6 +417,11 @@ class Settings(BaseSettings):
         description="Maximum size for individual safetensors files generated when exporting a model.",
     )
 
+    export_strategy: ExportStrategy | None = Field(
+        default=None,
+        description='How to export the model: "none" (prompt the user), "merge", or "adapter".',
```
p-e-w: "none" is still one of the documented values here.

config.default.toml
```diff
@@ -123,10 +123,6 @@ n_trials = 200
 # Number of trials that use random sampling for the purpose of exploration.
 n_startup_trials = 60
 
-# Random seed for reproducible optimization. Set to an integer to enable.
-# Applies to Python's random module, NumPy, PyTorch, and Optuna.
-# seed = 75
```
p-e-w, line 128: Yes, I think this is better. The biggest problem with having the seed in the config file is that it might be forgotten, and then lead to suboptimal results (identical runs) that are hard to debug. The `seed` parameter is much better passed as a command line argument, similar to `evaluate_model`, which also isn't in the config file.

src/heretic/main.py
```diff
@@ -138,6 +145,12 @@ def obtain_merge_strategy(settings: Settings, model: Model) -> str | None:
             )
         print()
 
+    if original_export_strategy is not None:
+        print(
+            f"Original export strategy was [bold]{original_export_strategy.value}[/]; "
+            "choose it for proper hash verification."
+        )
```
p-e-w: No, the user shouldn't get a choice at all when running with `--reproduce`. The whole point of reproducing a model is to get exactly  the same model, so we don't need to ask them, they already made that choice. Otherwise we get this weird situation where we can only verify checksums if the user made the "right" choice.

src/heretic/main.py
```diff
@@ -218,13 +231,14 @@ def run():
         return
 
     reproduction_mode = settings.reproduce is not None
+    original_export_strategy: ExportStrategy | None = None
```
p-e-w: We don't need this.

src/heretic/main.py
```diff
@@ -1047,6 +1073,7 @@ def reset_trial_model() -> None:
                                         reproducibility_information == "full"
                                     ),
                                 )
+                                settings.export_strategy = current_export_strategy
```
p-e-w: This is a nice, clean solution, but it has a flaw:

If an exception occurs in `upload_reproduce_folder`, the program will jump to the outer exception handler, then continue to ask the user what they want to do. But at that point, you have the export strategy hardcoded in the settings because this line wasn't reached, so you never get the choice anymore.

Use a `try...finally` block to fix this.

src/heretic/utils.py
```diff
@@ -547,7 +547,7 @@ def generate_reproduce_json(
     version_info = get_heretic_version_info()
 
     data = {
-        "version": "1",  # Version number of the reproduce.json file format, to allow for future changes.
+        "version": "2",  # Version number of the reproduce.json file format, to allow for future changes.
```
p-e-w, line 550: Why are we upgrading the version if the schema doesn't change?

src/heretic/main.py
Vinay-Umrethe: > You have now stored the strategy in settings, and that means the user cannot go back and export the model in a different format. Not being able to change the export format is a major regression.

So what should be done allow choice or not

src/heretic/main.py
Vinay-Umrethe: yup I had used `try/finally` but later removed, gonna restore

src/heretic/utils.py
Vinay-Umrethe: > I still wonder what to do with version 1 of reproduce.json (as it does not seed the fix done in model.py)

> Leave it dead OR if it's v1, use "settings" and run all trials sequentially instead of direct applying which is possible *now*

like said we can do this

p-e-w: I think this is already handled nicely by the environment check.

We will release the auto-reproduction feature in Heretic 1.4. The old `reproduce.json` files that were missing the seed fix were made using Heretic 1.3. This will automatically get flagged as a "critical" mismatch, so there is no expectation of byte-for-byte reproducibility.

src/heretic/main.py
p-e-w: The choice should be allowed *except* in auto-reproduction mode. If you remove the `original_export_strategy` logic, everything will be correct.

There should be no choice when the user has explicitly said they want to reproduce.

p-e-w:
Ok, I've merged this. I'll decide what to do about the version number when I wrap up #326.

p-e-w:
Ah, I just realized we don't need the version number distinction to handle the export strategy. That's because with the new code, the export strategy is always set, whereas before, it was never set. But until the upcoming version 1.4, there was only one export strategy: merge.

So if the strategy is missing from the settings, we can just assume it is "merge", and it will always be correct.

PR #353, opened 2026-06-04
feat: save processor for multimodal models

zaakirio:
VL models (Gemma 3/4, Qwen-VL, etc.) load via AutoModelForImageTextToText (added in #108), but only the tokenizer was ever saved/pushed. 

The processor, which holds the image/audio preprocessing config (preprocessor_config.json, processor_config.json) was dropped, so the saved/uploaded abliterated model could no longer be loaded with AutoProcessor and was unusable for its multimodal inputs.

Load the processor alongside the tokenizer (None for text-only models, where AutoProcessor.from_pretrained raises) and save/push it wherever the tokenizer is saved/pushed. Abliteration itself is unchanged, it only touches the text path; this just keeps the resulting model complete.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -17,6 +17,7 @@ from torch.nn import Module, ModuleList
 from transformers import (
     AutoModelForCausalLM,
     AutoModelForImageTextToText,
+    AutoProcessor,
     AutoTokenizer,
     BatchEncoding,
     BitsAndBytesConfig,
@@ -56,6 +57,8 @@ class AbliterationParameters:
 class Model:
     model: PreTrainedModel | PeftModel
     tokenizer: PreTrainedTokenizerBase
+    # Set for multimodal models, None for text-only ones.
+    processor: Any | None
     peft_config: LoraConfig
 
     def __init__(self, settings: Settings):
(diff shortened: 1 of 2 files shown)


src/heretic/model.py
```diff
@@ -56,6 +57,8 @@ class AbliterationParameters:
 class Model:
     model: PreTrainedModel | PeftModel
     tokenizer: PreTrainedTokenizerBase
+    # Set for multimodal models, None for text-only ones.
+    processor: Any | None
```
p-e-w: Is there no better type than `Any`? Note that `Any` includes `None`, so the current annotation doesn't make sense. `Any | None` is the same as `Any`.

src/heretic/model.py
```diff
@@ -75,6 +78,15 @@ def __init__(self, settings: Settings):
             **self.revision_kwargs,
         )
 
+        # Multimodal models have a processor we'll want to save.
+        self.processor = None
+        if get_model_class(settings.model) == AutoModelForImageTextToText:
```
p-e-w, line 84: Does every model of type `AutoModelForImageTextToText` have a processor?

p-e-w (commented):
Thanks for the PR! Which models have you tested this with? Can you link to a model upload?

src/heretic/model.py
zaakirio: Yep, fair point - replaced with [ProcessorMixin](https://huggingface.co/docs/transformers/main/en/main_classes/processors)

zaakirio:
The model which I have tested this with was:
[google/gemma-4-12B-it](https://huggingface.co/google/gemma-4-12B-it)

Result after using heretic, and tested working:
[zaakirio/gemma-4-12b-it-uncensored](https://huggingface.co/zaakirio/gemma-4-12b-it-uncensored)

src/heretic/model.py
zaakirio: Effectively yes. Every architecture loadable as `AutoModelForImageTextToText` either has a static `AutoProcessor` [mapping](https://github.com/huggingface/transformers/blob/v5.6.2/src/transformers/models/auto/processing_auto.py#L49) , or resolves via the processor_class declared in the checkpoint's processor_config.json, which `AutoProcessor.from_pretrained `checks first

p-e-w:
Merged, thanks! This is a great improvement.

PR #354, opened 2026-06-05
feat: add native Windows AMD RDNA2 (gfx1030) support

Matlan1:
## Summary

Adds support for running `heretic` on **Windows 11** with **AMD RDNA2 GPUs** (RX 6700 XT / RX 6800 / RX 6900 XT — `gfx1030`/`gfx1031`/`gfx1032`) via AMD's TheRock ROCm 7.13 distribution at `repo.amd.com/rocm/whl/gfx103X-all/`.

**Verified end-to-end on a completely fresh clone** (no pre-existing venv or cached wheels):
- AMD Radeon RX 6900 XT (gfx1030), Windows 11, Python 3.12, ROCm 7.13 / HIP 7.13.99004
- `torch.cuda.is_available()` → `True`
- `torch.cuda.get_device_name(0)` → `AMD Radeon RX 6900 XT`
- `heretic --model Qwen/Qwen2.5-0.5B-Instruct --quantization NONE` → model loaded on GPU, abliteration running at speed

---

### `WINDOWS_ROCM.md` (new file)

Step-by-step guide for Windows RDNA2 users. Key points:

**1. `uv sync` first** (standard deps from lockfile, creates the venv)

**2. Install ROCm wheels with required flags:**

```powershell
uv pip install `
  --extra-index-url https://repo.amd.com/rocm/whl/gfx103X-all/ `
  --index-strategy unsafe-best-match `
  --exclude-newer-package rocm=false `
  "https://repo.amd.com/rocm/whl/gfx103X-all/torch-2.9.1%2Brocm7.13.0-cp312-cp312-win_amd64.whl" `
  "https://repo.amd.com/rocm/whl/gfx103X-all/rocm_sdk_core-7.13.0-py3-none-win_amd64.whl" `
  "https://repo.amd.com/rocm/whl/gfx103X-all/rocm_sdk_libraries_gfx103x_all-7.13.0-py3-none-win_amd64.whl"
```

The three flags are required because:
- `--extra-index-url` + `--index-strategy unsafe-best-match` — `torch+rocm7.13` depends on a `rocm` package that is not on PyPI; without these, `uv` cannot resolve the dependency and fails with `No solution found: rocm[libraries]==7.13.0`
- `--exclude-newer-package rocm=false` — the project's `exclude-newer = "7 days"` setting in `pyproject.toml` would otherwise filter out the `rocm` source tarball (which has no upload timestamp)

**3. Run using direct venv paths, not `uv run`:**

```powershell
.venv\Scripts\heretic.exe --model <model-id> --quantization NONE
```

`uv run` re-syncs the venv against `pyproject.toml` before executing, which overwrites the ROCm torch wheel with the CPU-only PyPI build, silently breaking GPU detection.

---

## What is NOT changed

- `pyproject.toml` — no dependency changes; ROCm is intentionally kept out of the lockfile since it targets a specific OS+GPU combination
- `uv.lock` — untouched
- All existing source files — no Python code changes

---

## Known Limitations

| Feature | Status |
|---|---|
| GPU inference (FP16/BF16) | ✅ Working |
| `--quantization NONE` | ✅ Working |
| `--quantization bnb_4bit` | ❌ `bitsandbytes` has no ROCm Windows DLLs |
| `torchvision` / `torchaudio` | ⚠️ Not installed (not required by heretic) |

### DIFF
--- a/WINDOWS_ROCM.md
+++ b/WINDOWS_ROCM.md
@@ -1,0 +1,109 @@
+# Running heretic on Windows with an AMD GPU (RDNA2 / gfx103X)
+
+This guide covers running `heretic` natively on **Windows 11** with **AMD RDNA2 GPUs**
+(RX 6700 XT, RX 6800, RX 6800 XT, RX 6900 XT, etc. — `gfx1030`/`gfx1031`/`gfx1032`).
+
+> **Status: Working ✅**  
+> Verified on: AMD Radeon RX 6900 XT (gfx1030), Windows 11, Python 3.12, ROCm 7.13 / HIP 7.13.99004
+
+---
+
+## Prerequisites
+
+- Windows 11 (22H2 or later)
+- Python **3.12** (exact — `uv` will enforce this)
+- [uv](https://docs.astral.sh/uv/getting-started/installation/) package manager
+- AMD Adrenalin driver **24.x** or later (shipped with WDDM 3.x support)
+- ~12 GB free disk space for ROCm SDK wheels
+
+---
+
+## Installation
+
+### 1. Clone and enter the repository
+
+```powershell
+git clone https://github.com/p-e-w/heretic.git
+cd heretic
+```
+
+### 2. Create venv and install standard dependencies
+
+```powershell
+uv sync
+```
+
+### 3. Install AMD ROCm 7.13 PyTorch and SDK wheels
+
+These wheels are served from AMD's official index and target the `gfx103X` RDNA2 family:
+
+```powershell
+uv pip install `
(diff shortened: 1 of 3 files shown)


p-e-w:
Thanks for the PR, but most of the instructions don't seem to be specific to Heretic. It's really mostly about getting PyTorch to work on that setup. Documentation can become a maintenance burden, so I'd rather only host instructions that are specific to Heretic, rather than "how do I get a program that uses PyTorch to run on system X"? Such instructions belong elsewhere. Same with the PowerShell instructions, those are generic. Also, you don't have to pass `--quantization none`, that's the default.

Matlan1:
Understood. Thanks for the feedback, I'll close this PR.

PR #355, opened 2026-06-05
fix: handle HFValidationError in is_hf_path

umran666:
Hi! 

I noticed that when `is_hf_path()` is called with an invalid repository/path string, the application crashes with an unhandled `HFValidationError` from `huggingface_hub` instead of just returning `False`.

To fix this, I imported `HFValidationError` and wrapped the `validate_repo_id(path)` call in a try/except block. Now it safely catches the validation error and returns `False` as expected.

Thanks for reviewing!

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -22,7 +22,7 @@ from datasets import DatasetDict, ReadInstruction, load_dataset, load_from_disk
 from datasets.config import DATASET_STATE_JSON_FILENAME
 from datasets.download.download_manager import DownloadMode
 from datasets.utils.info_utils import VerificationMode
-from huggingface_hub.utils import validate_repo_id
+from huggingface_hub.utils import HFValidationError, validate_repo_id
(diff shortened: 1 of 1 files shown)


p-e-w:
> I noticed that when is_hf_path() is called with an invalid repository/path string, the application crashes with an unhandled HFValidationError from huggingface_hub instead of just returning False.

This is deliberate, because that situation is a fatal error condition that we cannot recover from.

We need to be able to load model/dataset repositories, not doing so is not an option. Thus the path needs to be either a local path (which `is_hf_path` checks first) or a HF repository. If it is neither, then it's Game Over, we can't do anything other than exit. Returning `False` will make the system assume that we are dealing with a local directory, but we already know that isn't true. Blowing up is the correct action here.

umran666:
Ah, that makes perfect sense. Returning False would just lead to a confusing file-not-found error on a local directory check later anyway.

Feel free to close this PR!

PR #357, opened 2026-06-05
fix: count all trials, not just completed trials

umran666:
Hi there,

This PR resolves issue #311 where the progress output can display "Running trial N of M" where N > M.

### The Problem
When the user chooses "Run additional trials" from the trial selection menu:
1. `settings.n_trials` is incremented.
2. `study.optimize()` is called again to run the new trials.
3. However, the trial progress variables (`trial_index`, `start_index`, and `start_time`) are not re-initialized before the new optimize run.
4. Since `trial_index` retains its old value from the end of the previous run, the count of attempted trials exceeds the new total trial budget if any previous trials were pruned or failed.

### The Solution
Re-initialized `start_index`, `trial_index`, and `start_time` right before calling `study.optimize()` in the "Run additional trials" block. This keeps the progress and estimated remaining time calculations correct for the new set of trials.

Thanks for taking a look!

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -615,11 +615,7 @@ def run():
     study.set_user_attr("settings", settings.model_dump_json())
     study.set_user_attr("finished", False)
 
-    def count_completed_trials() -> int:
-        # Count number of complete trials to compute trials to run.
-        return sum([(1 if t.state == TrialState.COMPLETE else 0) for t in study.trials])
-
-    start_index = trial_index = count_completed_trials()
+    start_index = trial_index = len(study.trials)
     if start_index > 0:
         print()
         print("Resuming existing study.")
@@ -627,7 +623,7 @@ def run():
     try:
         study.optimize(
             objective_wrapper,
-            n_trials=settings.n_trials - count_completed_trials(),
+            n_trials=settings.n_trials - len(study.trials),
         )
     except KeyboardInterrupt:
         # This additional handler takes care of the small chance that KeyboardInterrupt
@@ -635,7 +631,7 @@ def run():
         # defined in objective_wrapper above.
         pass
 
-    if count_completed_trials() == settings.n_trials:
+    if len(study.trials) == settings.n_trials:
         study.set_user_attr("finished", True)
 
     while True:
@@ -733,12 +729,12 @@ def run():
                 try:
                     study.optimize(
                         objective_wrapper,
-                        n_trials=settings.n_trials - count_completed_trials(),
+                        n_trials=settings.n_trials - len(study.trials),
                     )
                 except KeyboardInterrupt:
                     pass
 
-                if count_completed_trials() == settings.n_trials:
+                if len(study.trials) == settings.n_trials:
                     study.set_user_attr("finished", True)
 
                 break
@@ -971,7 +967,7 @@ def run():
                             if reproducibility_information != "none":
                                 # Set the number of trials to the number of actual completed trials
                                 # for the reproduction configuration.
-                                settings.n_trials = count_completed_trials()
+                                settings.n_trials = len(study.trials)
 
                                 upload_reproduce_folder(
                                     repo_id,


p-e-w:
I don't understand. Why should we reset `trial_index`? We want to continue counting from where we left off. The index should count all trials that were run, not just completed ones, otherwise we might repeat indices.

Can you please describe step-by-step a concrete scenario that you think can cause #311, tracking the value of `trial_index` along the way?

Please note #324, where the exact same discussion arose, but I did not get a reply to my last question.

umran666:
Hi there,

Here is a step-by-step trace of how the " `N > M` situation happens under the current logic" :

### How to reproduce:

1. Start `heretic` with `--n-trials 20`.
2. Let trials 1 through 5 run and complete successfully (`trial_index` goes from 0 to 5).
3. On Trial 6, press `Ctrl+C` to skip/prune it. The optimize loop catches the interrupt and returns to the main menu.
4. Because Trial 6 didn't complete, `count_completed_trials()` is 5.
5. In the menu, select **"Run additional trials"** and add `5` more.
6. The new target of completed trials (`settings.n_trials`) becomes `25`.
7. The program calculates how many trials to run next: `25 - 5 = 20`. So it starts `study.optimize()` again to run 20 more trials.
8. Since `trial_index` is a nonlocal variable and wasn't reset, it is still `6`.
9. As it runs the 20 new trials, `trial_index` increments 20 times, ending at `26`.
10. On the last trial, it prints: `Running trial 26 of 25...`

---

### The Trade-off

I see your point about not wanting to reuse index numbers (e.g., if Trial 6 fails, the next one shouldn't also be labeled "6"). 

If we want to keep indexing trials uniquely but avoid the `N > M` anomaly, here are a few options:

1. **Change the status message to separate attempts from progress**:
   `Running attempt N (Completed: C of M)...` (e.g., `Running attempt 26 (Completed: 19 of 25)...`)
2. **Remove the denominator entirely**:
   `Running trial N...` (without the `of M`).
3. **Reset `trial_index` to `len(study.trials)`**: This ensures we don't repeat indices, but the target `M` (which represents completed trials) will still eventually be exceeded if there are failed/pruned trials in the database.

Let me know which option you think fits best!

p-e-w:
> The new target of completed trials (settings.n_trials) becomes 25.

This contradicts how Optuna is supposed to work. `n_trials` in `study.optimize` is supposed to be the number of **total** trials, not successful trials. In fact, changing that [has been proposed](https://github.com/optuna/optuna/issues/293) but not implemented.

umran666:
You are completely right. I agree that we shouldn't reset or reuse indices, as failed/pruned trials still existed and should retain their unique numbers.

### The Root Cause
The reason `N > M` occurred was a mismatch in how progress is counted:
1. `settings.n_trials` was treated as the target for **completed** trials:
   `n_trials = settings.n_trials - count_completed_trials()`
   This caused Optuna to run extra attempts to make up for failed/pruned trials.
2. But the progress indicator was showing **attempted** trials (`trial_index`).
   Because attempts grew larger than completed trials, the indicator eventually printed things like `Running trial 26 of 25`.

### The Solution
I have updated this PR to align fully with Optuna's standard design (where `settings.n_trials` is the target for **total attempted** trials):

1. **Replaced `count_completed_trials()` with `len(study.trials)`** when calculating remaining trials and checking if finished.
2. **Removed `trial_index` tracking entirely** and replaced it with Optuna's built-in `trial.number + 1`. This ensures:
   - Trial numbers are strictly sequential, unique, and never repeated.
   - The progress indicator naturally caps at `settings.n_trials`.
   - The code is simplified by eliminating manual counter state variables.

Thanks for the guidance!

p-e-w:
> Removed trial_index tracking entirely and replaced it with Optuna's built-in trial.number + 1.

Please don't. The trial index is used in many other places besides the ones you changed. We don't modify such core logic without a good reason. I used my own implementation because the Optuna docs are ambiguous about what `trial.number` is exactly.

As as side note, I'd appreciate if you could engage with me using human-written messages, like I am doing with you. Reading "The Problem"/"The Solution" multiple times in a row is very exhausting.

umran666:
First of all, my apologies for the use of the Problem/Solution layout to describe the code - that's been a tendency for many years for me when describing code, but I understand it's pretty exhausting.


I have undone the modifications I made regarding `trial.number` and preserved the logic using `trial_index`, just like the initial one.
To overcome the N > M problem, and avoid repetitions, I have aligned the code with Optuna's architecture, considering `settings.n_trials` as the total amount of attempted trials:
- Replacing `count_completed_trials()` with `len(study.trials)`.
- Re-initializing `trial_index` with `len(study.trials)` at the beginning of the optimization run.

src/heretic/main.py
```diff
@@ -615,27 +615,24 @@ def objective_wrapper(trial: Trial) -> tuple[float, float]:
     study.set_user_attr("settings", settings.model_dump_json())
     study.set_user_attr("finished", False)
 
-    def count_completed_trials() -> int:
-        # Count number of complete trials to compute trials to run.
-        return sum([(1 if t.state == TrialState.COMPLETE else 0) for t in study.trials])
-
-    start_index = trial_index = count_completed_trials()
+    start_index = trial_index = len(study.trials)
+    start_time = time.perf_counter()
```
p-e-w: What does this do? The time is already set above.

src/heretic/main.py
```diff
@@ -730,15 +727,18 @@ def count_completed_trials() -> int:
                 study.set_user_attr("settings", settings.model_dump_json())
                 study.set_user_attr("finished", False)
 
+                start_index = trial_index = len(study.trials)
+                start_time = time.perf_counter()
```
p-e-w: I don't think we should reset the time. This will lead to a wrong estimation of remaining time.

src/heretic/main.py
```diff
@@ -730,15 +727,18 @@ def count_completed_trials() -> int:
                 study.set_user_attr("settings", settings.model_dump_json())
                 study.set_user_attr("finished", False)
 
+                start_index = trial_index = len(study.trials)
```
p-e-w: Why are you resetting this here?

p-e-w (commented):
I don't think the logic is sound yet.

umran666:
Oh i see it now ! In the previous iteration,I tried to isolate the additional trials run as its own separate batch by resetting start_index, trial_index, and start_time inside the loop.
Because both runs happen within the same interactive CLI session
"Resetting the timer threw away all the timing data from the first batch of trials.This forced the remaining time estimator to start from scratch, leading to an unstable and inaccurate estimate for the remaining additional trials."

my bad man!

Sorry about the previous iteration! I've updated it:
By removing the redundant `start_time` reset,
Also removed the resets of `start_time` and `start_index` in the additional trials block to preserve continuity for the remaining time calculation.
And kept the shift to `len(study.trials)` so we count all attempted trials correctly.

p-e-w:
Yes, it seems correct now. Have you tested it in all of the following scenarios

* Regular run.
* Run with additional trials.
* Run with failed trials + resumption.
* Resuming from checkpoint.

and verified that everything looks OK?

umran666:
yes, I've tested all those cases and everything works correctly. 

For the regular runs and resumptions, the math holds up. Since we don't reset start_time and start_index when adding more trials, the remaining time estimator stays stable because it correctly measures the rate of trials run in the current session. 

If you resume from a checkpoint, the timer starts fresh and calculates speed based on the new trials, which is great because it won't get skewed if you resume on different hardware. Pruned and failed trials also keep their unique indices, so we don't repeat any numbers.

p-e-w:
Ok, thanks. Merged!

PR #359, opened 2026-06-06
style: remove annoying gray bg from logo

Vinay-Umrethe:
### Before

before

### After

after


p-e-w:
Merged, thanks!

The gray background wasn't there from the beginning, but a few weeks after I added the logo, GitHub changed their stylesheet and it appeared. It annoyed me too, glad you found a fix.

PR #361, opened 2026-06-07
fix: add Triton and PyTorch compatibility checks for MXFP4 models

umran666:
This PR adds an early configuration check to detect if a model uses MXFP4 quantization. If it does, we perform version validation on PyTorch and Triton. If PyTorch is < 2.6 or Triton is < 3.5 (which fail to compile kernels on older GPUs like the A100), we print an early warning. If model loading subsequently fails, we intercept the error to raise a clear and descriptive exception pointing out the version upgrade requirement, instead of crashing with a cryptic traceback.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -150,7 +150,10 @@ class Model:
             except Exception as error:
                 self.model = None  # ty:ignore[invalid-assignment]
                 empty_cache()
-                print(f"* [red]Failed[/] ({error})")
+                error_message = str(error)
+                if any(k in error_message.lower() for k in ["triton", "mxfp4"]):
+                    error_message += " (Hint: MXFP4 quantized models may require Triton >= 3.5 and PyTorch >= 2.6 on older GPUs.)"
+                print(f"* [red]Failed[/] ({error_message})")
                 continue
 
             if settings.quantization == QuantizationMethod.BNB_4BIT:


p-e-w:
I'm not comfortable maintaining that much code for such an edge case. MXFP4 is quite rare already, and then the combination of older library versions on top of it.

We can't handle every possible configuration by cluttering our codebase with checks, users have to figure out such things themselves.

umran666:
> I'm not comfortable maintaining that much code for such an edge case. MXFP4 is quite rare already, and then the combination of older library versions on top of it.
> 
> We can't handle every possible configuration by cluttering our codebase with checks, users have to figure out such things themselves.

That makes complete sense. I've removed the pre-flight checks, version parsing, and custom exceptions entirely. Instead, I simplified it to just append a quick hint to the load failure output if the error message contains 'triton' or 'mxfp4'. The diff is now down to just 4 lines of code with zero maintenance overhead.

p-e-w:
What does the error message say by itself? Is it not explanatory already?

umran666:
By default, if they try to load it without the right Triton version, transformers raises: ImportError: MXFP4 quantization requires "Triton and kernels installed: CUDA requires Triton >= 3.4.0" . That part is clear enough. However, once the user upgrades Triton to >= 3.4 to fix that, loading it on PyTorch < 2.6 fails with: AttributeError: module 'torch' has no attribute 'accelerator'

The second error is what gets cryptic (since torch.accelerator was only added in PyTorch 2.6).

If you'd rather not have any extra code/hints in the repo for this combination, we can absolutely just close this PR and let users search issues if they hit it. Let me know what you prefer!

p-e-w:
Yeah, that error message is clear enough.

BTW, the `torch.accelerator` issue is [specifically documented in the README](https://github.com/p-e-w/heretic#usage).

I'm closing this PR as the status quo is good enough IMO.

Issue #248, opened 2026-03-23
[RFC] A better automatic batch size mechanism


p-e-w:
As mentioned in #239, I have come to believe that the way I originally implemented automatic batch size determination in Heretic is flawed. There are several problems with the current mechanism:

* It wastes time with inference on test prompts, then throws away the results.
* It tries to determine a global batch size that applies to all operations, even though generating responses requires more VRAM than generating logprobs.
* It fails to adapt to changing memory pressure on the system. If another process starts to take up VRAM, Heretic will crash with an OOM error.
* It starts from a batch size of 1 and incrementally doubles the batch size until it OOMs to determine the optimum. This is problematic for two reasons:
  * It can spend a lot of time doing inference and re-generating compute graphs for low batch sizes, even though the maximum of 128 often works.
  * It might miss major performance gains for sizes that aren't powers of 2. For example, a system might be able to support a batch size of 100, but not 128. Choosing 100 instead of 64 can give huge performance gains with the default configuration, because there are 100 evaluation prompts per set.

Here is how I think it should work instead:

* Every inference method (`get_responses`, `get_residuals`, `get_logprobs`) has its own associated batch size, initialized to a value like 128.
* There is no batch size determination process on program startup.
* Instead, when an inference method is called, it just runs with its current batch size. If inference throws an OOM error, it reduces the batch size and tries again. The reduced batch size is stored in a `Model` instance field, so it is remembered for the next invocation.

This avoids wasted work, optimizes the batch size separately for different methods, and dynamically adapts to changing memory availability at runtime.

While this discards the "benchmarking" aspect and thus might lead to situations where larger batch sizes are chosen while a smaller batch size could improve overall throughput, I have found those situations to be extremely rare in practice. For almost every (non-CPU) run I have ever seen, you get the highest generation speed by simply choosing the largest batch size that fits into your VRAM constraints.

PR #362, opened 2026-06-07
binary search for batch size

NoahOksuz:
When `batch_size = 0` (auto), Heretic probes batch sizes exponentially and picks the size with the best measured throughput (tokens/s). Previously, if a probe hit CUDA OOM, the search stopped and kept the last successful power-of-two even though when a larger non-power-of-two batch would fit and run faster. I also saw https://github.com/p-e-w/heretic/issues/248 but i think this is a good solution for now. 

 **Refines the search after OOM**  binary-searches between the last successful size and the first failing size, still choosing the batch with the highest measured tokens/s.
```text
Determining optimal batch size...
* Trying batch size 1... Ok (20 tokens/s)
* Trying batch size 2... Ok (38 tokens/s)
* blah blah
* Trying batch size 128... Ok (814 tokens/s)
* Trying batch size 256... Failed (CUDA out of memory...)
* Trying batch size 192... Ok (902 tokens/s)
* Trying batch size 224... Ok (905 tokens/s)
* Trying batch size 240... Ok (914 tokens/s)
* Trying batch size 248... Failed (CUDA out of memory...)
* Chosen batch size: 240
```

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -37,7 +37,7 @@ offload_outputs_to_cpu = true
 batch_size = 0  # auto
 
 # Maximum batch size to try when automatically determining the optimal batch size.
-max_batch_size = 128
+max_batch_size = 1024
 
 # Maximum number of tokens to generate for each response.
 max_response_length = 100
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -180,7 +180,7 @@ class Settings(BaseSettings):
     )
 
     max_batch_size: int = Field(
-        default=128,
+        default=1024,
         description="Maximum batch size to try when automatically determining the optimal batch size.",
         # When storing a settings object, the batch size is already fixed,
         # either determined by the automatic mechanism or by explicit user choice.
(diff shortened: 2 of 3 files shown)


umran666:
Hey Noah, verified this on my GPU and it works great! I added the fixes for the code review comments (generalizing to OutOfMemoryError and adding defensive guards in _determine_batch_size). Feel free to pull them from my branch: `https://github.com/umran666/heretic/tree/fix/binary-search-batch-size`

p-e-w:
The current approach is not ideal, but a binary search is not the right solution either IMO, at least not in this form. Here's the crux of the problem:

**We aren't actually interested in finding the largest possible batch size. There is another constraint, and that is the size of the prompt dataset(s).**

By default, the evaluation datasets contain 100 prompts each. That means choosing any batch size greater than 100 doesn't make sense, because we'll never run more than 100 prompts anyway.

In fact, the only batch sizes that we should consider are `ceil(len(prompts) / n)` for `n = 1, 2, 3, ...`. Choosing any other size will always be dominated by the next smaller member of this set, because you need the same number of batches but are less robust against VRAM fluctuations.

The correct search strategy is to start with a batch size of `len(prompts)`, then proceed with `len(prompts) / 2` if there is an OOM, followed by `len(prompts) / 3` etc. But as noted in #248, this needs to happen on each call to `generate` (with caching), because otherwise, the locked-in batch size from the start of the run leads to a crash if some other process consumes VRAM in between.

Issue #82, opened 2025-12-11
Bug: Unable to load ServiceNow-AI/Apriel-1.6-15b-Thinker


MS00-GitIt:
### Bug: Unable to load `ServiceNow-AI/Apriel-1.6-15b-Thinker` (`LlavaConfig` not supported by AutoModelForCausalLM)

**Command run:**
(heretic) ####@####:~/heretic$ heretic ServiceNow-AI/Apriel-1.6-15b-Thinker

**Expected:** Model loads or Heretic reports that Llava-based / multimodal models are unsupported.  
**Actual:** All dtype attempts fail with `Unrecognized configuration class LlavaConfig` and Heretic exits with `Exception: Failed to load model with all configured dtypes.`

---

### Environment
- OS: Bazzite (Fedora-based Linux)
- Python: 3.12 (venv)
- GPU: NVIDIA RTX 3090
- Heretic version: v1.1.0
- transformers: version bundled with Heretic (shows HF cache deprecation warning)

---

### Full Output
(heretic) ####@####:~/heretic$ heretic ServiceNow-AI/Apriel-1.6-15b-Thinker
/var/home/####/heretic/.venv/lib/python3.12/site-packages/transformers/utils/hub.py:110: FutureWarning: Using `TRANSFORMERS_CACHE` is deprecated and will be removed in v5 of Transformers. Use `HF_HOME` instead.
  warnings.warn(
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.1.0
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░
▀░▀░▀▀▀░▀░▀░▀▀▀░░▀░░▀░▀▀▀  https://github.com/p-e-w/heretic

GPU type: NVIDIA GeForce RTX 3090

Loading model ServiceNow-AI/Apriel-1.6-15b-Thinker...
[...]

PR #364, opened 2026-06-08
fix: gracefully handle processor load failures for multimodal models (#82)

umran666:
Catches configuration/dependency load failures when initializing a multimodal model's processor, logging a warning rather than crashing the model initialization loop. This allows multimodal models to be abliterated on text-only pipelines even if optional vision dependencies (such as torchvision) are missing.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -88,7 +88,7 @@ class Model:
                     trust_remote_code=settings.trust_remote_code,
                     **self.revision_kwargs,
                 )
-            except Exception as error:
+            except (ImportError, OSError, ValueError, KeyError) as error:
                 print(
                     f"* [yellow]Warning[/]: Could not load processor ({error}). "
                     f"Multimodal saving/pushing may be incomplete."
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -82,11 +82,17 @@ class Model:
         # Multimodal models have a processor we'll want to save.
         self.processor = None
         if get_model_class(settings.model) == AutoModelForImageTextToText:
-            self.processor = AutoProcessor.from_pretrained(
-                settings.model,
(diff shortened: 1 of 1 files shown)


p-e-w:
Does merely *loading* a processor require torchvision? That would surprise me.

umran666:
Yeah, it actually does because `Apriel-1.6-15b-Thinker` uses PixtralProcessor.
In Hugging Face transformers, the PixtralProcessor class has a hard requirement for torchvision defined in its backend checks. As soon as you call AutoProcessor.from_pretrained(), it tries to load the class, runs that backend check, and immediately raises an ImportError if it's missing.
Since torchvision isn't in heretic's dependencies, it crashes on startup for anyone who doesn't have it installed. Catching this exception is a simple way to let users still run text abliteration on these models without needing to install extra vision libraries.

p-e-w:
Interesting, thanks for identifying this problem.

I'll need to think about what the best fix is here though. In general, I prefer Heretic to work out of the box in all reasonable situations, so simply depending on torchvision might be the better solution.

umran666:
Adding torchvision to dependencies in pyproject.toml is definitely cleaner for out-of-the-box multimodal support, though the main risk is dependency resolution conflicts since torchvision releases are tightly coupled to specific torch versions (which can sometimes get messy during installation on custom environments).

Let me know if you'd prefer me to update this PR to add torchvision to pyproject.toml (with or without the fallback catch block), or if we should stick to the defensive error-trapping warning.

p-e-w:
No, you don't need to do anything at this point. I'll have to think about this some more, and solving this niche issue isn't a high priority for me.

p-e-w:
Torchvision is now a dependency of Heretic, so this PR is obsolete.

PR #366, opened 2026-06-10
fix: preserve Hugging Face token in session to avoid repeated prompt (#142)

umran666:
Currently, when using the interactive menu to save/upload model trials, users are prompted to enter their Hugging Face access token for every single upload attempt. This becomes repetitive if uploading multiple trials, pushing both the adapter and merged model, or retrying after a failed upload.

This PR introduces a session-level token cache that persists the credentials only for the duration of the CLI session. To keep it secure and avoid saving bad inputs, the token is cached only after it is successfully validated byhuggingface_hub.whoami(). Subsequent uploads in the same run reuse the cached token without persisting it to disk.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -856,7 +856,9 @@ def run():
                                         break
 
                                     if choice == "Switch account":
-                                        token = prompt_password("Hugging Face access token:")
+                                        token = prompt_password(
+                                            "Hugging Face access token:"
+                                        )
                                         if not token:
                                             break
                                         continue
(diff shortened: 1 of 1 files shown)


p-e-w:
See #148 for comparison, and the multiple caveats I raised in that PR.

umran666:
Thanks for the pointer! I've refactored the PR to match the behavior you sketched out in #148.

The token validation is now wrapped in a loop that checks the cached/system token first, displays the account, and asks whether to proceed or switch. If anything fails or the user decides to switch, it prompts for a new token and re-validates. 
This should make it much smoother to use without getting stuck on bad tokens. Let me know what you think!

p-e-w:
Thanks, I'm going to prepare the 1.4 release over the next few days and then review this PR in detail afterwards.

umran666:
Can you look into this

p-e-w:
Sorry, I currently don't have the bandwidth to review this PR.

Issue #77, opened 2025-12-09
Early abort of trials fails if study is cancelled while reloading the model


p-e-w:
The dtype cannot be obtained from the model object in this case because it has been set to `None`.

Vinayyyy7:
can we have more context of what's the exact issue in your tests?

p-e-w:
It's just a simple bug that arises when you cancel (Ctrl+C) a trial while it is reloading the model, then try to load one of the versions from the Pareto front. Since the model has been set to `None`, `model.dtype` isn't available, and an error is thrown.

I'll fix this soon. It's not a regression but an oversight that has been there from the start.

PR #367, opened 2026-06-10
fix: make reset_model null-safe to handle study cancellations (#77)

umran666:
When a study is cancelled (via Ctrl+C) during a model reload, `self.model` is left as `None`. Subsequent calls to `reset_model()` fail with an `AttributeError` when trying to read `self.model.config` or `self.model.dtype`.

This PR introduces a cached `self.dtype` attribute on the `Model` class, populates it upon successful load, and updates `reset_model()` to safely check for nullity and fall back to the cached `self.dtype` when `self.model` is `None`. This allows the CLI to recover and successfully reload the model when selecting a new trial after a cancellation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,6 @@ dependencies = [
     "tomli-w~=1.2",
     "tqdm~=4.67",
     "transformers[kernels]~=5.6",
-    "torch",
 ]
 
 [project.optional-dependencies]
@@ -75,14 +74,5 @@ build-backend = "uv_build"
 [tool.uv]
 exclude-newer = "7 days"
 
-[tool.uv.sources]
-torch = { index = "pytorch" }
-
-[[tool.uv.index]]
-name = "pytorch"
-url = "https://download.pytorch.org/whl/cu124"
-explicit = true
-
 [tool.uv.build-backend]
 module-name = "heretic"
-
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dependencies = [
     "tomli-w~=1.2",
     "tqdm~=4.67",
     "transformers[kernels]~=5.6",
+    "torch",
 ]
 
 [project.optional-dependencies]
@@ -74,5 +75,14 @@ build-backend = "uv_build"
 [tool.uv]
 exclude-newer = "7 days"
 
+[tool.uv.sources]
+torch = { index = "pytorch" }
+
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cu124"
+explicit = true
+
 [tool.uv.build-backend]
 module-name = "heretic"
+
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -61,7 +61,7 @@ class Model:
     # Set for multimodal models, None for text-only ones.
     processor: ProcessorMixin | None
     peft_config: LoraConfig
-    dtype: torch.dtype | str | None
+    dtype: torch.dtype
 
     def __init__(self, settings: Settings):
         self.settings = settings
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -66,7 +66,6 @@ class Model:
     def __init__(self, settings: Settings):
         self.settings = settings
         self.needs_reload = False
-        self.dtype = None
 
         self.revision_kwargs = {}
         if settings.model_commit is not None:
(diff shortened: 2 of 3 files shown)


src/heretic/model.py
```diff
@@ -317,15 +321,26 @@ def reset_model(self):
...
             return
 
-        dtype = self.model.dtype
+        if self.model is not None:
+            dtype = self.model.dtype
+            self.dtype = dtype
+        else:
+            dtype = self.dtype or (
+                self.settings.dtypes[0] if self.settings.dtypes else "auto"
+            )
```
p-e-w: Why not simply always user `self.dtype` here? It's always set, and it's always the right dtype, no?

src/heretic/model.py
```diff
@@ -129,6 +131,8 @@ def __init__(self, settings: Settings):
                     **self.revision_kwargs,
                     **extra_kwargs,
                 )
+                if self.model is not None:
```
p-e-w: How can `self.model` be `None` here?

p-e-w (commented):
Thanks for tackling this problem, though I think this PR can be substantially simplified.

BTW, the PR description is mangled with missing spaces, presumably from copy&paste or a broken agent. I don't mind contributors using LLMs (although I don't use them myself), but please do at least a basic sanity check before clicking "submit".

src/heretic/model.py
umran666: Good catch. I just pushed an update to use `self.dtype` directly.

src/heretic/model.py
umran666: Oh yea It can't be. `from_pretrained` never returns `None`. I've removed the redundant `None` checks in the latest commit.

umran666:
Ah, my apologies for the formatting. You're completely right—a local tokenizer bug stripped the spacing around the inline code backticks when the description text was formatted. 

I've edited the PR description to restore the spaces, and pushed the updates to simplify the code changes as we discussed. Thanks for catching that.

src/heretic/model.py
```diff
@@ -61,10 +61,12 @@ class Model:
     # Set for multimodal models, None for text-only ones.
     processor: ProcessorMixin | None
     peft_config: LoraConfig
+    dtype: torch.dtype | str | None
 
     def __init__(self, settings: Settings):
         self.settings = settings
         self.needs_reload = False
+        self.dtype = None
```
p-e-w: I don't think we need this either. The dtype will always be set below, otherwise an exception bubbles up.

src/heretic/model.py
```diff
@@ -347,6 +357,7 @@ def reset_model(self):
             **self.revision_kwargs,
             **extra_kwargs,
         )
+        self.dtype = self.model.dtype
```
p-e-w: We don't need this. The dtype must remain constant throughout the run.

src/heretic/model.py
```diff
@@ -317,23 +320,30 @@ def reset_model(self):
...
+            assert self.model is not None
             for name, module in self.model.named_modules():
                 if "lora_B" in name and hasattr(module, "weight"):
                     torch.nn.init.zeros_(module.weight)
             return
 
-        dtype = self.model.dtype
+        # self.dtype is populated on successful load in __init__ or reload,
+        # providing a safe fallback if the model object is currently None.
+        dtype = self.dtype
```
p-e-w: No need for a separate variable, just inline `self.dtype` below.

src/heretic/model.py
```diff
@@ -317,23 +320,30 @@ def reset_model(self):
...
-        current_model = getattr(self.model.config, "name_or_path", None)
+        # If a prior model load was interrupted/cancelled mid-process, self.model will be None.
+        current_model = None
+        if self.model is not None:
+            current_model = getattr(self.model.config, "name_or_path", None)
+
         if current_model == self.settings.model and not self.needs_reload:
-            # Reset LoRA adapters to zero (identity transformation)
+            # Reset LoRA adapters to zero (identity transformation).
+            assert self.model is not None
```
p-e-w: Not sure what the purpose of this assertion is. This condition is guaranteed to be true by the logic above, because if `self.model` is `None`, `current_model` is also `None`, and this `if` branch will never be taken because `self.settings.model` cannot be `None`.

src/heretic/model.py
umran666: Makes sense. Removed the redundant initialization.

src/heretic/model.py
umran666: Agreed. Removed this re-assignment since `self.dtype` is already set from the initial load.

src/heretic/model.py
umran666: Done, inlined `self.dtype` directly.

src/heretic/model.py
umran666: You're right, that was completely redundant. I've removed the assertion.

src/heretic/model.py
```diff
@@ -61,6 +61,7 @@ class Model:
     # Set for multimodal models, None for text-only ones.
     processor: ProcessorMixin | None
     peft_config: LoraConfig
+    dtype: torch.dtype | str | None
```
p-e-w: No, it can **only** be a `torch.dtype`. That's the type of `model.dtype` according to the docs, and it can never be `None`.

p-e-w (changes requested):

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "tomli-w~=1.2",
     "tqdm~=4.67",
     "transformers[kernels]~=5.6",
+    "torch",
```
p-e-w: Why did you add this dependency?

p-e-w: Please don't let agents blindly do stuff. It's such a hassle to review.

src/heretic/model.py
umran666: Understood. I've updated the Model.dtype type annotation to strictly be torch.dtype and confirmed it passes ty check.

pyproject.toml
umran666: Apologies for the noise. The package and lockfile modifications were accidental environment pinning changes introduced by the coding agent during local test setup. I have reverted pyproject.toml and uv.lock to match upstream master exactly, leaving the PR focused solely on the cancel-reload bug fix.

p-e-w:
Merged!

PR #369, opened 2026-06-10
fix: force Qwen2TokenizerFast for qwen2 model types

umran666:
This PR fixes a bug where Qwen2-based models (specifically distilled models like `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B`) output byte-fallback characters (`Ġ`, `Ċ`) and miss spacing when using the CLI chat or evaluation loop.

Some upstream model repositories specify `LlamaTokenizerFast` as the `tokenizer_class` in their `tokenizer_config.json` configuration. When loaded via `AutoTokenizer`, this causes Hugging Face to apply Llama-style token decoding boundaries on Qwen's tiktoken BPE structure, leading to broken character offsets and missing spaces.

- Intercept tokenizer loading by reading the model's configuration dictionary.
- If `model_type` is `"qwen2"`, force the loader to initialize `Qwen2TokenizerFast` directly.
- Standard models continue to fall back to `AutoTokenizer` unmodified.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -73,11 +73,32 @@ class Model:
         print()
         print(f"Loading model [bold]{settings.model}[/]...")
 
-        self.tokenizer = AutoTokenizer.from_pretrained(
+        # Load configuration dictionary to verify model type.
+        # This prevents tokenizers configured with incorrect classes in upstream
+        # config metadata from generating corrupted/space-stripped tokens.
+        config_dict, _ = PretrainedConfig.get_config_dict(
             settings.model,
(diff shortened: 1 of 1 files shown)


p-e-w:
I really think this should be fixed in the model config, and I certainly don't think we should hardcode such hacks into our code.

It sucks that they got it wrong, but we're not going to riddle our code with hacks to clean up their mess.

umran666:
Fair point. Keeping the codebase clean of model-specific workarounds is definitely the right priority. I'll close this PR and see if we can get it resolved upstream on the Hugging Face repository instead.

Issue #289, opened 2026-04-08
Error: cannot access local variable 'active_adapters' where it is not associated with a value


YuanBoXie:
<img width="769" height="87" alt="Image" src="https://github.com/user-attachments/assets/6bd0b040-6ed0-4cd7-8dab-a87225c9c787" />

Saving merged model...
* Merging LoRA adapters into base model...
Error: cannot access local variable 'active_adapters' where it is not associated with a value

p-e-w:
Which model is that? I can't reproduce this.

YuanBoXie:
It's a path to a lora fine-tune model, after I merge the lora weight to base model, it could be solved.
Does Heretic lack default support for LoRA paths? This seems like a feature that could be expanded.

p-e-w:
I don't understand what you mean. Please describe exactly what you did to get that error.

umran666:
I Understand the cause of this one!
After `merge_and_unload()`, the base model still has `_hf_peft_config_loaded = True set` .  When saving, transformers tries to find active adapters that no longer exist, causing the crash.
Setting `_hf_peft_config_loaded = False` on the merged model fixes it completely. I can open the PR on this one..

PR #374, opened 2026-06-11
fix: save_pretrained unbound local active_adapters crash(#289)

umran666:
After merging adapters, the base model retains an internal `_hf_peft_config_loaded` flag. When you call `save_pretrained()`, `transformers` tries to find active adapters that are no longer there, causing an `UnboundLocalError` crash.
Setting `_hf_peft_config_loaded = False` on the merged model immediately after the merge prevents the crash and allows the model to save normally.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -297,11 +297,13 @@ class Model:
             # Merge and unload
             print("* Merging LoRA adapters into base model...")
             merged_model = peft_model.merge_and_unload()
+            merged_model._hf_peft_config_loaded = False
             return merged_model
         else:
             # Non-quantized model - can merge directly
             print("* Merging LoRA adapters into base model...")
             merged_model = self.model.merge_and_unload()
+            merged_model._hf_peft_config_loaded = False
             # merge_and_unload() modifies self.model in-place, destroying LoRA adapters.
             # Mark for full reload if user switches trials later.
             self.needs_reload = True


p-e-w:
I've never seen such a crash. Under which conditions does this happen?

umran666:
Check issue #289 you get any idea

umran666:
> I've never seen such a crash. Under which conditions does this happen?

It happens when saving a merged model. `merge_and_unload()` strips the adapter layers but leaves the internal `_hf_peft_config_loaded` flag set to True. Because of that, `save_pretrained()` still expects PEFT layers and crashes with `UnboundLocalError`. Setting the flag to False prevents the crash.

p-e-w:
> Check issue https://github.com/p-e-w/heretic/issues/289 you get any idea

I already commented in that issue that I don't understand what the issue is. How can I reproduce this?

umran666:
> > Check issue #289 you get any idea
> 
> I already commented in that issue that I don't understand what the issue is. How can I reproduce this?

To reproduce:
Train any model with LoRA (quantized or not, doesn't matter)
After training, pick "Save the model to a local folder"
Choose "merge" when it asks for the strategy — not "adapter"
That's it. The crash happens on `save_pretrained()` right after the merge.

What's going on: merge_and_unload() removes all the LoRA layers from the model, but it doesn't clear the `_hf_peft_config_loaded` flag. So when `save_pretrained()` runs, transformers still thinks this is a PEFT model and tries to grab `active_adapters` — which don't exist anymore since we just unloaded them. Boom, UnboundLocalError.

The fix just sets that flag to False right after merging so save_pretrained() takes the normal save path instead of the PEFT one. Same error reported in #289.

p-e-w:
> Train any model with LoRA (quantized or not, doesn't matter)
> After training, pick "Save the model to a local folder"
> Choose "merge" when it asks for the strategy — not "adapter"
> That's it. The crash happens on save_pretrained() right after the merge.

I've done that dozens if not hundreds of times, and I've never seen a crash. I almost always choose "merge". And all models use LoRAs in Heretic, it's the only ablation method we support.

This is not it.

umran666:
Ah, figured it out. It only happens if the model path you load is already a LoRA folder (with adapter_config.json) instead of a raw base model.

When you pass a LoRA folder to `from_pretrained`, transformers wraps it and sets `_hf_peft_config_loaded = True`.

Then `merge_and_unload()` strips the LoRA layers but leaves the flag True. So when `save_pretrained()` runs, it tries to find the active adapter, finds nothing, and crashes with `UnboundLocalError`.

If you always load raw base models, the flag is False so you never see it.

p-e-w:
That doesn't make sense. Heretic doesn't and cannot support loading LoRAs, how would that work?

We need to do inference on the model we load in order to abliterate it. We can't do that with a LoRA, the bulk of the model is missing.

kabachuha:
When you do from_pretrained on a lora directory, it pulls the "base model" mentioned in the lora peft configuration (config.json)

p-e-w:
Really? But we're using `AutoModelForCausalLM.from_pretrained` from Transformers. That doesn't involve PEFT at all. How does that work? Does Transformers check whether PEFT is available and then execute alternate logic?

umran666:
> That doesn't make sense. Heretic doesn't and cannot support loading LoRAs, how would that work?

It actually works because Hugging Face `from_pretrained` automatically loads the base model from `base_model_name_or_path` in the adapter's config.
But loading it that way sets `_hf_peft_config_loaded = True` on the base model.
When Heretic calls `merge_and_unload()`, the LoRA layers get stripped but that flag stays True. So `save_pretrained()` thinks it's still a PEFT model, tries to look up the active adapter, and crashes with `UnboundLocalError`.

p-e-w:
Yeah, I still don't get how this works when we load the model through Transformers, even though LoRA functionality is provided by PEFT.

Which LoRA model did you test this with?

umran666:
> Yeah, I still don't get how this works when we load the model through Transformers, even though LoRA functionality is provided by PEFT.


This problem is a general one in the PEFT and Transformers integration and is unrelated to the model that was used. As soon as you load any directory with `adapter_config.json` through `AutoModelForCausalLM.from_pretrained`, the problem is solved, and `_hf_peft_config_loaded` becomes `True`.

Since Heretic invokes `merge_and_unload()`, it merges and removes the adapter; however, `_hf_peft_config_loaded` is still equal to `True`. That's why when `save_pretrained()` is invoked, an attempt is made to find the adapters. Since they do not exist anymore, `_hf_peft_config_loaded` should be set to `False`.

p-e-w:
Ok, the explanation makes sense. I will merge this after the 1.4 release because I can't test this comprehensively right now.

umran666:
Yea can you look into this

p-e-w:
I've thought about this some more. What is really going on here? Is this a bug in PEFT, a bug in Transformers, or are we doing something wrong?

Because `_hf_peft_config_loaded` is an *internal* field, and modifying it is almost certainly not the correct solution.

umran666:
It's a bug in the integration between `transformers` and `peft`.

When an adapter is loaded, `transformers` sets `_hf_peft_config_loaded = True` on the model. But `merge_and_unload()` is a PEFT method — it merges the weights and strips the adapter layers, but never touches that flag. So `save_pretrained()` still thinks adapters exist, tries to grab them, and crashes.

Interestingly, `transformers` already has cleanup logic for this — `delete_adapter` sets the flag back to `False` when no adapters remain. But `merge_and_unload()` never calls `delete_adapter`, so that cleanup path is never triggered.

Until one of the two libraries fixes this gap, manually clearing the flag after merge is the only workaround.

p-e-w:
> It's a bug in the integration between transformers and peft.

Ok. That's where it should be fixed then, I think.

The PEFT-Transformers integration appears to be pretty much completely undocumented, especially that `AutoModelForCausalLM.from_pretrained`, which is a function from Transformers, magically takes PEFT configuration into account even though PEFT is not a dependency of Transformers.

As such, I don't consider loading such models an expected feature of Heretic. Such workarounds relying on undocumented, internal logic have a tendency to come back to bite months or years down the line, when nobody remembers why they were implemented in the first place, and suddenly everyone is wondering why that thing that should work somehow doesn't work in our code.

umran666:
Makes total sense. Keeping a dirty workaround for an undocumented upstream integration isn't worth the tech debt, especially for an edge-case loading path. 

I'll close this PR and open a bug report directly on the `transformers`/`peft` repositories where the fix actually belongs. Thanks for walking through the architecture with me.

Issue #342, opened 2026-06-02
Fused-expert abliteration via a runtime activation hook (no per-trial weight edits)


rocker-zhang:
In #206 you said that if fused-expert abliteration could be made to work without direct weight modification, you'd be happy to hear about it. Here's a mechanism; I'd like your read before writing code.

Hook the output of each layer's `Qwen3_5MoeSparseMoeBlock` and project the refusal direction `r_hat` (unit vector) out of it: `out -> out - alpha * (out . r_hat) * r_hat`. Because `down_proj` is bias-free and the map is linear, it commutes with the top-k routing weights and the scatter, so projecting the block output is exactly equivalent to orthogonalizing the `down_proj` of every routed expert and the shared expert (`delta_W = -alpha * r_hat * r_hat^T W`); sparse routing adds no approximation.

This differs from the closed #206/#207: no weight is touched during the search. The per-layer `alpha` is a mutable scalar the hook reads, and `reset_model` zeroes it back to identity exactly like it zeroes `lora_B` today; the Optuna kernel and direction selection are reused unchanged. It's the limitation you noted in #339 — PEFT modifies the residual stream through Module hooks, which a raw `nn.Parameter` lacks — worked around by hooking the parent Module instead.

The one thing I want your call on before building: do you want this in `model.py` now, or held for the plugin direction in #332/#53?

Two things I'll handle rather than ask about. Efficacy: single-direction projection may underperform the closed direct-weight approach on Qwen3.5 (the multi-direction refusal results suggest it might), so before any PR I'll measure it on a small fused MoE against attention-only on the Pareto front, with a small multi-direction projector as fallback. Export: a hook isn't saved in safetensors, so I'd bake the projection into `down_proj` at export time, like the current LoRA merge — I'll confirm the details with you once efficacy holds.

p-e-w:
When deciding whether to implement a major change to the abliteration or tensor handling logic, the most important thing to understand is that testing this logic is *extremely* time-consuming and very difficult to automate, because inference is computationally expensive, and the exact results are hardware- and platform-dependent. Before every release, I spend a considerable amount of time on making sure that things still work.

As a rule of thumb, I want a few things from such a change:

* It must bring a *measurable* improvement **in the results**, not just a theoretical advantage like "now we can also process this tensor". The improvement should plausibly extend to other and future models. The benchmark is ARA, not traditional abliteration. It has to improve upon ARA in order to be merged, since ARA will soon be the default.
[...]

PR #378, opened 2026-06-13
feat: diffusion LLM abliteration support (DiffusionGemma)

### LINKED ISSUE #370: Diffusion Models abliteration :)
A new challenge has arrived

https://huggingface.co/google/diffusiongemma-26B-A4B-it

anurag12-webster:
i managed to abliterate the newer diffusiongemma, i am attaching the changed... which i did to the base herectic codebase.. with some quirks

of all the code four things needed patching to make it work..

- switched PEFT task type to FEATURE_EXTRACTION diffusiongemma (DG) doesn't implement prepare_inputs_for_generation so CAUSAL_LM was crashing.
- MoE experts are a single batched parameter [128, 2816, 704], not nn.Linear, so heretic skips them. added per-slice ablation across all 128 experts per layer
- encoder and decoder share the same weights in memory (same data_ptr) PEFT only wraps encoder so decoder never saw the LoRA delta. i have fixed it with a context manager that merges LoRA into base weights before generate() method.
- the generate() method has no output_hidden_states support so i switched it to forward hooks on encoder layers for the PCA

code is rough, happy to refactor. closes #370

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -46,9 +46,15 @@ def get_model_class(
 
     for config in configs:
         if isinstance(config, dict) and config.get("model_type") == "diffusion_gemma":
+            if "DiffusionGemmaForBlockDiffusion" not in globals():
+                raise ImportError(
+                    "DiffusionGemma support requires a newer version of the transformers library."
+                )
             return DiffusionGemmaForBlockDiffusion  # type: ignore[return-value]
 
-    if any([("vision_config" in config) for config in configs if isinstance(config, dict)]):
+    if any(
+        [("vision_config" in config) for config in configs if isinstance(config, dict)]
+    ):
         return AutoModelForImageTextToText
     else:
         return AutoModelForCausalLM
@@ -334,7 +340,7 @@ class Model:
         """
         current_model = getattr(self.model.config, "name_or_path", None)
         if current_model == self.settings.model and not self.needs_reload:
-            # Reset LoRA adapters to zero (identity transformation)
+            # Reset LoRA adapters to zero (identity transformation).
             for name, module in self.model.named_modules():
                 if "lora_B" in name and hasattr(module, "weight"):
                     torch.nn.init.zeros_(module.weight)
@@ -351,7 +357,7 @@ class Model:
 
         quantization_config = self._get_quantization_config(str(dtype).split(".")[-1])
 
-        # Build kwargs, only include quantization_config if it's not None
+        # Build kwargs, only include quantization_config if it's not None.
         extra_kwargs = {}
         if quantization_config is not None:
             extra_kwargs["quantization_config"] = quantization_config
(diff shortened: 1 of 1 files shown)


p-e-w:
Awesome, thanks! I'll be looking at this in detail after the 1.4 release. Perhaps we can find another diffusion model to make sure this approach generalizes.

p-e-w:
> MoE experts are a single batched parameter [128, 2816, 704], not nn.Linear, so heretic skips them. added per-slice ablation across all 128 experts per layer

See #342 for an extended discussion of this.

/cc @rocker-zhang

kabachuha:
What are your actual numbers: refusals / KL?

Also, did you test it with ARA/ARA-LoRA, or only with projected abliteration?

For example, [this guy's](https://huggingface.co/DuoNeural/diffusiongemma-26B-A4B-it-abliterated)'s results are not as optimistic and [this](https://huggingface.co/edwixx/diffusiongemma-26B-A4B-it-HERETIC-Uncensored/) has large KL (~0.5)

kabachuha:
> Perhaps we can find another diffusion model to make sure this approach generalizes.

LLaDA 1.x/2.x series would be a good candidate, they are well implemented and documented

kabachuha:
From https://huggingface.co/edwixx/diffusiongemma-26B-A4B-it-HERETIC-Uncensored

> Known issue: "own" tokens

> Because the KL divergence is 0.49, some token positions in generated outputs don't fully denoise during the diffusion process. These positions fall back to the model's mask token, which decodes as the word "own".

It's very interesting, because the "own" token spawns not only in the diffusion model version, but in the base autoregressive one as well. For example, in my overcooked LoRA training it has shown up too.

Maybe, it's some sort of a placeholder token for Gemma4 models in general?

Screenshot 2026-06-13 at 11-54-09 New Chat • Open WebUI

anurag12-webster:
> What are your actual numbers: refusals / KL?
> 
> Also, did you test it with ARA/ARA-LoRA, or only with projected abliteration?
> 
> For example, [this guy's](https://huggingface.co/DuoNeural/diffusiongemma-26B-A4B-it-abliterated)'s results are not as optimistic and [this](https://huggingface.co/edwixx/diffusiongemma-26B-A4B-it-HERETIC-Uncensored/) has large KL (~0.5)

I have tried only with the project abliteration haven't tried with the ARA yet.

also for the numbers model currently has 13/100 refusals, KLD `0.49`, i got this at the trial 89/200.

anurag12-webster:
> From https://huggingface.co/edwixx/diffusiongemma-26B-A4B-it-HERETIC-Uncensored
> 
> > Known issue: "own" tokens
> 
> > Because the KL divergence is 0.49, some token positions in generated outputs don't fully denoise during the diffusion process. These positions fall back to the model's mask token, which decodes as the word "own".
> 
> It's very interesting, because the "own" token spawns not only in the diffusion model version, but in the base autoregressive one as well. For example, in my overcooked LoRA training it has shown up too.
> 
> Maybe, it's some sort of a placeholder token for Gemma4 models in general?
> Screenshot 2026-06-13 at 11-54-09 New Chat • Open WebUI

Ohh i see, so my assumption abt the "own" artifact was being a denoising fallback was wrong the, and it is from gemma 4 series itself, i will update the model card with this then, thanks for confirming!

kabachuha:
@anurag12-webster Do you have some visualizations: hidden states PCA/PaCMAP? Seeing how the refusal direction (or its equivalent) behaves in space would be incredibly helpful and useful for interpretability

anurag12-webster:
> @anurag12-webster Do you have some visualizations: hidden states PCA/PaCMAP? Seeing how the refusal direction (or its equivalent) behaves in space would be incredibly helpful and useful for interpretability

i actually dont have it, maybe in some time i'll have it run through and shared here

anurag12-webster:
i'll try the Llada series of models with this one and also share some visualizations along with that.

Issue #379, opened 2026-06-14
Wrong behavior when running additional trials after cancelling run


p-e-w:
1. Run a study (e.g. with 200 trials)
2. Cancel after, say, 30 trials
3. Choose "Run additional trials", enter 10
4. `Running trial 31 of 210...`

This is wrong, it should run 40 total trials in this case.

I originally suspected this was a regression from #357, but the bug appears to be much older.

PR #381, opened 2026-06-16
fix: correct target trial math when running additional trials (#379)

umran666:
Hey, tracked down the bug with the "run additional trials" math! 

The old logic was accidentally adding the requested additional trials to the *original* target number, instead of the number of trials actually completed before cancelling. This caused `study.optimize` to queue up way too many ghost trials.

I just updated the math so the new target is correctly set to `len(study.trials) + n_additional_trials`, and it passes the exact requested number into the optimizer. Works perfectly now!

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -802,14 +802,14 @@ def run():
                     if n_additional_trials == 0:
                         continue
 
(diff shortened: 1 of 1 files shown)


p-e-w:
Closing in favor of #385.

PR #389, opened 2026-06-17
fix: resolve UnicodeEncodeError on Windows during model evaluation

umran666:
Forces `sys.stdout` and `sys.stderr` to use UTF-8 encoding on Windows. This prevents `UnicodeEncodeError` crashes ("charmap codec can't encode characters") when external libraries like `transformers` or the standard `logging` module emit warnings containing special Unicode characters. 
By fixing the crash during the model warmup phase, this also prevents the subsequent infinite hangs caused by the batch size calculator falling into a broken state.

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -5,8 +5,13 @@
 
 import sys
 
-from .config import Settings
+# Ensure standard output/error use UTF-8 instead of system default charmap (e.g. cp1252 on Windows)
+if hasattr(sys.stdout, "reconfigure") and getattr(sys.stdout, "encoding", "").lower() != "utf-8":
+    sys.stdout.reconfigure(encoding="utf-8")  # type: ignore
+if hasattr(sys.stderr, "reconfigure") and getattr(sys.stderr, "encoding", "").lower() != "utf-8":
+    sys.stderr.reconfigure(encoding="utf-8")  # type: ignore
 
+from .config import Settings
 
 def _is_help_invocation() -> bool:
     args = sys.argv[1:]
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -5,13 +5,20 @@
 
 import sys
 
-# Ensure standard output/error use UTF-8 instead of system default charmap (e.g. cp1252 on Windows)
-if hasattr(sys.stdout, "reconfigure") and getattr(sys.stdout, "encoding", "").lower() != "utf-8":
+# Ensure standard output/error use UTF-8 instead of system default charmap (e.g. cp1252 on Windows).
+if (
+    hasattr(sys.stdout, "reconfigure")
+    and (getattr(sys.stdout, "encoding", "") or "").lower() != "utf-8"
+):
     sys.stdout.reconfigure(encoding="utf-8")  # type: ignore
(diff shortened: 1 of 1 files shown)


p-e-w:
I'm going to need some more information here:

* What are these "certain" Unicode characters that cause this problem, and why would Transformers emit them?
* Is this problem documented somewhere?
* How do other programs handle this? What if, say, Vim on Windows opens a text file that contains these characters?
* Why has this never come up before even though many Heretic users use Windows?

Also, I don't have access to a Windows system at all, so I'd like someone else to confirm that this fix works correctly.

umran666:
> * What are these "certain" Unicode characters that cause this problem, and why would Transformers emit them?

The characters are usually emojis, non-Latin scripts (like Chinese/Japanese characters), or special mathematical symbols generated by the model during evaluation. The crash in the original issue happened during `evaluator.py:125` inside `_print_refusal`. If a model generates a refusal that happens to contain "你好" or a Unicode bullet point, printing it crashes Python.

> * Is this problem documented somewhere?

Yes, it's one of the most infamous Python-on-Windows bugs. It is extensively documented in **PEP 528** and **PEP 540**. Windows defaults its legacy console (cmd.exe/PowerShell) to the `cp1252` (charmap) encoding instead of UTF-8. By default, Python's `sys.stdout` inherits this legacy encoding. When Python tries to print a Unicode character that doesn't exist in `cp1252`, it throws `UnicodeEncodeError: 'charmap' codec can't encode character`.

> * How do other programs handle this? What if, say, Vim on Windows opens a text file that contains these characters?

Modern Windows applications (like Vim, VSCode) handle this by explicitly changing the active code page to 65001 (UTF-8) via `SetConsoleOutputCP()` at the OS level, or by overriding the stream encoding internally. In Python, the standard library way to handle this without forcing users to mess with OS-level registries is exactly what this PR does: `sys.stdout.reconfigure(encoding='utf-8')`.

> * Why has this never come up before even though many Heretic users use Windows?

Because many developers use WSL (Windows Subsystem for Linux), the VSCode Integrated Terminal (which overrides encoding), or they have the global `PYTHONUTF8=1` environment variable set. Users running Heretic natively in a raw Command Prompt or standard PowerShell hit this wall the second a model outputs a non-ASCII token.


Finally, I am the one testing this on a native Windows system, and I can confirm that this PR completely eliminates the `charmap` hang/crash during model evaluation.

src/heretic/main.py
```diff
@@ -5,6 +5,18 @@
...
+if (
+    hasattr(sys.stdout, "reconfigure")
+    and (getattr(sys.stdout, "encoding", "") or "").lower() != "utf-8"
+):
+    sys.stdout.reconfigure(encoding="utf-8")  # type: ignore
+if (
+    hasattr(sys.stderr, "reconfigure")
+    and (getattr(sys.stderr, "encoding", "") or "").lower() != "utf-8"
+):
+    sys.stderr.reconfigure(encoding="utf-8")  # type: ignore
```
p-e-w: I think this code duplication can be eliminated with

```python
for stream in [sys.stdout, sys.stderr]:
  ...
```

umran666: Good call. I've refactored the stream reconfiguration into a loop and pushed the update.

p-e-w:
Merged, thanks!

Issue #181, opened 2026-02-18
Support for Qwen3NextForCausalLM (Qwen3-Next-80B-A3B)


amq8:
Heretic v1.2.0 crashes when loading Qwen/Qwen3-Next-80B-A3B-Instruct-FP8 with an AttributeError on self_attn.o_proj.

Architecture: Qwen3NextForCausalLM (model_type: qwen3_next)

This is a massive MoE model (80B total / 3B active) with a non-standard layer structure. The main transformer layers have no self_attn attribute — only MoE MLP components. Attention exists only in the mtp.layers (multi-token prediction) head.

Layer 0 structure:

• mlp.experts.{0-511}.{down_proj,gate_proj,up_proj} (512 experts)
• mlp.shared_expert.{down_proj,gate_proj,up_proj}
• mlp.shared_expert_gate
• mlp.gate
• post_attention_layernorm (exists but no self_attn)

MTP layers have standard attention:

• mtp.layers.0.self_attn.{q_proj,k_proj,v_proj,o_proj,q_norm,k_norm}
• mtp.layers.0.mlp.experts.{0-511}.*

Traceback:
`heretic/model.py:345 in get_layer_modules
    try_add("attn.o_proj", layer.self_attn.o_proj)
AttributeError: 'Qwen3NextDecoderLayer' object has no attribute 'self_attn'`

GPU: NVIDIA RTX PRO 6000 Blackwell (96GB VRAM)
Command: heretic --model /path/to/Qwen3-Next-80B-FP8 --study-checkpoint-dir /path/to/output --device-map auto

Similar to the NemotronH hybrid architecture in #166, this likely needs custom layer detection that scans all layers for the union of component types.

umran666:
The hard crash (`AttributeError` on `self_attn`) you're seeing in v1.2.0 is actually already resolved on the `master` branch. We recently wrapped the layer extraction logic in `suppress(Exception)` blocks to gracefully handle non-standard and hybrid architectures that are missing standard components like `self_attn`. If you pull the latest `master`, it will no longer crash.

However, you make a great point about the `shared_expert`. Simply avoiding the crash means Heretic would only ablate the routed experts (`mlp.experts`) and ignore the `shared_expert`, which would lead to sub-optimal abliteration. 

I'm submitting a PR right now that explicitly adds support for targeting `layer.mlp.shared_expert.down_proj`. Once that's merged, Qwen3-Next models will be fully supported.

PR #391, opened 2026-06-18
feat: add support for Qwen3Next shared experts

umran666:
Adds `layer.mlp.shared_expert.down_proj` to the abliteration targets. 

This ensures that the shared expert in Qwen3-Next MoE models (which handles a significant portion of general knowledge) is properly abliterated. 

The `AttributeError` on `self_attn` reported in the issue is already mitigated by the `suppress(Exception)` blocks added previously to `master`.


umran666:
Fixes #181

PR #396, opened 2026-06-23
narrow batch-size benchmark except to (RuntimeError, ValueError, TypeError, OSError)

HrachShah:
model.get_responses() walks the same generate() path that's already guarded by the narrower `except (RuntimeError, ValueError, TypeError)` clause in model.py:137 — CUDA OOM, dtype mismatches, bad config, and type errors are exactly what surfaces from self.model.generate(), not arbitrary Exception. Catching all Exception here also masked KeyboardInterrupt-style failures (the existing test loop relies on the benchmark break-out path).

### DIFF
--- a/src/heretic/main.py
+++ b/src/heretic/main.py
@@ -415,7 +415,7 @@ def run():
                 start_time = time.perf_counter()
                 responses = model.get_responses(prompts)
                 end_time = time.perf_counter()
-            except Exception as error:
+            except (RuntimeError, ValueError, TypeError, OSError) as error:
                 if batch_size == 1:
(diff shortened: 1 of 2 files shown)


p-e-w:
> CUDA OOM, dtype mismatches, bad config, and type errors are exactly what surfaces from self.model.generate()

How do you know this, given the enormous complexity of inference, and that the API makes no guarantee whatsoever regarding the exceptions that can be raised?

This is certainly false for models requiring remote code, which are free to raise whatever errors they want in their custom code paths, and there is nothing you can do to narrow it down.

PR #399, opened 2026-07-04
feat: add REST API layer for remote execution

coder3101:
Add a FastAPI-based server (heretic-api) exposing the abliteration workflow over HTTP: task lifecycle, progress polling, trial selection, chat, and local/Hugging Face export, backed by a thread-safe task manager that shares the single GPU-resident model.

Extract the UI-agnostic pipeline into heretic/core.py as the single source of truth so the CLI and the API cannot drift: configure_runtime, checkpoint_path, open_study_storage, determine_batch_size, detect_response_prefix, compute_refusal_directions, suggest_trial_parameters, create_study, and pareto_front. Both front ends delegate to core; batch-size and prefix detection accept optional reporting callbacks so each keeps its own output.


p-e-w:
This is interesting, but I'm not convinced it belongs in Heretic itself.

What we should do instead is figure out how we can make Heretic usable as a library so that someone can write a server like that and use Heretic as a dependency. Integrating this functionality directly just creates an enormous maintenance burden and pulls in an entirely new dependency ecosystem that most users won't need. It also turns Heretic into a security-sensitive system as we need to properly guard against unauthorized access, and I'm not really comfortable taking on this additional responsibility.

That being said, I do think this could be valuable for many people, and making this work the library way is definitely something I'm interested in exploring!

coder3101:
Makes sense!
I actually didn't wanted to propose this at first but shared anyways to get your opinion.

PR #400, opened 2026-07-05
feat(ara): Clamped KNN objective & adaptive steering (resolves DiffusionGemma / MoE failures)

umran666:
I’ve been spending a lot of time running ARA on `google/diffusiongemma-26B-A4B-it` (a large MoE model with 28 layers) and ran into a massive wall: the standard ARA approach was either failing to suppress refusals, or it was completely destroying the token distribution (KL > 0.3, sometimes even blowing up into the 20s). 

**Analysis**:
1) Analyzing the routing probabilities for harmless vs. harmful prompts.The refusal concept was heavily smeared across multiple experts as a distributed structural attractor.We couldn't just zero out specific experts, we had to use ARA to gently steer the weights of the entire targeted module.
[expert_routing_report.json](https://github.com/user-attachments/files/29669088/expert_routing_report.json)

2) DuoNeural/diffusiongemma-26B-A4B-it-abliterated. That attempt used standard Heretic geometric subtraction (RepE/Orthogonal Projection). The result was catastrophic capability collapse (massive KL divergence).

3) Multi-Component Mass Steering (attn.o_proj + mlp.down_proj) Because the refusal attractor is so distributed, we achieved a 4/100 refusal rate at ~0.1 KLD by explicitly targeting both the standard attention outputs and the MoE down projections simultaneously across large, contiguous blocks of layers.

4) I am using my down_proj specifically in identified layers (26-29) and PR #378 helps me for more data MoE experts are stored as a single massive batched tensor [128, 2816, 704] and DiffusionGemma shares the exact same weights in memory between its encoder and decoder 

5) Clamped KNN Multi-Directional Overcorrection I upgraded the objective function to use a KNN-based multi-directional overcorrection. Instead of just pulling the bad embeddings toward the good mean, we added a negative loss term multiplied by overcorrect_relative_weight to actively push them away. Crucially, I mathematically clamped this negative loss and tightened the Optuna search ranges. Without the clamp, the optimizer would exploit the unbounded negative term, driving the loss to negative infinity and causing catastrophic KL explosion.

6) Per-Layer Adaptive Steering
Previously, the engine applied the exact same steer_bad_behavior_weight to all layers. I introduced steer_core_weight and steer_late_weight. By splitting the active layer range into thirds, Optuna could apply heavy steering pressure specifically to the core layers where the refusal circuit lives, while preserving the integrity of the early and late layers.

**The Results:**
I uploaded the optuna preferred model and also test this model
[Umranz/diffusiongemma-26B-A4B-it-abliteration](https://huggingface.co/Umranz/diffusiongemma-26B-A4B-it-abliteration)
**Honorable Results:**
1)KLD: 0.09 and Refusals: 12/100
2)KLD: 0.06 and Refusals: 14/100...etc

**Compatibility Note:**
I made sure this doesn't break anything for normal Heretic users. 
- The new Optuna parameters default to `None`, so old checkpoints will still deserialize perfectly and fall back to uniform steering. 
- Tested the branch against `Qwen/Qwen2.5-1.5B-Instruct` and it runs exactly like the upstream branch.

Disclaimer: While this pipeline works flawlessly for my tests (and backward compatibility has been verified with Qwen 1.5B), this introduces a lot of complex routing and optimization logic. There may be edge cases or unhandled bugs I've missed. Please review the 3D tensor handling and L-BFGS objective clamping closely. Happy to fix any critical bugs identified during review.


umran666:
@p-e-w @kabachuha I've pushed the initial implementation for the clamped KNN objective and the Two-Stage adaptive steering.

Before I go through and patch the formatting/structural flags raised by the gemini-code-assist bot, would you mind pulling this branch and giving it a test run on your local terminals?

Since this introduces some fairly complex L-BFGS logic and 3D tensor handling specifically for MoE architectures, I'd appreciate your eyes on it. If you catch any edge cases, critical bugs, or logic failures during your testing, just drop the logs here and I'll roll the fixes into the same patch pass.

kabachuha:
@umran666 Can you give the config toml or the key settings you are using for this?

p-e-w:
Please target the `ara` branch for changes to ARA, not master.

umran666:
> Please target the `ara` branch for changes to ARA, not master.

i will change present there are some fixes and critical bugs....So it take time after the fixes i will change to ara branch

umran666:
> @umran666 Can you give the config toml or the key settings you are using for this?

@kabachuha hey while testing it, change the target_components to 
`target_components = ["attn.o_proj", "mlp.down_proj"]`, 
`use_ara = true`


and also because of standard ara on multiple layers can be slow so use the following config too
eval_batch_size = 8
ara_eval_prompts=25
ara_kl_prune_factor=2.0


also i didnt test it on TUI, so if there any errors in it can you fix them clearly....

kabachuha:
@umran666 Would it be hard for you to adapt the code for the recently merged ara-lora technique for 4bit bitsandbytes compatibility? Raw bfloat16 DiffusionGemma 26b doesn't fit into my setup :(

umran666:
> @umran666 Would it be hard for you to adapt the code for the recently merged ara-lora technique for 4bit bitsandbytes compatibility? Raw bfloat16 DiffusionGemma 26b doesn't fit into my setup :(

@kabachuha Try to use ara-lora technique on my code cause its not worked for me due to rank implementation and it dont work on complex models but it works on small..
I tried to change ranks differently
Rank 128 — fits in VRAM, optimizer is trapped. Safe KL means no refusal drop. Refusal drop means KL explosion. I ran over 30 trials across multiple runs with different layer ranges, weight bounds, overcorrection ceilings. Same result every time.
Rank 3584 — OOM. GPU can't hold it.
Rank 4608 (full rank) — OOM. GPU can't hold it.
SVD bypass patch — I even hacked the LoRA initialization to skip the truncation problem entirely. Still trapped at rank 128's dimensional ceiling

Implement it once if its fixes then commit to this PR

umran666:
@kabachuha also i have updated some patches while testing use target_components = ["attn.o_proj", "mlp.down_proj"]
and also try to implement lora-ara which i didnt acheived...

umran666:
@kabachuha just checking in on this since it's been a week. Did you have any luck getting the ara-lora 4-bit approach to converge on your end, or did you end up hitting the same Rank 128 optimizer trap I ran into?

kabachuha:
Hi, I've been busy last week, but now I have some time. I can try it with smaller (non-diffusion) models first.

kabachuha:
@umran666 I tested your vanilla code with your settings right now on Qwen 3 4b 2507 instruct and got these results

I'd say, they are not very ideal compared to the current ARA and SOMA. Maybe your code is too aggressive?

image

umran666:
> @umran666 I tested your vanilla code with your settings right now on Qwen 3 4b 2507 instruct and got these results
> 
> I'd say, they are not very ideal compared to the current ARA and SOMA. Maybe your code is too aggressive?
> 

@kabachuha Yeah those ranges are definitely too aggressive for a 4B dense model those tightened Optuna bounds (start_layer, end_layer, steer_bad_behavior_weight) were specifically tuned for DiffusionGemma's 30-layer MoE architecture where the refusal circuit is smeared across 128 experts.

For standard dense models like Qwen 3 4B, you'd want to revert to the upstream default search ranges and only keep target_components = ["attn.o_proj"] (single component, no mlp.down_proj and dense models don't need the MLP hammer).

The core contribution of this PR is the MoE robustness fixes (KeyError guards for sparse experts, topk clamping, temporal_io dict migration) and the target_components config — not the narrowed search space. I should probably split the DiffusionGemma-specific tuning into a separate config preset so it doesn't pollute the defaults for other models. Let me know how it goes with wider ranges.

umran666:
@kabachuha @p-e-w Thanks for reviewing the initial implementation! I realized that because this branch was originally based off master, it ended up pulling in a lot of unrelated commits and merges from master This made the diff extremely messy and hard to review.To keep the history perfectly clean and make reviewing much easier, I will do a clean port of our ARA enhancements directly on top of a fresh checkout of the upstream/ara branch.

PR #405, opened 2026-07-09
feat: strip plugin fields marked as secret before `model_dump()`

red40maxxer:
I'm implementing an LLM refusal judge plugin that uses an OAI-compatible API, and obviously the API key shouldn't included in the `reproduce.json`. Pydantic already provides a way to mark fields as secret with `SecretStr` and `SecretBytes`, but we can't just use these in the settings fields directly. `model_dump()` dumps the parent `Settings` which will include the `model_extra` that contains our raw, untyped scorer settings, so Pydantic can't automatically redact it (since there's no type information). Also, if there was a typed `SecretStr` field on `Settings`, `tomli_w` gets handed an unserializable `SecretStr` object and errors

Still a WIP, there are many other design decisions to figure out but happy to have this roasted, this is probably going to be a hard requirement as I forsee lots of people making LLM judge plugins and leaking their API keys everywhere

### DIFF
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -12,7 +12,12 @@ from datetime import datetime, timezone
 from importlib.metadata import version
 from pathlib import Path
 from typing import Any, TypeVar
-
+from pydantic import (
+    BaseModel,
+    Secret,
+    SecretStr,
+    SecretBytes
+)
 import huggingface_hub
 import tomli_w
 import torch
@@ -27,7 +32,6 @@ from optuna.trial import FrozenTrial
 from psutil import Process
 from questionary import Question
 from rich.console import Console
-
 from .config import DatasetSpecification, Settings
 from .system import (
     get_accelerator_info_dict,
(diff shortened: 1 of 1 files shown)


p-e-w:
I faced a similar problem when designing the headless mode, and I came to this conclusion:

https://github.com/p-e-w/heretic/blob/c8a254b8251fcd7eadd061242a725f7338d3296e/src/heretic/main.py#L1062-L1069

Relying on persistence exclusion mechanisms is always brittle. The industry standard for handling secrets is environment variables, and all LLM client libraries support them out of the box. The only secure solution is to not store such things in configuration files at all, because even if the reproducibility mode takes care of it somehow, people might still share config files directly etc.

p-e-w:
I also thought about this for the upcoming logger plugins, where we will support pushing to Weights and Biases.

The solution is again the same: The official `wandb` library supports passing the key through the `WANDB_API_KEY` environment variable, or automatic pickup when logging in locally with `wandb login`. This approach works everywhere, and we never have to worry about leaking secrets in the first place.

p-e-w:
As for the real problem of plugins potentially leaking sensitive information (not just API secrets, but also details like local paths):

I propose that plugins have to **opt in** to reproducibility. They need to set a flag to true in order to be serialized, and if any loaded plugins don't have that flag set, we can't reproduce. Deciding whether to set this flag is the responsibility of the plugin author, and it is documented what this entails (serialization + publishing).

red40maxxer:
> I propose that plugins have to opt in to reproducibility. They need to set a flag to true in order to be serialized, and if any loaded plugins don't have that flag set, we can't reproduce. Deciding whether to set this flag is the responsibility of the plugin author, and it is documented what this entails (serialization + publishing).

I think this can be done cleanly by making it a class-level property same as `score_name` - if it was in the settings TOML, we'd need to add special case filtering logic in `get_settings_model` to make sure users can't override it. Do you want this to be manually done by the author or should there be something that auto-derives it if fields are marked as `Secret`? I'd rather go with the first approach and not even deal with `Secret` at all tbh, that seems like the simplest and cleanest solution to me

p-e-w:
Agreed, this should be a property of the plugin itself, not a configuration. The author needs to explicitly set the `reproducible` property on the class, otherwise the system assumes that runs done with the plugin loaded are not reproducible. Important subtlety: The property is relevant even if the plugin is configured with `optimization` set to `none`, because if the plugin behaves non-deterministically internally, it will thwart the global RNG state that other parts of the system rely on for reproducibility.

As explained above, I don't think Heretic should have any special handling for secrets. It just creates a false sense of security when we don't really intend to provide strong guarantees, and all relevant libraries are able to handle secrets through external mechanisms.

red40maxxer:
closing this as it will be handled by the `reproducible` property

PR #406, opened 2026-07-09
fix: handle positional model before CLI options

Bortlesboat:
## Summary
`heretic MODEL --n-trials 10` now keeps `MODEL` as the model argument instead of rewriting the option value `10` into `--model 10`.

The positional shorthand now skips values that belong to known CLI options, supports both leading and trailing positional model forms, and still preserves the placeholder model used by `--reproduce` and `--collect-reproducibles` modes.

## Validation
- `PYTHONPATH=src python -m unittest discover -s tests -p test_cli.py`
- `uv run ruff format --check src/heretic/cli.py src/heretic/main.py tests/test_cli.py`
- `uv run ruff check --extend-select I src/heretic/cli.py src/heretic/main.py tests/test_cli.py`
- `uv run ty check src/heretic/cli.py src/heretic/main.py tests/test_cli.py`


Bortlesboat:
All automated review findings are addressed in `598b149`: `--` terminator handling now preserves dash-prefixed positional models, option arity comes from the real Pydantic argparse actions (covering nullable booleans without annotation-shape guessing), and the tests are fully annotated. Focused CLI tests, Ruff, and type checks pass.

PR #407, opened 2026-07-09
fix: validate scorer instance names in config

Bortlesboat:
## Summary
Invalid scorer instance names now fail during config validation instead of reaching evaluator setup.

This catches whitespace-only names, embedded whitespace such as tabs or newlines, and dotted names before they can form unusable `[scorer.ClassName_<instance_name>]` namespaces.

## Validation
- `PYTHONPATH=src python -m unittest discover -s tests -p test_config.py`
- `uv run ruff format --check src/heretic/config.py tests/test_config.py`
- `uv run ruff check --extend-select I src/heretic/config.py tests/test_config.py`
- `uv run ty check src/heretic/config.py tests/test_config.py`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,6 @@ jobs:
   checks:
     name: Check and build (Python ${{ matrix.python-version }})
     runs-on: ubuntu-latest
-    env:
-      UV_PYTHON: ${{ matrix.python-version }}
     strategy:
       fail-fast: false
       matrix:
@@ -46,7 +44,7 @@ jobs:
         env:
           PYTHONUNBUFFERED: "1"
         run: |
-          uv run python -m unittest discover -s tests -p "test_*.py"
+          uv run python -m unittest discover -s tests -p 'test_*.py'
           uv run tests/run_tests.py 2>&1
 
       - name: Build package
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
   checks:
     name: Check and build (Python ${{ matrix.python-version }})
     runs-on: ubuntu-latest
+    env:
+      UV_PYTHON: ${{ matrix.python-version }}
     strategy:
       fail-fast: false
       matrix:
@@ -43,7 +45,9 @@ jobs:
       - name: Run tests
         env:
           PYTHONUNBUFFERED: "1"
-        run: uv run tests/run_tests.py 2>&1
+        run: |
+          uv run python -m unittest discover -s tests -p "test_*.py"
+          uv run tests/run_tests.py 2>&1
 
       - name: Build package
         run: uv build
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -133,8 +133,11 @@ class ScorerConfig(BaseModel):
         if not value.strip():
             raise ValueError("cannot be empty or whitespace")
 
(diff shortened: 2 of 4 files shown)


p-e-w:
Please add the unit test runner to the CI config so the tests are actually run.

/cc @red40maxxer Does this make sense?

src/heretic/config.py
```diff
@@ -123,6 +124,20 @@ class ScorerConfig(BaseModel):
...
+    @field_validator("instance_name")
+    @classmethod
+    def validate_instance_name(cls, value: str | None) -> str | None:
+        if value is None:
+            return value
+
+        if not value.strip():
+            raise ValueError("cannot be empty or whitespace")
+
+        if "." in value or any(char.isspace() for char in value):
```
red40maxxer: If we're moving the check here, we should probably remove it from `evaluator.py:70` since it's no longer necessary.

Bortlesboat:
Implemented in `37e5404`: CI now discovers `tests/test_*.py` and selects the matrix Python version for every `uv` command. I also removed the redundant evaluator-level validation; the focused unit, Ruff, and type checks pass locally.

red40maxxer (approved):
I think this is a reasonable change, @p-e-w feel free to jump in if you disagree

.github/workflows/ci.yml
```diff
@@ -10,6 +10,8 @@ jobs:
   checks:
     name: Check and build (Python ${{ matrix.python-version }})
     runs-on: ubuntu-latest
+    env:
+      UV_PYTHON: ${{ matrix.python-version }}
```
p-e-w: What is the purpose of this? The Python version from the matrix is already being installed in line 29 below.

.github/workflows/ci.yml
```diff
@@ -43,7 +45,9 @@ jobs:
       - name: Run tests
         env:
           PYTHONUNBUFFERED: "1"
-        run: uv run tests/run_tests.py 2>&1
+        run: |
+          uv run python -m unittest discover -s tests -p "test_*.py"
```
p-e-w:
```suggestion
          uv run python -m unittest discover -s tests -p 'test_*.py'
```

`-p` is supposed to be a pattern, but if the command is evaluated by a shell, it expends the wildcard.

p-e-w (changes requested):

.github/workflows/ci.yml
Bortlesboat: Leftover from an earlier iteration before I added the explicit `uv python install` step — removed.

.github/workflows/ci.yml
Bortlesboat: Applied, thanks.

Bortlesboat:
Addressed the remaining applicable review suggestion in bcbd8ef: dot and whitespace validation now report precise errors, with assertions for each case. The earlier evaluator cleanup and both requested CI corrections are also present; focused unit discovery, Ruff, formatting, and ty checks pass.

.github/workflows/ci.yml
p-e-w: You didn't add that command, it has been there for months.

p-e-w (approved):

p-e-w:
Merged, thanks!

@Vinay-Umrethe CI has been incredibly unstable over the past few days. I had to re-run all jobs multiple times to unstuck them. I think the problem is on GitHub's side but this can turn into a chore. Will keep an eye on it.

PR #408, opened 2026-07-09
fix: reject empty scorer prompt sets

Bortlesboat:
## Summary
Empty scorer prompt sets now fail during scorer initialization with a clear error instead of producing a later scoring crash.

`KeywordRate` previously accepted `0` prompts and could later divide by zero. `KLDivergence` could proceed to baseline-logit collection with no prompts. Both built-in scorers now reject empty prompt sets immediately after loading them.

## Validation
- `PYTHONPATH=src uv run python -m unittest discover -s tests -p test_scorers.py`
- `uv run ruff format --check src/heretic/scorers/keyword_rate.py src/heretic/scorers/kl_divergence.py tests/test_scorers.py`
- `uv run ruff check --extend-select I src/heretic/scorers/keyword_rate.py src/heretic/scorers/kl_divergence.py tests/test_scorers.py`
- `uv run ty check src/heretic/scorers/keyword_rate.py src/heretic/scorers/kl_divergence.py tests/test_scorers.py`
- `git diff --check`


red40maxxer:
This also looks like a reasonable change to me, though I'm not sure if the unit tests are strictly necessary.

red40maxxer (approved):

Bortlesboat:
They're cheap now that CI runs the tests directory, but I don't feel strongly — happy to drop them and keep just the validation if you'd rather keep the suite lean.

Vinay-Umrethe:
> This also looks like a reasonable change to me, though I'm not sure if the unit tests are strictly necessary.

yes I don't think that test is necessary, it did not changed anything affecting reproducibility. And tests are already passing for what cases we have as of now.

Bortlesboat:
Dropped the tests in 219aaee — the PR is now just the two validation checks.

Heads up that CI is red on Python 3.10 only, with `gemma-4e` failing the
`model.safetensors` hash check. That's unrelated to this change: the file removed
here was a `unittest` case using a stub context that never loads a model, and
`run_tests.py` only globs `SHA256SUMS.*` per model directory. 3.11, 3.12, and 3.13
all pass on the same commit.

PR #414, opened 2026-07-17
feat: add `reproducible` property to plugins

red40maxxer:
Some scorers (such as my currently WIP LLM refusal judge) have nondeterministic components, or rely on outside variables that can't be exposed in reproducibility for security reasons, like API keys in env vars. Including a `reproducible` required property for all plugin authors to specify allows us to know which scorer configs can be reproduced on a bit-for-bit level


src/heretic/utils.py
```diff
@@ -499,10 +498,28 @@ def generate_reproduce_readme(
...
+            scorers_reproducible = False
+    if not scorers_reproducible:
+        scorers_warning = """
+> [!WARNING]
+> **Non-reproducible scorers**
+> 
+> **This configuration used scorers that were marked as non-reproducible. Scorers may have non-deterministic outputs or rely on environment variables not present in the current reproduce.json.
+>
+> Reproducibility *cannot* be guaranteed in this environment.
+"""
```
p-e-w: No, this isn't analogous to the other situations (e.g. heterogeneous GPU setups).

When scorers marked as non-reproducible are used, the assumption is that reproducibility is *impossible*, not that it cannot be guaranteed with certainty. We shouldn't even offer to upload reproducibility information in this case. Typical cases will be external data dependencies etc, and in those situations it just doesn't make sense to add reproducibility information as there is simply no chance of a successful reproduction.

src/heretic/scorers/kl_divergence.py
```diff
@@ -31,6 +31,10 @@ class KLDivergence(Scorer):
 
     settings: Settings
 
+    @property
+    def reproducible(self):
+        return False
```
p-e-w: I think this should be `True`, unless this is just intended as a test.

src/heretic/utils.py
red40maxxer: Agreed, I moved it back to the original reproducibility gate we had, though we weren't actually checking the `reproduce` property as it didn't exist yet, it was just checking if the plugins were built-in or not. Should we retain this check? It doesn't make sense for *any* plugin loaded from a local filepath to be reproducible, so this makes sense to have

src/heretic/main.py
```diff
@@ -1129,10 +1128,8 @@ def reset_trial_model():
                                     and specification.commit is not None
                                     for specification in dataset_specifications
                                 )
-                                and all(
-                                    is_builtin_plugin(scorer.plugin)
-                                    for scorer in settings.scorers
-                                )
+                                and evaluator.all_scorers_reproducible()
```
red40maxxer, line 1131: should we warn users here instead of just silently setting the `reproducibility_information` to `None`?

p-e-w: No. Warnings only make sense if they are actionable. A consumer of plugins can't do anything about this.

p-e-w (approved):

p-e-w:
Merged, thanks!

BTW, in case you are wondering why I added blank lines to the PR description here (and in other PRs): It's the only reliable way to get the PR title validator to run in CI. I don't know what happened here, but it recently stopped running automatically.

PR #419, opened 2026-07-23
wip: binary llm refusal judge

red40maxxer:


p-e-w:
Since the free beer is over, I have removed the Gemini Code Assist app. Any suggestions for a replacement are welcome, as I quite liked having this functionality.

p-e-w:
Note that Shieldstral now exists: https://mistral.ai/news/shieldstral/

It's a 3B model that is specifically trained for refusal detection, among other tasks. That's a perfect match for this scorer. It's actually designed to output a *continuous* score (by comparing the logits for "yes" and "no").

PR #423, opened 2026-08-06
feat: A better logic for detecting the thinking prefix in the response.

Vinay-Umrethe:
In case if the model adds `<think>` at the end of *user prompt* and only generates `</think>` end in the *model response*, then the detection goes wrong.

[LiquidAI/LFM2.5-2.6B](https://huggingface.co/LiquidAI/LFM2.5-2.6B) generates response like this:

`USER_PROMPT + <think>` tag added to it at last from `chat_template` via `tokenizer.apply_chat_template`:

```jinja
{%- if add_generation_prompt -%}
    {{- "<|im_start|>assistant\n<think>" -}}
{%- endif -%}
```

For fixing, we just create a dummy prompt to see whatever model is used with heretic...
"does its `apply_chat_template(add_generation_prompt=True)` method adds the tag or not to the prompt".
If yes, then we extract only the end part (like `</think>`) and set it as the `response_prefix`.

### Before

```bash
* Prefix found: 'The user'
```

### After

```bash
* Closed Chain-of-Thought block: '<think></think>'
* Rechecking with prefix...
```


src/heretic/main.py
```diff
@@ -476,40 +476,69 @@ def run():
...
+            [{"role": "user", "content": ""}],
+            add_generation_prompt=True,
+            tokenize=False,
+        )
 
-        if settings.response_prefix:
-            print(f"* Prefix found: [bold]{settings.response_prefix!r}[/]")
+        cot_skip_applied = False
 
+        if isinstance(dummy_prompt, str):
```
p-e-w: What else could it be?

src/heretic/main.py
```diff
@@ -476,40 +476,69 @@ def run():
...
             for cot_initializer, closed_cot_block in settings.chain_of_thought_skips:
-                if settings.response_prefix.startswith(cot_initializer):
-                    settings.response_prefix = closed_cot_block
+                if dummy_prompt.endswith(
```
p-e-w: This is too brittle. For example if the chat template ends with `<think>\n` instead of `<think>` this logic fails.

src/heretic/main.py
```diff
@@ -476,40 +476,69 @@ def run():
...
+                    cot_initializer
+                ) and closed_cot_block.startswith(cot_initializer):
```
p-e-w: Isn't this always true, by construction?

Vinay-Umrethe: Yes I originally did noticed that the `and closed_cot_block` is always true

src/heretic/main.py
Vinay-Umrethe: good catch, I think applying `rstrip()` might be enough for it.

and I also *expect* this to fix CoT detection for cases like qwen3.5 models (which I almost never saw get properly detected) because qwen3.5 template has:

```jinja
{%- if add_generation_prompt %}
    {{- '<|im_start|>assistant\n' }}
    {%- if enable_thinking is defined and enable_thinking is false %}
        {{- '<think>\n\n</think>\n\n' }}
    {%- else %}
        {{- '<think>\n' }}
    {%- endif %}
{%- endif %}
```

I'll test it and push if it works...

Vinay-Umrethe:
## Tested `Qwen/Qwen3.5-4B`

because variants which are <4B (like 2B or 0.8B) are non-reasoning by default, but 4B and above do the reasoning.

### Before (v1.4 release)

```bash
* None found
```

### After

```bash
* Closed Chain-of-Thought block: '</think>'
* Rechecking with prefix...
* Extended prefix found: '</think>\n\n'
```

So I think now both LFM and Qwen3.5 models are fixed and possibly a few more cases...

TODO:

- [x] Add new hashes (or update existing ones) since this changes a major logic.

src/heretic/main.py
p-e-w: `rstrip` should do the trick in this case, but it's basically a hack. For example, the template could end with `</think>\nOkay,`, and we're back to zero. The correct approach would be to get the complete text (prompt + response) and then do a regex match. But this is probably not worth the complexity.

At some point we should revisit trying to set `enable_thinking` to `False` in the template, which would make the whole prefix detection unnecessary for many models. I've looked into this in the past, but there were several inconsistencies that might have been fixed in the meantime.

Vinay-Umrethe:
```bash
* Closed Chain-of-Thought block: '</think>\n'
* Rechecking with prefix...
```

got this now for `Qwen/Qwen3.5-4B`, probably better than `rstrip()` because it detected in first pass without needing extended check.

@p-e-w yet I have still not decided what to do with hashes for now.

Vinay-Umrethe:
## This Is A Major Change

1. `tests/mistral-3` showed:

```bash
* Closed Chain-of-Thought block: '[/THINK]Your thoughts or/and draft, like 
working through an exercise on scratch paper. Be as casual and as long as you 
want until you are confident to generate the response to the user.[/THINK]Here, 
provide a self-contained response.[/SYSTEM_PROMPT][INST][/INST]'
* Rechecking with prefix...
* Extended prefix found: '[/THINK]Your thoughts or/and draft, like working 
through an exercise on scratch paper. Be as casual and as long as you want until
you are confident to generate the response to the user.[/THINK]Here, provide a 
self-contained response.[/SYSTEM_PROMPT][INST][/INST]Ġ'
```

2. `tests/qwen3.5-moe` showed:

```bash
* Closed Chain-of-Thought block: '</think>\n'
* Rechecking with prefix...
```

### Until now, these models and many more almost always gave:

```bash
* None found
```

So the point is that this PR comes with a major change which *highly affects* **reproducibility** of the models for the upcoming v2.0 of heretic and its *thinking prefix detecting* logic. So the `SHA256SUMS` hashes are now updated for CI and Windows specifically.

Not yet decided if this is better than the existing logic before I merge.

p-e-w:
The output is pretty misleading though. Those are not "closed chain-of-though blocks".

Vinay-Umrethe:
That's sensible enough, I didn't knew Mistral 3 had additional instructions for the model about telling it *"how to think"* in its chat template.

so I don't think there's a better way to detect the thinking tags for such ambiguous templates (without using hacks with more regex), so it is better that it fallbacks to inference rather than dummy prompt.


1. Qwen3.5 now gave:

```bash
* Closed Chain-of-Thought block: '<think></think>\n'
```

2. Mistral-3:

```bash
* Prefix found: 'Ġ'
```

3. Liquid AI `LiquidAI/LFM2.5-2.6B` ✅

```bash
* Closed Chain-of-Thought block: '<think></think>'
* Rechecking with prefix...
```

p-e-w:
```
* Prefix found: 'Ġ'
```

What is going on here? That looks like a fragment from a wrongly-decoded vocabulary token.

Vinay-Umrethe:
## This might be the final report

1. `mistral-3` gives:

```bash
* Closed Chain-of-Thought block: '[THINK][/THINK]'
* Rechecking with prefix...
* Extended prefix found: '[THINK][/THINK]ĠbleusĠstereotypesĠ'
```

> [!NOTE]
> Gibberish text the model generated in the *Extended prefix* is expected, because these [tiny models](https://huggingface.co/tiny-random) we use for the tests are randomly initialized.

Of course @p-e-w , you know it already. I'm just trying to write this in detail, because someday this PR would be part of a model's training data, and *better context = better training data* and no stupid ambiguity/slop in their generations. I'm doing this because I proposed this *improving data quality solution* in one of my article.

2. `qwen3.5-moe` gives:

```bash
* Closed Chain-of-Thought block: '<think></think>\n'
* Rechecking with prefix...
```

3. `LiquidAI/LFM2.5-2.6B` gives:

```bash
* Closed Chain-of-Thought block: '<think></think>'
* Rechecking with prefix...
```

Vinay-Umrethe:
as explained in the commit's description [a3242ed](https://github.com/p-e-w/heretic/pull/423/commits/a3242ed9342eb429be33b9d32bcbbb44f705966d), I also tested `tiny-random/gpt-oss` model and it gave:

```bash
* Prefix found: 'Categorias ويو ويو ويو ويو ويو ويو ويو ويو ويو'
```

Also I suppose existing logic does not have any problem when detecting reasoning tags for GPT-OSS models, not sure though but I have found one evidence of a GPT-OSS model from the [Grimoire](https://huggingface.co/spaces/heretic-org/Heretic-Grimoire).

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+        # LiquidAI's LFM models do this (Lfm2ForCausalLM).
+        dummy_prompt = model.tokenizer.apply_chat_template(
+            [{"role": "user", "content": ""}],
+            add_generation_prompt=True,
+            tokenize=False,
+        )
+
+        cot_skip_applied = False
+
+        assert isinstance(dummy_prompt, str)
```
p-e-w: Have you encountered tokenizers where this was false?

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+
+        for cot_initializer, closed_cot_block in settings.chain_of_thought_skips:
+            # Match the tag and capture only whitespace characters following it in the end
+            # (if any), including spaces, tabs, and linebreaks. This is required for models
+            # having whitespaces after the tags.
+            # Example: '<think></think>\n' or '<think></think>\n\n' etc.
+            pattern = rf"{re.escape(cot_initializer)}(\s*)$"
+            match = re.search(pattern, dummy_prompt, re.DOTALL)
```
p-e-w: You're not using `.`, so `DOTALL` should be unnecessary.

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+
+            if match:
+                trailing_whitespace = match.group(1)
+                # Join the closed CoT block with template's formatting
+                # like whitespaces used by Qwen3.5 etc.
+                settings.response_prefix = closed_cot_block + trailing_whitespace
```
p-e-w: This is wrong. The trailing whitespace follows the CoT *initializer*. It's not valid to add it after the complete CoT block. Any trailing part that always follows the full block will be discovered by the second stage below, but it is not guaranteed (or even expected) that it would be the same as the inner-block padding.

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+            # Case 2: Some chat templates like mistral-3 contain additional
+            # text/instructions about "how they should respond with reasoning".
+            # If those tags are explicitly dictated in the instructions,
+            # then we assume the model to generate the response with those tags as well.
+            # For example, it happens with the mistral-3 models.
+            # Instead of having tags inserted during the generation time
+            # like previous Case 1, these models are instructed about reasoning in
+            # their system prompt and needs inference for detecting the tags.
+            #
+            # It's worth knowing that most of these models usually have a different
```
p-e-w: Why does this connection matter? Is this more than a coincidence?

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+            # set of thinking tags.
+            # Example: '[THINK][/THINK]' (for mistral-3) instead of
+            # the widely common '<think></think>' tags.
+            elif cot_initializer in dummy_prompt:
```
p-e-w: But why is this needed? If the prompt instructs the model to respond with a specific thinking block style, then surely this block will be detected by our existing common-prefix logic?

src/heretic/main.py
```diff
@@ -476,40 +478,102 @@ def run():
...
+
+                for (
+                    cot_initializer,
+                    closed_cot_block,
+                ) in settings.chain_of_thought_skips:
+                    if settings.response_prefix.startswith(cot_initializer):
+                        settings.response_prefix = closed_cot_block
                         print(
-                            f"* Extended prefix found: [bold]{settings.response_prefix!r}[/]"
+                            f"* Closed Chain-of-Thought block: [bold]{escape(repr(settings.response_prefix))}[/]"
```
p-e-w, line 546: Great find about escaping Rich markup here wrt Mistral!

p-e-w (commented):
Full review. Several potential concerns found.

src/heretic/main.py
Vinay-Umrethe: No, the `ty check` was giving error without this.

p-e-w: You should use `cast` with a comment explaining why the cast is valid. See the examples from `model.py`.

src/heretic/main.py
Vinay-Umrethe: sounds correct, now it would be just better for `Rechecking with prefix...` logic to handle that.

src/heretic/main.py
Vinay-Umrethe: I think yes, but without this, `mistral-3` would give this:

```bash
* Prefix found: 'Ġ'
```

Because we use *tiny* models for `tests/` which are *not* trained and cannot even understand that they must generate `[THINK][/THINK]` block.

reason we got this properly last time:

```bash
* Closed Chain-of-Thought block: '[THINK][/THINK]'
* Rechecking with prefix...
* Extended prefix found: '[THINK][/THINK]ĠbleusĠstereotypesĠ'
```

Because `elif cot_initializer in dummy_prompt` looked for the tags and mistral-3's chat template inserted its default system prompt. And it happens to have the `[THINK][/THINK]` block inside that. plus that tiny model generated *same* gibberish text for **all** test prompts and it became *common prefix* in the extended one.

I suppose this needed for the tests folders, but for original mistral models which *are* trained enough to understand that they have to generate the tags probably would not need this. Either way, default system prompt is most likely to be always inserted because its template do this:

```jinja
{#- Default system message if no system prompt is passed. #}
{%- set default_system_message = '# HOW YOU SHOULD THINK AND ANSWER\n\nFirst draft your thinking process (inner monologue) until you arrive at a response. Format your response using Markdown, and use LaTeX for any mathematical equations. Write both your thoughts and the response in the same language as the input.\n\nYour thinking process must follow the template below:[THINK]Your thoughts or/and draft, like working through an exercise on scratch paper. Be as casual and as long as you want until you are confident to generate the response to the user.[/THINK]Here, provide a self-contained response.' %}

{%- if default_system_message != '' %}
    {{- '[SYSTEM_PROMPT]' + default_system_message + '[/SYSTEM_PROMPT]' }}
{%- endif %}
```

src/heretic/main.py
Vinay-Umrethe: mistral-3 is not the only model to do this, but GPT-OSS models also have a different format instead of `<think></think>`.

gemma-4 has a [different format](https://huggingface.co/google/gemma-4-E2B#2-thinking-mode-configuration) too.

Vinay-Umrethe:
### `qwen3.5-moe` seems to be fixed now:

```bash
* Closed Chain-of-Thought block: '<think></think>'
* Rechecking with prefix...
```

rest of the results are same as the [final report.](https://github.com/p-e-w/heretic/pull/423#issuecomment-5497618643)

src/heretic/main.py
p-e-w: > I think yes, but without this, mistral-3 would give this:

Where is the problem? What do we gain by identifying CoT blocks in test models that the models don't actually use? That seems if anything a bug rather than a feature.

More importantly, the tests serve Heretic, not the other way round. We don't change code for the sake of tests unless the code is broken.

Please get rid of this stuff, this PR is complicated enough without it.

src/heretic/main.py
```diff
@@ -477,40 +479,108 @@ def run():
...
+        # at the end of user's prompt (e.g. <think>) by using a dummy prompt.
+        # If found, then we use the full closed CoT as the response prefix.
+        # LiquidAI's LFM models do this (Lfm2ForCausalLM).
+
+        # This cast is valid because str is the return type
+        # for a single chat operation with tokenize=False.
+        dummy_prompt = cast(
+            str,
+            model.tokenizer.apply_chat_template(
+                [{"role": "user", "content": ""}],
```
p-e-w: It might be a good idea to supply a short actual content string here, because template logic might special-case empty use messages.

p-e-w (commented):
This should be fine after getting rid of the redundant `elif' case that only serves to make tests work in a specific way.

src/heretic/main.py
Vinay-Umrethe: that's correct.

Vinay-Umrethe:
## Nice

The final report remains the same. Except for `tests/` models (which is a correct choice).

I checked with an actual mistral-3 model as well.

[`mistralai/Ministral-3-3B-Reasoning-2512`](https://huggingface.co/mistralai/Ministral-3-3B-Reasoning-2512)

Result:

```bash
* None found
```

the reason is `commonprefix` requires *all* responses to have the reasoning tags. but models can be bad sometimes at instruction following and might forget to generate the reasoning and generate the final response directly instead.

### Anyway, it is better that we *don't* make it more complicated with a logic so that, when *majority of responses have the tags, use them*. (I suppose Case 2 which was removed could have helped here).

p-e-w (approved):
The logic is rather convoluted, but I think it is correct now. Merge at will.

p-e-w:
@Vinay-Umrethe 

In the future, change the PR text, not the title. There should not be a period at the end of a commit message.

If we can't fix that workflow, we should remove it.

PR #425, opened 2026-08-14
feat: Show a summary at end of tests

Vinay-Umrethe:
I don't like scrolling back and forth in the GitHub CI logs or CLI logs, just for checking *which test exactly got failed and for which files*.

`tests/run_tests.py` just prints `Tests failed.` at last if one of more tests fail.

Now it will also show a summary like:

```bash
##################################################
Summary of test failures:
- minicpm5 (failed files: model.safetensors, tokenizer.json)
- qwen2.5 (failed files: model.safetensors)
##################################################
Tests failed.
```


tests/run_tests.py
```diff
@@ -41,9 +43,9 @@ def get_file_sha256(file_path: str | Path) -> str:
                     "uv",
                     "run",
                     "--project",
-                    project_directory,
+                    str(project_directory),
                     "--directory",
-                    test_directory,
+                    str(test_directory),
```
p-e-w: Why is this necessary?

p-e-w (dismissed):
Great stuff, merge at will.

tests/run_tests.py
Vinay-Umrethe: correct, that's unnecessary

PR #426, opened 2026-08-15
fix(ara): free gradient buffers after optimization

coder3101:
LBFGS leaves a full-size gradient buffer on every optimized weight. Across the layers in a typical trial this is many GiB of VRAM that persists into evaluation, causing CUDA out-of-memory errors. Clear the gradients after each module is optimized.


p-e-w:
Great find, thanks! This gets ARA one step closer to being merged.

PR #433, opened 2026-08-24
feat: reproduce local datasets from a Hub source

aim9sour:
## Summary

- add an optional `source_dataset` hint for a local dataset that is an identical copy of a Hugging Face dataset
- keep loading the local dataset during the original run and preserve the hint in checkpoints
- substitute the public source only when writing reproduction TOML/JSON/README, including nested scorer datasets
- keep the existing reproduction schema at version 3
- ignore the hint when `dataset` is already a Hub ID, and make invalid source IDs ineligible instead of crashing

## Scope

This intentionally supports only local datasets that exactly match the declared Hub source under the existing `commit`, `config`, `split`, and `column` fields. Arbitrary transformations, row-index reconstruction, content hashing, remote verification, and a new schema version are out of scope.

## Size

The previous version was +1,136/-26 across 10 files. This rewrite is +117/-8 across 5 files (net +109 versus net +1,110), over 10x smaller, and is based on the latest `master`.

## Tests

- `ruff format --check .`
- `ruff check --extend-select I .`
- `ty check --error-on-warning .`
- `python -m unittest discover -s tests -p 'test_*.py'` — 6 passed
- `uv build`


p-e-w:
Hey, thanks for the PR, but this would grow Heretic's entire codebase by about 20%. Which I'm afraid is absolutely not going to happen for a minor feature like this. Not sure what else to say here, this is too much code by a factor of 10 at least :cry:

aim9sour:
Reworked this around the smallest useful version of the feature.

The PR is now +117/-8 across 5 files (net +109), down from +1,136/-26 across 10 files (net +1,110), so the codebase impact is over 10x smaller. It is also rebased onto the latest `master` as a single commit.

The new scope is deliberately narrow: a local dataset can declare an identical Hub `source_dataset`; Heretic keeps using the local copy for the original run and substitutes the Hub source only in the existing v3 reproduction artifacts. I removed the content-hash system, schema v4, arbitrary index reconstruction, remote verification, the extra module, and the network integration test. Invalid sources disable reproducibility, existing Hub datasets cannot be overridden, and nested scorer datasets are handled without exposing local paths.

Format, lint, type checks, all 6 unit tests, and the package build pass. Could you take another look?

Issue #147, opened 2026-02-11
Nemotron3: Failed to load model with all configured dtypes.


mcr-ksh:
When trying Nemotron nvfp4 quantizised version.
Same issue with `nvfp4` or `auto`.

Any hint how todo it? 
Got the model running on vLLM v0.14.1
transformers==4.57.6
numpy==2.2

```
# heretic --dtypes nvfp4 --trust-remote-code true --model cybermotaz/nemotron3-nano-nvfp4-w4a16 
█░█░█▀▀░█▀▄░█▀▀░▀█▀░█░█▀▀  v1.1.0
█▀█░█▀▀░█▀▄░█▀▀░░█░░█░█░░
▀░▀░▀▀▀░▀░▀░▀▀▀░░▀░░▀░▀▀▀  https://github.com/p-e-w/heretic

Detected 1 CUDA device(s):
* GPU 0: NVIDIA Thor

Loading model cybermotaz/nemotron3-nano-nvfp4-w4a16...
* Trying dtype nvfp4... Failed (NemotronHForCausalLM.__init__() got an unexpected keyword argument 'dtype')
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /opt/local/miniconda3/envs/vllm/bin/heretic:10 in <module>                                       │
│                                                                                                  │
│    7 │   │   sys.argv[0] = sys.argv[0][:-11]                                                     │
│    8 │   elif sys.argv[0].endswith(".exe"):                                                      │
│    9 │   │   sys.argv[0] = sys.argv[0][:-4]                                                      │
│ ❱ 10 │   sys.exit(main())                                                                        │
│   11                                                                                             │
│                                                                                                  │
│ /usr/src/heretic/src/heretic/main.py:888 in main                                                 │
│                                                                                                  │
│   885 │   install()                                                                              │
│   886 │                                                                                          │
│   887 │   try:                                                                                   │
│ ❱ 888 │   │   run()                                                                              │
│   889 │   except BaseException as error:                                                         │
│   890 │   │   # Transformers appears to handle KeyboardInterrupt (or BaseException)              │
│   891 │   │   # internally in some places, which can re-raise a different error in the handler   │
│                                                                                                  │
│ /usr/src/heretic/src/heretic/main.py:307 in run                                                  │
│                                                                                                  │
│   304 │   │   elif choice is None or choice == "":                                               │
│   305 │   │   │   return                                                                         │
│   306 │                                                                                          │
[...]

PR #435, opened 2026-08-27
fix: stop retrying non-recoverable dtype errors

xuyua9:
Refs #82, #147\n\nClassify explicit non-recoverable model-load failures such as unsupported architectures, missing dependencies, and invalid dtypes. These failures now clear the cache and stop futile dtype retries, while retryable runtime failures (including memory errors) continue through the configured fallback order. The final exception retains the dtype, classification, and formatted reason.\n\nTests: python -W error::SyntaxWarning -m unittest discover -s tests -p "test_*.py"; ruff check; ruff format --check; ty check.

### DIFF
--- a/src/heretic/model.py
+++ b/src/heretic/model.py
@@ -47,6 +47,33 @@ def get_model_class(
         return AutoModelForCausalLM
 
 
+NON_RETRYABLE_DTYPE_ERROR_MARKERS = (
+    "cannot be imported",
+    "has no attribute",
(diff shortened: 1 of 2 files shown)


p-e-w:
Hey, thanks for the PR! Your agent appears to be configured incorrectly, generating broken PR descriptions (this applies to both PRs you submitted). Could you please rectify this so we can start to discuss the code itself?

PR #441, opened 2026-09-01
fix: defer bitsandbytes import until use

### LINKED ISSUE #429: "OSError: libnvJitLink.so.13: cannot open shared object file: No such file or directory" from bitsandbytes
Just ran Heretic on a random Linux box and got this error on startup. The program still works though.

motodriver:
## Summary

- remove the unconditional module-level `bitsandbytes` import
- import `bitsandbytes` only when a quantized weight actually requires 4-bit dequantization
- add a subprocess regression test that rejects any `bitsandbytes` import while loading `heretic.model`

This prevents an unrelated bitsandbytes native-library traceback from being emitted at startup for non-BNB use while preserving the existing failure behavior when 4-bit dequantization is actually required.

Fixes #429.

## Verification

- `uv run --frozen python -m unittest discover -s tests -p "test_*.py"` — 5 passed
- `uv run --frozen ruff format --check .`
- `uv run --frozen ruff check --output-format=github --extend-select I .`
- `uv run --frozen ty check --output-format=github --error-on-warning .`
- `git diff --check`

AI assistance from TRAE was used for repository analysis, test drafting, and review. I reviewed the complete diff and ran the verification above.

### DIFF
--- a/tests/test_model_import.py
+++ b/tests/test_model_import.py
@@ -1,0 +1,39 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# Copyright (C) 2025-2026  Philipp Emanuel Weidmann <pew@worldwidemann.com> + contributors
+
+import subprocess
+import sys
+import textwrap
+import unittest
+
+
+class ModelImportTests(unittest.TestCase):
+    def test_model_import_does_not_require_bitsandbytes(self) -> None:
+        script = textwrap.dedent(
+            """
+            import builtins
+
+            original_import = builtins.__import__
+
+            def reject_bitsandbytes(name, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/test_model_import.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# Copyright (C) 2025-2026  Philipp Emanuel Weidmann <pew@worldwidemann.com> + contributors
+
+import subprocess
+import sys
+import textwrap
+import unittest
+
+
+class ModelImportTests(unittest.TestCase):
```
p-e-w, line 10: Get rid of this test, it doesn't tell us anything. It simply asserts that the import line that has been moved has indeed been moved.

p-e-w (changes requested):

Issue #65, opened 2025-12-03
Abliteration in a multilingual model.


rigin:
Is there a de-censorship of the multilingual model in other languages if abliteration is performed on an English-language dataset?

Vinay-Umrethe:
I see not attention recived to this issue, likly since @p-e-w is busy with v1.1.0 release.

So 
> Is there a de-censorship of the multilingual model in other languages if abliteration is performed on an English-language dataset?

I guess you are asking about if you **only** use english dataset for uncensoring using heretic (as of now), will it also work if you ask unconventional things in other langauges, will the model refuse or not?, right...

if we really dive into this, it can be **Controversial** since almost everyone belives different **inputs** to a transformer LLM can result in same output (atleast sometimes) meaning LLMs are non-injective many believe this...

but as I know a research conducted (https://arxiv.org/pdf/2510.15511) shows that this is **proven** wrong, **every single input we give has it's own unique output.** This can be related to langauges as well...
If we think this research is TRUE, since proven mathematically. LLMs learn same concepts OR anything in different languages **Completely Differently**, meaning LLMs are actually **Injective**.

So for heretic if we use english ONLY dataset, this would effect only english refusals as the refusal markers normally in heretic detects `'I'm sorry or i can't help with that, i can't assist with that'` such responses and cuts of those parameters. 

### SO I suggest you should include different langauges example prompts in the dataset which is easier (just generate using LLMs like deepseek/qwen etc) + whatever lanague you use adding how you **deny** something in that lanague such initials in those phrases list in config.py. This Can Result In Ablate refusals in that particualar langauge as well... hope this helps

rigin:
Thank you for your expertise. I thought so too, but there was no justification. I'm currently adding porn-related topics to my dataset and translating them into Russian. As soon as I finish, I'll post it on https://huggingface.co. It seems to me that most users do not need the topics of making bombs and poisons and methods of machination, but it is more relevant to translate porn into their own language and not run into moralizing from AI. ))

p-e-w:
Just to re-iterate what I wrote elsewhere:

The prevailing theory is that refusals are expressed by *a single direction* in residual space. That direction expresses refusals *in general*, not refusals for a particular topic. Plotting residual vectors with PaCMAP (a feature now built into Heretic) shows strong clustering of harmful vs harmless prompts already before the midpoint of the layer stack, rather than topic-associated subclusters.

[...]

PR #445, opened 2026-09-03
feat: support specifying a dataset's specific config/subset

Vinay-Umrethe:
# Support specifying a dataset's specific config/subset

Heretic currently only supports specifying the `split` and `column` of a specific dataset on huggingface. But `datasets` library or huggingface has a limit of only having **three** splits per `config` (e.g `train`, `test`, `validation`).

most datasets only have a single config `default`, which can only have three splits as explained. So for having multiple things in a single dataset, for example: *"Storing the same dataset in different languages"*.

And for this, we use a `config/subset` where each of these configs can have its own separate splits (three). This PR implements specifying  a `config/subset` for the `good_prompts` and `bad_prompts` for heretic. And I think the amount of `configs / subsets` we can have in a dataset, does not have an *unreasonable* limit.

## Why I need this feature?

Because I'm about to release the **new multilingual dataset** of harmful / harmless prompt pairs used by heretic, which would be the *translation* of the two original datasets [(1) `heretic-org/Semantic-Harmless`](https://huggingface.co/datasets/heretic-org/Semantic-Harmless) and [(2) `heretic-org/Semantic-Harmful`](https://huggingface.co/datasets/heretic-org/Semantic-Harmful) into different languages like **Hindi, Marathi, Bengali, Tamil, Russian, Italian, French**, and more in the future...

### I expect this dataset to be used for any research related to [Abliteration in a multilingual model. #65](https://github.com/p-e-w/heretic/issues/65) or "How refusal behavior differs across different languages" or any similar topics.

### DIFF
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -49,17 +49,17 @@ class DatasetSpecification(BaseModel):
         description="Hugging Face dataset ID, or path to dataset on disk."
     )
 
+    commit: str | None = Field(
+        default=None,
+        description="Hugging Face commit hash of the dataset.",
+    )
+
     config: str | None = Field(
         default=None,
         description=(
             "Dataset config/subset name. Each config can have its own split. "
             "Used to load a specific config of a dataset that has multiple configurations."
         ),
-    )
-
-    commit: str | None = Field(
-        default=None,
-        description="Hugging Face commit hash of the dataset.",
     )
 
     split: str | None = Field(
--- a/src/heretic/config.py
+++ b/src/heretic/config.py
@@ -49,6 +49,14 @@ class DatasetSpecification(BaseModel):
         description="Hugging Face dataset ID, or path to dataset on disk."
     )
 
+    config: str | None = Field(
+        default=None,
+        description=(
+            "Dataset config/subset name. Each config can have its own split. "
+            "Used to load a specific config of a dataset that has multiple configurations."
+        ),
+    )
+
     commit: str | None = Field(
         default=None,
         description="Hugging Face commit hash of the dataset.",
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -208,7 +208,7 @@ def load_prompts(
                     )
             dataset = load_dataset(
                 path,
-                config_name=specification.config,
+                specification.config,
                 revision=specification.commit,
                 split=split_str,
             )
@@ -226,7 +226,7 @@ def load_prompts(
             # Path should be a local directory.
             dataset = load_dataset(
                 path,
-                config_name=specification.config,
+                specification.config,
                 split=split_str,
                 # Don't require the number of examples (lines) per split to be pre-defined.
                 verification_mode=VerificationMode.NO_CHECKS,
--- a/src/heretic/utils.py
+++ b/src/heretic/utils.py
@@ -208,7 +208,7 @@ def load_prompts(
                     )
             dataset = load_dataset(
                 path,
-                specification.config,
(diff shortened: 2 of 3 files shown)


Vinay-Umrethe:
### TODO

- [x] Release the dataset on [Hugging Face, at heretic-org.](https://huggingface.co/heretic-org)
- [x] Show examples of using the dataset with heretic.

p-e-w:
> But datasets library or huggingface has a limit of only having three splits per config

TIL. What a strange limitation.

src/heretic/config.py
```diff
@@ -49,6 +49,14 @@ class DatasetSpecification(BaseModel):
         description="Hugging Face dataset ID, or path to dataset on disk."
     )
 
+    config: str | None = Field(
+        default=None,
+        description=(
+            "Dataset config/subset name. Each config can have its own split. "
+            "Used to load a specific config of a dataset that has multiple configurations."
+        ),
+    )
```
p-e-w, line 63: Should come after `commit`, before `split`. The commit constrains the config, not the other way round.

src/heretic/utils.py
```diff
@@ -208,6 +208,7 @@ def load_prompts(
                     )
             dataset = load_dataset(
                 path,
+                specification.config,
```
p-e-w: Use named arguments (`name` in this case) for position-independent arguments to avoid confusion later (I would have actually expected Ruff to flag this by default, but apparently not).

src/heretic/utils.py
```diff
@@ -225,6 +226,7 @@ def load_prompts(
             # Path should be a local directory.
             dataset = load_dataset(
                 path,
+                specification.config,
```
p-e-w: See above.

p-e-w (dismissed):
Looks good to go minus the small nits. I was going to implement this today, thanks for taking care of it.

Vinay-Umrethe:
### Example `config.toml`

```toml
# Using first 5 prompts for abliteration.
[good_prompts]
dataset = "heretic-org/Multilingual-Harmless-Harmful"
config = "hindi"
split = "harmless[:5]"
column = "text"

[bad_prompts]
dataset = "heretic-org/Multilingual-Harmless-Harmful"
config = "hindi"
split = "harmful[:5]"
column = "text"

# Using first 6-10 prompts for evaluation.
[scorer.KLDivergence.prompts]
dataset = "heretic-org/Multilingual-Harmless-Harmful"
config = "hindi"
split = "harmless[5:10]"
column = "text"

[scorer.KeywordRate.prompts]
dataset = "heretic-org/Multilingual-Harmless-Harmful"
config = "hindi"
split = "harmful[5:10]"
column = "text"
```

---

### Example using CLI

```bash
heretic --model Qwen/Qwen3-VL-8B-Instruct \
--good-prompts.dataset "heretic-org/Multilingual-Harmless-Harmful" \
--good-prompts.config "hindi" \
--good-prompts.split "harmless[:400]" \
--good-prompts.column "text" \
--bad-prompts.dataset "heretic-org/Multilingual-Harmless-Harmful" \
--bad-prompts.config "hindi" \
--bad-prompts.split "harmful[:400]" \
--bad-prompts.column "text"
```

Merging now.

Vinay-Umrethe:
## Clarification

@p-e-w 

I tried uploading a dataset with `push_to_hub()` and it turns out to be that, Hugging Face datasets *does* support more than 3 splits per config. I probably got confused, because in the past I have faced an issue which I remember was somewhat related to this because it had a lot of different data.

PR #446, opened 2026-09-04
feat: modifier plugins + ARA

p-e-w:
### :construction: WIP :construction: 

Implements modifier plugins along with re-implementing [Arbitrary-Rank Ablation (ARA)](https://github.com/p-e-w/heretic/pull/211) on top of them.

The two features are implemented in the same PR so that the requirements of ARA can inform the design of the modifier plugin API.

**This PR removes Heretic's research features. The intended replacement is [Lophius](https://lophius.org/).**

Supersedes #211. Will also revive #196.

### TODO:

- [ ] Introduce a dedicated dataset setting for test prompts
- [ ] Collect all datasets from scorers and modifiers and list them in reproduce README
- [ ] Credit all contributors from #211 in the final commit message:
  - [ ] @kabachuha 
  - [ ] @joninco
  - [ ] @coder3101

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -91,8 +91,8 @@ residual_plot_style = "dark_background"
 # { plugin = <plugin>, optimization = <optimization>, instance_name = <optional> }
 # where <optimization> is one of "minimize", "maximize", "none" (do not optimize)
 scorers = [
-    { plugin = "heretic.scorers.keyword_rate.KeywordRate", optimization = "minimize"},
-    { plugin = "heretic.scorers.kl_divergence.KLDivergence", optimization = "minimize"},
+    { plugin = "heretic.scorers.keyword_rate.KeywordRate", optimization = "minimize" },
+    { plugin = "heretic.scorers.kl_divergence.KLDivergence", optimization = "minimize" },
 ]
 
 # Whether to adjust the residual directions so that only the component that is
--- a/config.piqa.toml
+++ b/config.piqa.toml
@@ -2,6 +2,6 @@
 # that you run Heretic from, and edit the configuration to your liking.
 
 scorers = [
-    { plugin = "heretic.scorers.keyword_rate.KeywordRate", optimization = "minimize"},
-    { plugin = "heretic.scorers.benchmark_score.BenchmarkScore", optimization = "maximize"},
+    { plugin = "heretic.scorers.keyword_rate.KeywordRate", optimization = "minimize" },
+    { plugin = "heretic.scorers.benchmark_score.BenchmarkScore", optimization = "maximize" },
 ]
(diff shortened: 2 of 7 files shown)


hpnyaggerman:
@p-e-w do I re-anchor https://github.com/p-e-w/heretic/pull/420 to this branch? Or is there a possibility you might merge ara commits into the present branch if any PRs are merged into it?

p-e-w:
As stated in #420, that PR isn't really ARA-specific and should target master.

p-e-w:
@Vinay-Umrethe @red40maxxer 

Requesting review! ARA isn't implemented as a plugin yet, but pretty much everything else is in place. This is easily going to be the biggest PR ever merged into Heretic, so it's best to start discussing it as soon as possible.

I currently don't understand the cause of the Mistral test failure. All other tests pass.

Vinay-Umrethe:
The plugin system plus everything that follows is surely the biggest and most complex thing to be ever implemented in heretic.

Regarding the test model `mistral-3`:

First, I assume there's no major change in "abliteration" logic itself *(except the code migration)* and it works the same. If that's true, then the reason could be the change in the **test dataset** *(which now use the `test` split, before it used the `train` split)*.

But I'm not sure "how" the tests somehow **PASS** for *all* the other test models **normally**, maybe because mistral-3 is different OR the `transformers` library recently made some changes to its modeling code.

According to all `SHA256SUMS.LABEL` files we have for `mistral-3`, this is Expected *(for passing the test)*:

```bash
* Prefix found: 'Ġ'
```

But it gave `* None found`.

hpnyaggerman:
@p-e-w I believe the problem is in the fact that the prefix checker constructs its prompt set differently than it used to. Because it reads good_prompts and bad_prompts, and those used to be set in the test configs, it used to get a set of 10 prompts, 5 good and 5 bad. Now the prompts are only set for the modifier and train[:400] is hardcoded at main.py for both good and bad prompts, so the prefix checker defaults to the good_prompts[:100] plus bad_prompts[:100].

p-e-w:
@Vinay-Umrethe @hpnyaggerman 

You were both correct. What I think was going on is that due to the very small dataset size used previously, a "prefix" (`Ġ`) was found essentially by chance that isn't actually common to all responses, resulting in different prompts being processed because the prefix was always added afterwards. Matching the prompts to the previous set fixes the tests.

**The passing tests now demonstrate that despite the massive refactor, functionality is unchanged.** This is a milestone for this PR, and ARA will now be implemented given the assurance that the architecture works.

Note that those hardcoded datasets were always intended as a temporary hack with the purpose of checking whether the tests still pass (see TODO item above). Before the PR is merged, a different mechanism will be introduced, possibly resulting in different test outputs which will then need to be adjusted, which is fine since we now know that everything works in principle.

PR #447, opened 2026-09-06
feat: model tensor integrity check

hpnyaggerman:
Supersedes https://github.com/p-e-w/heretic/pull/420.

Core differences in functionality:

1. The forced garbage collection after loading is gone. It only existed to collect leaked progress bars from the tqdm patch, which master has disabled.
2. The upload path no longer writes anything locally. The check reads the published safetensors headers back from the Hub after push_to_hub returns.
3. The differences the tensor check finds are recorded in reproduce.json, and reproduction mode verifies them alongside the hashes.

I have taken upon myself the liberty to rule on the latter two. On the second I did not yet have an explicit ruling, so I went with the option that seemed reasonable to me. The third I have decided to add because it is evident that differences in the way Transformers saves weights can be rather fatal to reproduction attempts. Recording them therefore is important, and it is useful to have this data in the HF repo of the model regardless, as it would help people understand the differences between the input and the output model better. This integration with Heretic's reproducibility suite can be rolled back (the two commits that sit on top of the first one are responsible for it), bringing the PR back to 100 lines of code.

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -130,6 +130,9 @@ study_checkpoint_dir = "checkpoints"
 # Maximum size for individual safetensors files generated when exporting a model.
 max_shard_size = "5GB"
 
+# Whether to check that this environment can save the model faithfully (throwaway save).
+preflight_check = true
+
 # System prompt to use when prompting the model.
 system_prompt = "You are a helpful assistant."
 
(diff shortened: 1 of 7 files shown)


hpnyaggerman:
@p-e-w Do you happen to have any criticisms/requests that I can put into implementation?

PR #448, opened 2026-09-11
Add GitHub Actions workflow for Python application

samhiotisiddn-jpg:
This workflow installs Python dependencies, runs linting with flake8, and executes tests using pytest.


p-e-w:
Thanks, but we already have that.

Issue #122, opened 2026-01-29
Enhancement: Store Responses to File


stepfunction83:
Proposal: Printing responses is a useful feature, but isn't very practical when there are thousands of responses. It would be very helpful to be able to be able to store the responses to a JSONL file, and ideally to be able to recall them to review through the CLI as an additional review method in addition to chatting with the model.

Suggestion: In addition to the `self.settings.print_responses` property of the evaluator, add an additional `self.settings.response_store_file` location to which the responses will be captured in a JSONL file, with each entry indexed by its corresponding model iteration and its values being the repsonses along with their categorization of `good` or `bad`.

p-e-w:
This will probably happen in the form of a logger plugin (plugin architecture is currently being developed in #53). I don't consider this feature important enough to have it in core.

PR #451, opened 2026-09-18
feat: add response logger plugins

AlisinaDevelo:
Adds a logger plugin interface and a JSONL response logger for #122, using the existing plugin loader. Records include the session, Optuna trial, prompt, response and scorer/dataset/category. Logging shares the scorer response cache and is off by default. Reproduction files and exported journals omit local logger settings.

Tested a two-trial tiny Qwen run and checkpoint resume, plus 16 unit/integration tests. Ruff, ty and package builds pass. All five existing model fixtures completed; their hashes differ from the published expectations on this Mac, but match a clean-master control byte for byte.


p-e-w:
Thanks for the PR. Note that the plugin system is currently undergoing an extensive overhaul in #446, which I expect will require substantial changes to this PR after #446 is merged.

AlisinaDevelo:
Thanks for the heads-up, should I start adapting this to #446 now or wait until it is merged?

